PR #7, opened 2015-01-03
Update to rust master: fixed array length fixes + lib stabilizations

vhbit:


pyfisch:
Build passes.
But tests still fail to build for me.

vhbit:
It looks pretty much like an error from fixed length array change, but I can't test yet as I have a local dylib mess and `regex_macros` fails to build at all.

vhbit:
@pyfisch I've pushed an update, can you test it as I'm not sure how long it will take to solve dylib issue?

pyfisch:
All builds pass. (Like on travis) for ac2a0dc

pyfisch:
All is ok. 
The deprecated things can probably changed in a separate pr.
* deprecated syntax; use the `self` keyword now
* `deriving` is deprecated; use `derive`

vhbit:
Pushed update for deprecation just a couple seconds ago :smile:

alexcrichton:
Thanks!

PR #10, opened 2015-01-05
Keep version in README in sync with latest version

carols10cents:
The current version in the Cargo.toml and on crates.io is 0.1.7, so the README should have the same version... right?


alexcrichton:
Thanks! It's also fine to show something like `regex = "0.1"` which indicates the "0.1 series", but this is fine by me!

PR #12, opened 2015-01-06
Update to rust master - no more boxed closures

johannhof:
I hope simply using FnMut is the right thing here :)


alexcrichton:
Thanks!

PR #19, opened 2015-01-13
Fixes #18.

### LINKED ISSUE #18: iterating over empty matches is wrong
This code:

```rust
extern crate regex;

use regex::Regex;

fn main() {
    let re = Regex::new(".*?").unwrap();
    for m in re.find_iter("ΛΘΓΔα") {
        println!("{:?}", m);
    }
}
```

When run violates an assert in `std::str`:

```
[andrew@Liger play] ./target/scratch 
thread '<main>' panicked at 'assertion failed: (w != 0)', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcore/str/mod.rs:1477
```

The precise reason is probably in how `find_iter` (and `captures_iter`) handles zero-length matches. It *has* to make progress after a zero-length match in the input, otherwise the iterator won't terminate. [Currently, it makes progress by adding `1` to a *byte offset into a UTF-8 encoded string*, which is obviously wrong.](https://github.com/rust-lang/regex/blob/master/src%2Fre.rs#L881-L884) This should be relatively easily fixed by setting `self.last_end = self.last_end + length_of_char_that_starts_at(self.last_end)`.

BurntSushi:
Previously, iterating over zero-length matches did not advance
correctly at UTF-8 character boundaries.

(The `empty_match_unicode_*` tests are regression tests.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /Cargo.lock
 /regex_macros/target
 /regex_macros/Cargo.lock
+.*.swp
--- a/src/re.rs
+++ b/src/re.rs
@@ -15,6 +15,7 @@ use std::borrow::IntoCow;
 use std::collections::HashMap;
 use std::fmt;
 use std::string::CowString;
+use unicode::str::utf8_char_width;
 
 use compile::Program;
 use parse;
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks!

PR #20, opened 2015-01-14
Resolve #17 optimize use of '^' / '\A'

### LINKED ISSUE #17: optimize use of `^`/`\A`
At the moment, the VM always scans the [entire input](https://github.com/rust-lang/regex/blob/master/src/vm.rs#L139) unless one of the two following things happens:

* The regex has a *literal prefix* and that prefix cannot be found anywhere in the input. (This always scans the entire input.)
* A match is found. This may *not* scan the entire input, but it requires a match.

Ideally, given a regex `^{something}...`, it could terminate the search early if `{something}` doesn't match at the start of the input.

tmerr:
The problem was that a regular expression like `^x` would take linear time to is_match a string `aaa....` when all it needed to do was check the first character then bail out. This PR makes it so if the regex begins with `^` (without multiline) or `\A`, then it stops iterating over characters as soon as there are no possible matches.

Benchmarks:

```
dynamic:
anchored_literal_long_non_match: 7492 ns/iter => 435 ns/iter
anchored_literal_short_non_match: 917 ns/iter => 431 ns/iter

native:
anchored_literal_long_non_match: 5020 ns/iter => 50 ns/iter
anchored_literal_short_non_match: 375 ns/iter => 50 ns/iter
```

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -144,12 +144,18 @@ impl<'r, 't> Nfa<'r, 't> {
                     break
                 }
 
+                // If the expression starts with a '^' we can terminate as soon
+                // as the last thread dies.
+                if self.ic != 0 && prefix_anchor {
+                    break;
+                }
+                
                 // If there are no threads to try, then we'll have to start
                 // over at the beginning of the regex.
                 // BUT, if there's a literal prefix for the program, try to
                 // jump ahead quickly. If it can't be found, then we can bail
                 // out early.
-                if self.prog.prefix.len() > 0 && clist.size == 0 {
+                if self.prog.prefix.len() > 0 {
                     let needle = self.prog.prefix.as_bytes();
                     let haystack = &self.input.as_bytes()[self.ic..];
                     match find_prefix(needle, haystack) {
--- a/regex_macros/src/lib.rs
+++ b/regex_macros/src/lib.rs
@@ -191,6 +191,11 @@ fn exec<'t>(which: ::regex::native::MatchKind, input: &'t str,
                     if matched {
                         break
                     }
+
+                    if $prefix_anchor && self.ic != 0 {
+                        break
+                    }
+
                     $check_prefix
                 }
                 if clist.size == 0 || (!$prefix_anchor && !matched) {


huonw:
Thanks!

Could I ask you to expand the commit message a little to describe what the problem is and how it is fixed? Just saying the issue number is better than nothing, but it does mean that someone looking the source and checking the git history needs to cross reference against github (which may have disappeared at that point).

(I'm happy to help if you need assistance with git or anything else.)

tmerr:
Sure, expanded. Is that alright?

src/vm.rs
```diff
@@ -144,12 +144,18 @@ impl<'r, 't> Nfa<'r, 't> {
...
+                if self.ic != 0 && prefix_anchor {
+                    break;
+                }
+                
                 // If there are no threads to try, then we'll have to start
                 // over at the beginning of the regex.
                 // BUT, if there's a literal prefix for the program, try to
                 // jump ahead quickly. If it can't be found, then we can bail
                 // out early.
-                if self.prog.prefix.len() > 0 && clist.size == 0 {
```
BurntSushi: Why did you remove `&& clist.size == 0`? That doesn't seem right to me. I think literal prefixes should only be tried if there are no other threads to try. Moreover, that check is still in place in the `regex!` macro code: https://github.com/tmerr/regex/blob/master/regex_macros/src/lib.rs#L548-L553

BurntSushi: I'm kind of concerned that this doesn't cause a test to fail... Perhaps I need to come up with one. :-)

BurntSushi:
@tmerr This looks great. I'm happy to see that the change is small. I just had one question, but otherwise this looks awesome. Thank you!

src/vm.rs
huonw: This whole section of code seems to be inside a `if clist.size == 0 {` branch.

BurntSushi: Ah... doy. Thanks :-)

tmerr: Ahh it's hard to see from the diff snippets but in the original code it was redundant
```
if clist.size == 0 {
    ...
    if self.prog.prefix.len() > 0 && clist.size == 0 {
    ...
    }
}
```
It's not my top priority, I just tried to sneak it the removal :P

huonw:
@tmerr that messsage looks much better, thanks! Could you place it into the git commit message (e.g. `git commit --amend` and then `git push -f` to update the PR by force pushing over this branch.)

src/vm.rs
tmerr: I'm perfectly happy to revert that part if you want

BurntSushi: Oh no no. It's great. :-) It's my misunderstanding!

tmerr:
@huonw Done, thanks for the help!

It's been a blast reading through this code by the way. I had been aware of NFAs but only  the "draw pretty circles on paper with lines between them" version, not as a machine with split and jump instructions. The two models of thinking about it meld together nicely

huonw:
Thanks! I merged as 10edff6bb76f1c9f4423f3bb8ad373d0c7a69773 (there was a piece of trailing whitespace which I took the initiative to remove).

PR #32, opened 2015-01-28
Fix for rust nightly

jrasky:
Fix for breaking lifetime changes introduced on the 27th.

Of note: tests *do* pass, but only build if linked against a libregex and a libregex_macros that *wasn't* built as a test. Building the tests against a libregex_macros built with --test causes Rust to complain about regex($re) to not live long enough.

I'm not entirely sure as to the cause, but given compiler plugins are experimental it isn't entirely surprising.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -119,8 +119,8 @@ impl<'r, 't> Nfa<'r, 't> {
         };
         let mut matched = false;
         let ninsts = self.prog.insts.len();
-        let mut clist = &mut Threads::new(self.which, ninsts, ncaps);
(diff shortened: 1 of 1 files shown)


alexcrichton:
I'm a little confused by github right now, it says this can merge cleanly but I think I already fixed this in https://github.com/rust-lang/regex/commit/6c0f8aa82582441e9cb3813048ef9b098c026db0, so I'm not totally sure what's going on!

Regardless, I think this fix is already in master, so just be sure to run `cargo update` to get the latest version, thanks anyway though!

Issue #33, opened 2015-01-28
plan for 1.0 beta/stable channel?


BurntSushi:
This is mostly about `regex_macros` and its future. Once beta hits, `regex_macros` will fail to compile on anything but the nighties until a plugin API has been stabilized. How are we going to manage that in this repository? i.e., The repo would contain one crate (`regex`) that should work on Rust 1.0 stable (which currently has a `dev-dependency` on `regex_macros`) and another crate (`regex_macros`) that can only work on Rust nightlies. Is it possible for them to cohabitate?

Of secondary concern is the docs. There is a lot of language strongly suggesting use of the `regex!` macro. Methinks this has to be removed or changed, because teasing users with such an awesome feature is going to make them mad. :-)

Of third concern is the programming shootout. There is a `regex-dna` benchmark. How do we handle it? Oh, wait, errmm, it [appears to have disappeared](https://github.com/rust-lang/rust/tree/master/src/test/bench)? Ah, [found the commit](https://github.com/rust-lang/rust/commit/6c29708bf906fa9075bb96b76fd7f6cc81eda43c). That's reasonable. Will it not be included in the shootout any more?

N.B. @erickt is working on [rust-syntex](https://github.com/erickt/rust-syntex), which I think ought to support use of the `regex!` macro. This may help with the docs, but `regex_macros` still needs to be handled.

alexcrichton:
> How are we going to manage that in this repository?

Having a `dev-dependency` on `regex_macros` only kicks in during development, so publication to crates.io for `regex` should work fine on and off nightly. It just means that if you use `regex_macros` with stable rust you'll get some compile errors.

> Is it possible for them to cohabitate?

I think the implication is that this repository must be developed with nightly, but that's not necessarily the end of the world!

> Methinks this has to be removed or changed, because teasing users with such an awesome feature is going to make them mad. :-)

I'd be fine with this!

> Of third concern is the programming shootout.

Oh I meant to do this, I'd be fine moving the shootout benchmark into this repo under `benches`

> N.B. @erickt is working on rust-syntex, which I think ought to support use of the regex! macro

That'd be pretty sweet to get the compile time optimizations on stable rust so quickly!

BurntSushi:
> I think the implication is that this repository must be developed with nightly, but that's not necessarily the end of the world!

[...]

PR #43, opened 2015-02-28
Divorce regex_macros from regex.

### LINKED ISSUE #31: Regex docs don't point people to crates.io dependency
Reading http://doc.rust-lang.org/regex/regex/index.html and http://doc.rust-lang.org/regex/regex/enum.Regex.html there is no hint to the user that they should be updating `Cargo.toml` with something like:

```toml
[dependencies]
regex = "*"
regex_macros = "*"
```
The error doesn't help either:

```
   Compiling example v0.0.1 (file:///path/to/projects/example)
src/main.rs:5:1: 5:27 error: can't find crate for `regex_macros`
src/main.rs:5 extern crate regex_macros;
```

BurntSushi:
Fixes #31 and #33.

There are a number of related changes in this commit:

1. A script that generates the 'match' tests has been reintroduced.
2. The regex-dna shootout benchmark has been updated.
3. Running `cargo test` on the `regex` crate does not require
   `regex_macros`.
4. The documentation has been updated to use `Regex::new(...).unwrap()`
   instead of `regex!`. The emphasis on using `regex!` has been reduced,
   and a note about its unavailability in Rust 1.0 beta/stable has been
   added.
5. Updated Travis to test both `regex` and `regex_macros`.


BurntSushi:
cc @alexcrichton

BurntSushi:
I moved all of the tests into `regex_macros/tests`. The `Cargo.toml` for `regex` points to `regex_macros/tests/test_dynamic.rs`, which defines a macro `regex!`. This lets us reuse all of the tests. :-)

.travis.yml
```diff
@@ -4,6 +4,8 @@ script:
   - cargo build --verbose
   - cargo test --verbose
   - cargo doc
+  - cd regex_macros
+  - cargo test --verbose
```
alexcrichton: This could also be replaced with `cargo test --manifest-path=regex_macros/Cargo.toml` (avoiding the `cd`)

Cargo.toml
```diff
@@ -13,12 +12,14 @@ An implementation of regular expressions for Rust.
 """
 
 [[test]]
-path = "tests/mod.rs"
+path = "regex_macros/tests/test_dynamic.rs"
+name = "all"
+
+[[bench]]
 name = "all"
+path = "regex_macros/tests/bench_dynamic.rs"
```
alexcrichton: Perhaps this benchmark could be placed in the `benches` directory instead of the `tests` directory?

regex_macros/Cargo.toml
```diff
@@ -14,6 +13,23 @@ An implementation of statically compiled regular expressions for Rust.
...
+path = "tests/test_native.rs"
+name = "all"
+
+[[test]]
+path = "benches/shootout-regex-dna.rs"
+name = "shootout-regex-dna"
+
+[[bench]]
+name = "all"
+path = "tests/bench_native.rs"
```
alexcrichton: (same s/tests/benches/ thought as above here)

.travis.yml
alexcrichton: Ah and with benchmarks to run we may want to consider running `cargo bench` here as well (just to make sure things compile)

alexcrichton:
Looks good to me! Just a few questions about organization and otherwise I'm fine merging.

.travis.yml
BurntSushi: Neat trick! Done and done.

Cargo.toml
BurntSushi: Yup, makes sense! Done.

regex_macros/Cargo.toml
BurntSushi: Done.

BurntSushi:
Great, thank you for the review! I think I've addressed everything. I'll merge once Travis gives the all clear.

PR #44, opened 2015-02-28
Implement the new `Pattern` API on `Regex`.

BurntSushi:
This probably suggests some future work too. A lot of the methods implemented in `src/re.rs` can probably be implemented with the `StrExt` methods now.

There's also the question of whether we want to remove methods on `Regex` that are available in `StrExt`, but I think I'd prefer not to. It's nice to have a comprehensive API in one place, and it gives users the option to call these methods on a `Regex` type directly.

cc @kimundi

### DIFF
--- a/regex_macros/tests/tests.rs
+++ b/regex_macros/tests/tests.rs
@@ -8,6 +8,8 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use std::str::{Pattern, Searcher};
+use std::str::SearchStep::{Match, Reject, Done};
 use regex::{Regex, NoExpand};
 
 #[test]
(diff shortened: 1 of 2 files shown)


regex_macros/tests/tests.rs
```diff
@@ -96,6 +98,68 @@ fn empty_match_unicode_captures_iter() {
...
+                }
+            }
+            assert_eq!(got_steps, $expect_steps);
+        }
+    );
+}
+
+searcher!(searcher_empty_regex_empty_haystack, r"", "", Match(0, 0));
+searcher!(searcher_empty_regex, r"", "ab",
+          Match(0, 0), Reject(0, 1), Match(1, 1), Reject(1, 2), Match(2, 2));
```
Kimundi: Heh, I think I got this one wrong in libcore myself.

Kimundi:
Looks good! 

Ideally `Reject`s would be discovered along the way, rather than just after-the-fact of a match, but I guess thats kinda hard/impossible with the way `Regex` is implemented.

BurntSushi:
> Ideally Rejects would be discovered along the way, rather than just after-the-fact of a match, but I guess thats kinda hard/impossible with the way Regex is implemented.

Yup. The regex machinery only knows how to report the location of the next match, and it doesn't get much more granular than that.

Thanks for the review!

PR #45, opened 2015-03-01
Fixed changes to MacResult implementation

faineance:
As broken in https://travis-ci.org/rust-lang/regex/builds/52576711


BurntSushi:
Thanks! I've uploaded this fix to crates.io in `0.1.10`.

PR #47, opened 2015-03-01
Added new feature to iterate over named groups

### LINKED ISSUE #46: [feature request] Feature to iterate over named groups
It could be useful to be able to iterate over named groups from a regex capture.

Something like:

```rust
let re = regex!(r"(?P<id>\d{2})(?P<name>\w+)");
let caps = re.captures("12david").unwrap();
for (k, v) in caps.iter_named() {
    println!("group name: {0}, value: {1}", k, v);
}
```

PS: I'm new to contributing the rust project and open source in general, so I'm sorry if I'm not following standard procedures.

havarnov:
Added new feature to the Captures struct to iterate
over the named groups.

This pull request resolves #46.

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -801,12 +801,12 @@ pub struct SubCaptures<'t> {
 }
 
 impl<'t> Iterator for SubCaptures<'t> {
-    type Item = &'t str;
+    type Item = Option<&'t str>;
 
-    fn next(&mut self) -> Option<&'t str> {
+    fn next(&mut self) -> Option<Option<&'t str>> {
         if self.idx < self.caps.len() {
             self.idx += 1;
-            Some(self.caps.at(self.idx - 1).unwrap_or(""))
+            Some(self.caps.at(self.idx - 1))
         } else {
             None
         }
--- a/src/re.rs
+++ b/src/re.rs
@@ -10,6 +10,7 @@
 
 use std::borrow::{IntoCow, Cow};
 use std::collections::HashMap;
+use std::collections::hash_map::Iter;
 use std::fmt;
 use std::str::{Pattern, Searcher, SearchStep};
 
@@ -745,6 +746,16 @@ impl<'t> Captures<'t> {
         SubCapturesPos { idx: 0, caps: self, }
     }
 
+    /// Creates an iterator of all named groups as an tuple with the group
+    /// name and the value. The iterator returns these values in arbitrary
+    /// order.
+    pub fn iter_named(&'t self) -> SubCapturesNamed<'t> {
+        match self.named {
+            Some(ref n) => SubCapturesNamed { caps: self, inner: Some(n.iter()) },
+            None => SubCapturesNamed { caps: self, inner: None }
+        }
+    }
+
     /// Expands all instances of `$name` in `text` to the corresponding capture
     /// group `name`.
     ///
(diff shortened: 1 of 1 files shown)


src/re.rs
```diff
@@ -745,6 +746,16 @@ impl<'t> Captures<'t> {
...
+    /// Creates an iterator of all named groups as an tuple with the group
+    /// name and the value. The iterator returns these values in arbitrary
+    /// order.
+    pub fn iter_named(&'t self) -> SubCapturesNamed<'t> {
+        match self.named {
+            Some(ref n) => SubCapturesNamed { caps: self, inner: Some(n.iter()) },
+            None => SubCapturesNamed { caps: self, inner: None }
+        }
+    }
+
```
BurntSushi: I think this could probably be written more concisely:

```rust
SubCapturesNamed { caps: self, inner: self.named.as_ref().map(|n| n.iter()) }
```

src/re.rs
```diff
@@ -829,6 +840,31 @@ impl<'t> Iterator for SubCapturesPos<'t> {
...
+pub struct SubCapturesNamed<'t>{
+    caps: &'t Captures<'t>,
+    inner: Option<Iter<'t, String, usize>>,
+}
+
+impl<'t> Iterator for SubCapturesNamed<'t> {
+    type Item = (&'t str, &'t str);
+
+    fn next(&mut self) -> Option<(&'t str, &'t str)> {
+        match self.inner {
```
BurntSushi: Probably something similar here:

```rust
match self.inner.as_mut().map(|it| it.next()) {
    Some((name, pos)) => ...,
    None => None,
}
```

src/re.rs
```diff
@@ -829,6 +837,26 @@ impl<'t> Iterator for SubCapturesPos<'t> {
...
+    caps: &'t Captures<'t>,
+    inner: Option<Iter<'t, String, usize>>,
+}
+
+impl<'t> Iterator for SubCapturesNamed<'t> {
+    type Item = (&'t str, &'t str);
+
+    fn next(&mut self) -> Option<(&'t str, &'t str)> {
+        match self.inner.as_mut().map(|it| it.next()).unwrap_or(None) {
+            Some((name, pos)) => Some((name, self.caps.at(*pos).unwrap_or(""))),
```
BurntSushi: I didn't catch this the first time around, but given my comments in https://github.com/rust-lang/regex/issues/46#issuecomment-76642886, I think the iterator should probably return `Option<Option<(&'t str, &'t str)>>`. i.e., remove `unwrap_or("")`. Adding the `unwrap_or` in the iterator prevents the client from being able to differentiate between an empty matching capture group and a non-matching capture group.

This only slightly impacts the ergonomics of the iterator. e.g., you can do `caps.iter_name().filter_map(|x|x)` to skip non-matching captures of `caps.iter_named().map(|x| x.unwrap_or(""))` to get the behavior you have here.

havarnov:
Is it normal practice to "squash" the code review commits together? I.e. squash HEAD~1 to HEAD~3?

BurntSushi:
Yup, that's fine. You'll have to force push, but that's OK.

BurntSushi:
OK, this  looks great, thank you.

I'm going to work on coming up with a better design for captures though, so stay tuned. :-)

Issue #50, opened 2015-03-04
HEAD rust fails to compile


steveklabnik:
Sorry for not writing a PR, but my jetlagged brain isn't working for some reason. Seems easy:

```
   Compiling regex v0.1.16 (file:///home/steve/src/regex)
src/parse.rs:661:26: 661:41 error: mismatched types:
 expected `&'static [(&'static str, &'static &'static [(char, char)])]`,
    found `&'static [(&'static str, &'static [(char, char)])]`
(expected &-ptr,
    found slice) [E0308]
src/parse.rs:661         match find_class(UNICODE_CLASSES, name.as_slice()) {
                                          ^~~~~~~~~~~~~~~
error: aborting due to previous error
Could not compile `regex`.

```

gchp:
I fixed this in my fork, however it still fails to build:

```
rustc src/lib.rs --crate-name regex --crate-type lib -g -C metadata=e4601a68be576f97 -C extra-filename=-e4601a68be576f97 --out-dir /home/gchp/github.com/gchp/regex/target --emit=dep-info,link -L dependency=/home/gchp/github.com/gchp/regex/target -L dependency=/home/gchp/github.com/gchp/regex/target/deps
[2]    19122 illegal hardware instruction (core dumped)  rustc src/lib.rs --crate-name regex --crate-type lib -g -C  -C  --out-dir   -
```

My changes are here: https://github.com/gchp/regex/commit/f1a13c8fbab94f2833dc3fa0c29aa3193be4647a

BurntSushi:
`rustc` dumped core? That doesn't seem like a problem with `regex`...

gchp:
@BurntSushi yup! I'm not sure how to get more information than that though, to create an issue against rust.

Should I submit those changes in a PR anyway, or wait until it actually builds?

BurntSushi:
Yeah, please submit a PR. I'm getting the same behavior here too. (The latest nightly is out.)

gchp:
Ok, done!

laumann:
I'm having the exact same problem with `rustc`, getting the illegal hardware instruction. Should we open an issue on rust?

PR #51, opened 2015-03-04
Fixed compilation errors

gchp:


gchp:
I currently can't compile regex. See this comment for info: https://github.com/rust-lang/regex/issues/50#issuecomment-77153302

BurntSushi:
I get the same over here. I ran it under `valgrind`:

```
 valgrind rustc src/lib.rs --crate-name regex --crate-type lib -g -C metadata=64952606b89c7990 -C extra-filename=-64952606b89c7990 --out-dir /home/andrew/data/projects/rust/regex/target --emit=dep-info,link -L dependency=/home/andrew/data/projects/rust/regex/target -L dependency=/home/andrew/data/projects/rust/regex/target/deps
```

and got this: https://gist.github.com/087f0afda9e422cd44b1 --- bug in rustc? This is all the investigation I can do at the moment though. :-(

BurntSushi:
@alexcrichton ping. :-)

alexcrichton:
Hm, fascinating!  Looks like rust-lang/rust#23032 has the segfault issue covered and the changes in #52 don't cause a segfault, so I'll merge that for now.

alexcrichton:
Thanks regardless!

PR #52, opened 2015-03-04
Fix compilation error + Static to const

laumann:
This also includes @gchp's changes, and changes `static` modifiers on ASCII classes to `const`.

All tests pass.


alexcrichton:
Thanks!

PR #56, opened 2015-03-14
Drop dependence on the unicode crate

alexcrichton:
This commit pulls in the script to generate tables.rs in the main distribution
and strips it down to just the bare bones necessary for regexes (which is still
quite a lot!). The script was used to generate a `unicode.rs` file which
contains all the data needed from the libunicode crate.

Eventually we hope to provide libunicode in some form on crates.io or perhaps
stabilize it in the distribution itself, but for now it's not so bad to vendor
the dependency (which doesn't change much) and it's required to get libregex
building on stable Rust.


BurntSushi:
LGTM. :D thanks!

PR #57, opened 2015-03-16
Fix Noncharacter_Code_Point_table

dscorbett:
rust-lang/rust#22690 describes a bug in `unicode.py`, where U+10FFFE and U+10FFFF are not put in `Noncharacter_Code_Point_table`. The bug is present in the version of the script recently copied to `regex`. This commit fixes the bug by parsing Unicode data files in accordance with [UAX #44](http://www.unicode.org/reports/tr44/#Data_Fields).

### DIFF
--- a/scripts/unicode.py
+++ b/scripts/unicode.py
@@ -176,8 +176,8 @@ def format_table_content(f, content, indent):
 def load_properties(f, interestingprops):
     fetch(f)
     props = {}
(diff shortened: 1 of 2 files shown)


BurntSushi:
Excellent, thank you!

BurntSushi:
I've uploaded this patch to crates.io in `0.1.20`.

PR #58, opened 2015-03-17
Move str::Pattern impl behind a feature

alexcrichton:
It's not clear whether this API of the standard library will be stable at 1.0 so
in the meantime this implementation can be put behind a Cargo feature so users
of the crate are not required to use nightly Rust by default. The feature will
still be available to users of nightly Rust, however.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 script:
   - cargo build --verbose
   - cargo test --verbose
+  - cargo test --verbose --features pattern
   - cargo doc
   - cargo bench --verbose
   - cargo test --verbose --manifest-path=regex_macros/Cargo.toml
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,6 @@ bench = true
 
 [dev-dependencies]
 rand = "0.1"
+
+[features]
+pattern = []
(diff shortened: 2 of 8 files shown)


alexcrichton:
r? @BurntSushi 

(I'd hold off merging until travis finishes though)

.travis.yml
```diff
@@ -3,6 +3,7 @@ sudo: false
 script:
   - cargo build --verbose
   - cargo test --verbose
+  - cargo test --verbose --features pattern
   - cargo doc
```
sfackler: Docs should probably be built with the feature as well.

BurntSushi:
:+1: I like the `searcher_expr!` trick. :-)

.travis.yml
alexcrichton: I was a little bit torn on that, I'm not sure whether the documentation should describe the "vanilla crate" or the crate with all features enabled. For example if the documentation claims the impl but then I can't use it it may be confusing. I figured that for now the failure mode of "I didn't realize I could use it even though I could" may be more reasonable.

BurntSushi: I agree. The docs also used to emphasize and use the `regex!` macro everywhere, but I removed almost all of it and de-emphasized use of `regex!` in the crate docs.

PR #60, opened 2015-03-22
Expose SubCapturesNamed to docs

Ryman:
Otherwise the type returned from iter_named isn't hyperlinked.


BurntSushi:
Great, thank you!

PR #63, opened 2015-03-27
Correct the documentation for \d.

### LINKED ISSUE #62: Regex documentation on \d possibly misleading
Not sure if it's just because I don't know enough about character classes, but when I read:

```
\d          Perl character class ([0-9])
\D          Negated Perl character class ([^0-9])
```

(Source: http://doc.rust-lang.org/regex/regex/index.html#matching-one-character)

I thought this meant that `\d` would *only* match the ASCII characters 0 through 9, but in fact it'll match other Unicode digits too:

```rust
#![feature(plugin)]
#![plugin(regex_macros)]
extern crate regex;

fn main() {
    let re = regex!(r"\d");
    println!("{}", re.is_match("٩")); // prints "true"
}
```

Maybe the wording should be changed?

ucarion:
Fixes #62 .

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -210,8 +210,8 @@
 //! [xyz]       A character class matching either x, y or z.
 //! [^xyz]      A character class matching any character except x, y and z.
 //! [a-z]       A character class matching any character in range a-z.
-//! \d          Perl character class ([0-9])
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks!

PR #70, opened 2015-04-14
implement FromStr for Regex

blaenk:
As discussed with @BurntSushi on IRC.

Let me know if something is missing!


alexcrichton:
Thanks!

PR #73, opened 2015-04-16
optimize generated tables ; clean up unicode.py

kwantam:
There was an easy opportunity to better optimize the tables generated
by unicode.py. Not sure why I didn't catch this long ago, but in any
case now the tables are substantially smaller and should maybe improve
performance slightly.

There was also some dead code sitting in unicode.py that I pulled out.


huonw:
Could you give a high-level summary of how the tables changed? (github isn't showing the diff and I'm finding it hard to visualise from just the python changes.)

BurntSushi:
My suspicion is that this script was once used to generate other Unicode data that wasn't used by `regex` (e.g., east asian character widths?). So this is probably trimming it back down?

@kwantam I'm still interested in a high level summary though. I think I wrote the original version of this over a year ago for `regex` stuff only, but I no longer understand the code.

kwantam:
Quick background: the property tables that `unicode.py` generates are sorted arrays of non-overlapping ordered pairs of `char`s, `(x, y)`, x<=y. When matching against a certain property (say, with `\p{Alphabetic}`), the regex code does a binary search in the corresponding table to identify whether there is a pair that contains a given character.

Now, in some cases, the tables that were previously generated contained ordered pairs that should have been merged together. For example, `Alphabetic_table` previously contained a sequence like this:

    ...
        ('\u{d8}', '\u{f6}'), ('\u{f8}', '\u{1ba}'), ('\u{1bb}', '\u{1bb}'),
        ('\u{1bc}', '\u{1bf}'), ('\u{1c0}', '\u{1c3}'), ('\u{1c4}', '\u{293}'),
        ('\u{294}', '\u{294}'), ('\u{295}', '\u{2af}'), ('\u{2b0}', '\u{2c1}'),
    ...

Note that, for example, `('\u{f8}', '\u{1ba}'), ('\u{1bb}', '\u{1bb}')` can be collapsed into a single range, because 0x1bb = 0x1ba + 1. In fact, the above sequence, in the new version of the table, is just

    ...
        ('\u{d8}', '\u{f6}'), ('\u{f8}', '\u{2c1}')
    ...

In case you're wondering, the reason this was happening is because of the way that the Unicode-provided tables were formatted. The function that imports these tables, `load_properties()`, was written with the assumption that all sequential ranges in the tables were collapsed, but this isn't the case. The very small addition, among all the cuts in `unicode.py`, that fixes this, is at the end of the `load_properties()`, function:

    for prop in props:
        props[prop] = group_cat(ungroup_cat(props[prop]))

alexcrichton:
Nice! Thanks @kwantam!

PR #77, opened 2015-04-19
Map char ranges for case-insensitivity at parse time.

### LINKED ISSUE #75: Binary search is used with incorrectly-sorted array
The “dynamic” matching of `CharClass` [uses a binary search](https://github.com/rust-lang/regex/blob/399758aeae3dcab382b0af5fa9964c1e32066dda/src/vm.rs#L236) within `char` ranges, which relies on the input being sorted. The input is indeed sorted in its “natural” order, but the comparison function in case-insensitive mode uses a different order.

This leads to incorrect results:

```rust
assert!(Regex::new(r"(?i)[a_]+$").unwrap().is_match("A_"));
```

The above fails, because `_` in ASCII is between upper-case and lower-case letters. The comparison function maps the `'a'..'a'` range to its upper case `'A'..'A'`, which has a different order relative to `'_'..'_'`.

Compare with e.g. the code below, which succeeds.
```rust
assert!(Regex::new(r"(?i)[a=]+$").unwrap().is_match("A="));
```

The comparison function has a `FIXME` to move the case mapping outside of it and have the `Vec` of ranges be already mapped. I assume this was intended for performance, but I think it’ll also fix this issue. I’ll try it.

SimonSapin:
Fix #75.


SimonSapin:
Included in #78.

PR #80, opened 2015-04-22
replace strings index fix

nathanfaucett:
allow replace strings with adjacent text like "$1ies" "$1s"


alexcrichton:
r? @BurntSushi 

Could you also add a test for this?

BurntSushi:
@nathanfaucett With a test I think this looks OK. Thanks!

BurntSushi:
Thanks!

Issue #66, opened 2015-04-01
implement a DFA matcher


BurntSushi:
One of the reasons why RE2/C++ is so fast is because it has two implementations of regex matching: a limited DFA matcher (no sub-capture support) and a full NFA simulation. This crate has the latter, but not the former.

Adding a DFA matcher should be an implementation detail and shouldn't require any public facing changes.

This is a pretty involved project. I hope to find time to do this some day, but if someone else wants to tackle it, I'd be happy to help mentor it. (Part of this will be figuring out how to handle the `regex!` macro. Do we replicate the DFA there too like we do the NFA?)

WillEngler:
I'll do it!

@BurntSushi recommends that I look into an [existing DFA engine](https://github.com/carllerche/automaton) by @carllerche.

I'll be starting by getting familiar with the code in this repo and [RE2](https://github.com/google/re2).

carllerche:
Feel free to ping me on IRC or wherever if you want to talk about [carllerche/automaton](https://github.com/carllerche/automaton)

WillEngler:
Here's a summary of some research I've done and my intended next steps.

### TL;DR

The algorithms that @carllerche uses in [carllerche/automaton](https://github.com/carllerche/automaton) make a different time/space tradeoff than that made by RE2 and rust-lang/regex so far. For now, I recommend that we keep modeling rust-lang/regex after RE2 and use the same method of lazy DFA construction used in RE2 rather than incorporate Carl's work. However, using Carl's work to construct a DFA at compile time with the `regex!` macro could be a _killer_ future enhancement.

### A Comparison of rust-lang/regex, RE2, and carllerche/automaton

I consulted [The Dragon Book](http://en.wikipedia.org/wiki/Compilers:_Principles,_Techniques,_and_Tools) and [Russ Cox's blog series](https://swtch.com/~rsc/regexp/) on implementing RE2 for background on algorithms for constructing and simulating automata. I'll refer to algorithms by numbers given to them in the Dragon Book for anyone who wants to take a closer look.

[...]

PR #81, opened 2015-04-23
Implement Eq for Regex

yorhel:
Note that this is just a comparison of the original strings, it's still possible that two different strings result in the same matching behaviour.

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -160,7 +160,7 @@ impl fmt::Debug for Regex {
 }
 
 /// Equality comparison is based on the original string. It is possible that different regular
-/// expressions have the same matching behaviour, but are still compared inequal.  For example,
+/// expressions have the same matching behavior, but are still compared unequal.  For example,
 /// `\d+` and `\d\d*` match the same set of strings, but are not considered equal.
 impl PartialEq for Regex {
     fn eq(&self, other: &Regex) -> bool {
--- a/src/re.rs
+++ b/src/re.rs
@@ -159,6 +159,9 @@ impl fmt::Debug for Regex {
     }
 }
 
+/// Equality comparison is based on the original string. It is possible that different regular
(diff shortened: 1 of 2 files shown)


BurntSushi:
I don't think there's any feasible way to detect whether two arbitrary expressions match the same set of strings, so I'm OK with this definition.

src/re.rs
```diff
@@ -159,6 +159,14 @@ impl fmt::Debug for Regex {
     }
 }
 
+impl PartialEq for Regex {
```
BurntSushi: @yorhel Could you please put a doc comment on this impl explaining the definition of equality used? (Something like what you said in your initial comment is just fine.)

BurntSushi:
I have one nit, but other than that, looks good!

yorhel:
Doc added. Thanks for the feedback!

BurntSushi:
Awesome. Good docs! Thanks!

huonw:
> I don't think there's any feasible way to detect whether two arbitrary expressions match the same set of strings, so I'm OK with this definition.

I looked into it a bit, and yes, it seems to be non-trivial: http://cs.stackexchange.com/q/12267 (that said, it is stated that one can test equality by computing minimal DFAs and comparing).

BurntSushi:
@huonw Yeah, when we get DFAs (#66) then maybe we can expose that sort of operation, but it probably shouldn't be in an impl of `PartialEq`, which most would expect to be fast. (Computing the DFA can take quite a lot of time!)

Even better, maybe @carllerche [automaton](https://github.com/carllerche/automaton) package can do it for us.

carllerche:
Automaton currently computes the minimal DFA using hopcroft's algorithm. Of course, there could always be bugs and the lib is still under heavy development :smile:

PR #82, opened 2015-05-02
Add usage docs in-crate

steveklabnik:
People constantly miss that they need to use Cargo, so add those instructions in-crate as well as to the README.

Regex is a bit better than most, but would still like to have it inline


BurntSushi:
Excellent, thank you! :-)

Issue #79, opened 2015-04-19
Optimize case_fold_and_combine_ranges?


SimonSapin:
I haven’t timed it, but the `case_fold_and_combine_ranges` function introduced in #78 is *probably* slow for large sets of character ranges like `\p{Lu}`. (That is, slow to parse/compile a regex.) It can probably be improved to not consider individual `char`s when we can determine that none in a given range (or subrange) is affected by case folding.

BurntSushi:
Yeah, I saw that. I'm not terribly concerned with parsing/compiling performance. I'll take a pass over it in my parser rewrite. (Which I'm doing as part of #29.)

BurntSushi:
I think I have at least this optimization implemented in the parser rewrite: https://github.com/rust-lang/regex/blob/new-parser/regex_syntax/src/lib.rs#L148

SimonSapin:
Nice! I feel it might be more clever still, but this seems to be a nice compromise of optimizing some (most common?) cases while keeping it relatively simple. Have you measured it before/after and against case-sensitive for large classes like `\p{Lu}`?

BurntSushi:
Not yet. Once I finish the rewrite, I'll add some benchmarks and maybe do some profiling to see if I can pick off low hanging fruit.

But yeah, I feel like the entire way classes are constructed (particularly canoncalizing them) could be more clever. The code is pretty short/clear now though. :)

BurntSushi:
Weird that github didn't close this. Fixed by #87.

PR #87, opened 2015-05-25
Rewrite parser as part of new regex-syntax crate.

### LINKED ISSUE #29: [feature request] Expose libregex's parsing/compiling internals
<a href="https://github.com/andrew-d"></img></a> **Issue by [andrew-d](https://github.com/andrew-d)**
_Thursday Nov 06, 2014 at 19:54 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/18710_

_This issue was labelled with:  in the Rust repository_

----

I was looking at implementing something similar to [this](http://swtch.com/~rsc/regexp/regexp4.html) - a trigram-index-aided search.  I'd rather not reproduce the code necessary to parse the regex, considering it already lives in libregex.  It'd be nice if the parsing/compiling was exposed for use - perhaps similar to how Go does it with their `regexp` and `regexp/syntax` [packages](http://golang.org/pkg/regexp/).

cc @burntsushi

### LINKED ISSUE #67: impose a limit on the size of a compiled regex program
From the docs:

> Currently, there are no counter-measures in place to prevent a malicious user from writing an expression that may use a lot of resources. One such example is to repeat counted repetitions: ((a{100}){100}){100} will try to repeat the a instruction 100^3 times. Essentially, this means it's very easy for an attacker to exhaust your system's memory if they are allowed to execute arbitrary regular expressions. A possible solution to this is to impose a hard limit on the size of a compiled expression, but it does not yet exist.

The conclusion of this is that regexes specified by a user cannot be blindly trusted, since they can trivially exhausted all memory on your system. We can fix this by imposing some limit on the size of a regex program. (In fact, this probably has to be a limit on the size of a regex AST, which will need to be checked during construction.)

### LINKED ISSUE #69: Underscore in replacement string treated as backspace
Using a `_` in the replacement string of `Regex::replace` leads to unexpected behaviour.  The `_` seems to be treated as a backspace.  The documentation should either make mention of this, or this seems to be a bug.

``` rust
#[test]
fn replacement_with_underscore() {
    let re = regex!(r"(.)(.)");
    let s1 = re.replace("ab","$1-$2");
    let s2 = re.replace("ab","$1_$2");
    assert_eq!("a-b", &s1);
    assert_eq!("a_b", &s2); // Fails here "a_b" != "b"
}
```

### LINKED ISSUE #84: Attempting to parse {2} causes panic
```rust
use regex::Regex;

extern crate regex;


fn main() {
    let _ = Regex::new("{2}");
}
```

```
/t/test (master|…) $ cargo run
     Running `target/debug/test`
thread '<main>' panicked at 'called `Option::unwrap()` on a `None` value', /Users/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-mac/build/src/libcore/option.rs:362
An unknown error occurred

To learn more, run the command again with --verbose.
```

Was found using https://github.com/kmcallister/afl.rs 👍

BurntSushi:
This commit introduces a new `regex-syntax` crate that provides a
regular expression parser and an abstract syntax for regular
expressions. As part of this effort, the parser has been rewritten and
has grown a substantial number of tests.

The `regex` crate itself hasn't changed too much. I opted for the
smallest possible delta to get it working with the new regex AST.
In most cases, this simplified code because it no longer has to deal
with unwieldy flags. (Instead, flag information is baked into the AST.)

Here is a list of public facing non-breaking changes:

* A new `regex-syntax` crate with a parser, regex AST and lots of tests.
  This closes #29 and fixes #84.
* A new flag, `x`, has been added. This allows one to write regexes with
  insignificant whitespace and comments.
* Repetition operators can now be directly applied to zero-width
  matches. e.g., `\b+` was previously not allowed but now works.
  Note that one could always write `(\b)+` previously. This change
  is mostly about lifting an arbitrary restriction.

And a list of breaking changes:

* A new `Regex::with_size_limit` constructor function, that allows one
  to tweak the limit on the size of a compiled regex. This fixes #67.
  The new method isn't a breaking change, but regexes that exceed the
  size limit (set to 10MB by default) will no longer compile. To fix,
  simply call `Regex::with_size_limit` with a bigger limit.
* Capture group names cannot start with a number. This is a breaking
  change because regexes that previously compiled (e.g., `(?P<1a>.)`)
  will now return an error. This fixes #69.
* The `regex::Error` type has been changed to reflect the better error
  reporting in the `regex-syntax` crate, and a new error for limiting
  regexes to a certain size. This is a breaking change. Most folks just
  call `unwrap()` on `Regex::new`, so I expect this to have minimal
  impact.

Closes #29, #67, #69, #79, #84.

[breaking-change]


BurntSushi:
cc @alexcrichton @huonw @andrew-d

I think the thing I am most concerned about here is the definition of the `Error` type. It is a publicly exposed `enum` that I could envision changing in the future. (For example, if the `regex-syntax` crate grew to accept a superset of the syntax supported by `regex`, then there would need to be an additional error for rejecting unsupported syntax.) My instinct is that I should not expose the error as an enum and instead provide methods. (e.g., `fn too_big(&self) -> Option<usize>` and `fn syntax_error(&self) -> Option<&regex_syntax::Error>`.)

Also, part of the enum exposes the `regex_syntax::Error` type, which could also evolve. I think this means that if there's a breaking change in `regex_syntax::{Error, ErrorKind}`, then the major version number of `regex` would also need to be bumped. Not sure how to deal with this one.

Cargo.toml
```diff
@@ -21,8 +21,16 @@ path = "regex_macros/benches/bench_dynamic.rs"
 test = false
 bench = true
 
+[dependencies.regex-syntax]
+path = "regex_syntax"
+version = "*"
```
alexcrichton: I personally prefer to write out this section like so nowadays:

```toml
[dependencies]
regex-syntax = { path = "regex_syntax", version = "*" }
```

Also, perhaps this could have a more strict version requirement than `*`? And finally, perhaps the folder could be called `regex-syntax` to match the crate name?

.gitignore
```diff
@@ -2,4 +2,7 @@
 /Cargo.lock
 /regex_macros/target
 /regex_macros/Cargo.lock
+/regex_syntax/target
+/regex_syntax/Cargo.lock
```
alexcrichton: It's probably also fine to just ignore everything called `target` and `Cargo.lock` here

Cargo.toml
BurntSushi: > And finally, perhaps the folder could be called regex-syntax to match the crate name?

Happy to oblige. I originally had that, but it looks very weird next to `regex_macros`, which I think is stuck with an underscore? :-(

:+1: to everything else!

alexcrichton:
This looks awesome, thanks so much @BurntSushi!

> And a list of breaking changes:

These all seem minor enough that I would be fine releasing this version as a minor update to this library (not a major version change).

> I think the thing I am most concerned about here is the definition of the Error type. It is a publicly exposed enum that I could envision changing in the future.

I personally like having an enum for errors like this, and I think that they should be considered "defacto extensible" in the sense that although it's a breaking change to add a variant, I would not consider it one. You could add a `#[doc(hidden)]` variant named something like `__DoNotMatchMeIfYouWishToNotBreak` which may also deter exhaustive matches.

> Also, part of the enum exposes the regex_syntax::Error type, which could also evolve. I think this means that if there's a breaking change in regex_syntax::{Error, ErrorKind}, then the major version number of regex would also need to be bumped. Not sure how to deal with this one.

Hm, this is interesting! I would believe that there's definitely a good deal of API leakage between crates in terms of a major version bump of your deps may require a major version bump for you, but I think with a dependency like `0.1.0` on `regex-syntax` that this crate will be safe. That basically means that regex 0.1.x will use the 0.1.x version of syntax, and it can evolve to a more updated syntax later if necessary.

Cargo.toml
alexcrichton: Yeah unfortunately we can't switch underscore-to-dash after the fact just yet :( 

I would like to enable this kind of feature on crates.io though as I suspect it comes up more often than not!

Cargo.toml
```diff
@@ -21,8 +21,16 @@ path = "regex_macros/benches/bench_dynamic.rs"
...
+
 [dev-dependencies]
 rand = "0.3"
 
 [features]
 pattern = []
+
+[profile.bench]
+opt-level = 3
```
alexcrichton: I think that 3 is the default, so perhaps this could just be specified to set `lto = true`

Cargo.toml
BurntSushi: OK, hypens it is!

BurntSushi:
I think I've addressed everything. Just followed all of your suggestions. :-) I can merge once Travis gives the OK!

alexcrichton:
Sounds good to me!

frewsxcv:
Welp, time to AFL :)

BurntSushi:
Haha. Do your worst. This time, quickcheck has my back. :P

Issue #88, opened 2015-05-28
Allow escaping `#` when using the x flag


killercup:
It's nothing really important, but it would be nice if one could escape a literal `#` so it doesn't start a comment when using the `x` flag.

The current workaround is using the ASCII literal `\x23`, but something like `\#` would be clearer.

Workaround example:

```rust
let markdown_headline = regex!(r"(?x)
    ^
    (?P<level>[\x23]+)  # A bunch of hash symbols
    \s
    (?P<title>.+)       # Title
    $
");
```

BurntSushi:
Seems reasonable! I think adding `#` to [`is_punct`](https://github.com/rust-lang/regex/blob/master/regex-syntax/src/parser.rs#L1059) should do the trick.

killercup:
Cool! I'll give it a try.

BurntSushi:
@killercup If you can write a test for it that passes, then that's great. But on second thought, it might be a bit trickier than that. The problem is that insignificant whitespace/comments are handled at a layer below the parser: https://github.com/rust-lang/regex/blob/master/regex-syntax/src/parser.rs#L804-L818 So I think the mini-state machine in the `skip` method will need to be made aware of the escape. (But `#` should still be part of `is_punct`.)

If you get bogged down, no worries and I can take a crack at it later.

PR #89, opened 2015-05-28
Add `\#` escaping

killercup:
Thanks for your suggestions in #88, @BurntSushi!

I added some tests, but I'm not sure they're covering all edge cases.

### DIFF
--- a/regex-syntax/src/parser.rs
+++ b/regex-syntax/src/parser.rs
@@ -801,14 +801,25 @@ impl<'a> Chars<'a> {
         }
     }
 
+    /// Skip line comments and whitespace
     fn skip(&mut self) {
         if !self.ignore_space { return; }
         while self.cur < self.chars.len() {
+            // Handle escaping of `#`, i.e. don't start a comment with `\#`.
+            if !self.in_comment && (self.c() == '\\')
+                && ((self.cur + 1) < self.chars.len())
+                && (self.chars[self.cur + 1] == '#')
+            {
+                self.cur = checkadd(self.cur, 1);
+                break;
+            }
+
             if !self.in_comment && self.c() == '#' {
                 self.in_comment = true;
             } else if self.in_comment && self.c() == '\n' {
                 self.in_comment = false;
             }
+
             if self.in_comment || self.c().is_whitespace() {
                 self.cur = checkadd(self.cur, 1);
             } else {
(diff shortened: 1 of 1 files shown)


regex-syntax/src/parser.rs
```diff
@@ -1047,19 +1058,19 @@ fn rev_concat(mut exprs: Vec<Expr>) -> Expr {
     }
 }
 
-// Returns ture iff the given character is allowed in a capture name.
+// Returns true if the given character is allowed in a capture name.
```
BurntSushi: Hah. The "iff" actually isn't a typo. It means, "if and only if." It's leftover shorthand from getting a math degree. :P

regex-syntax/src/parser.rs
```diff
@@ -801,14 +801,25 @@ impl<'a> Chars<'a> {
...
     }
 
+    /// Skip line comments and whitespace
     fn skip(&mut self) {
         if !self.ignore_space { return; }
         while self.cur < self.chars.len() {
+            // Handle escaping of `#`, i.e. don't start a comment with `\#`.
+            if !self.in_comment && (self.c() == '\\')
+                && ((self.cur + 1) < self.chars.len())
+                && (self.chars[self.cur + 1] == '#')
```
BurntSushi: Arg. Get those redundant parens off my lawn!

regex-syntax/src/parser.rs
killercup: Oh, haven't seen that before. (Haven't read any math papers in English, though.) I'll change it to "if and only if".

regex-syntax/src/parser.rs
```diff
@@ -801,14 +801,25 @@ impl<'a> Chars<'a> {
         }
     }
 
+    /// Skip line comments and whitespace
     fn skip(&mut self) {
         if !self.ignore_space { return; }
         while self.cur < self.chars.len() {
+            // Handle escaping of `#`, i.e. don't start a comment with `\#`.
+            if !self.in_comment && (self.c() == '\\')
+                && ((self.cur + 1) < self.chars.len())
```
BurntSushi: Could you please use `checkadd(self.cur, 1)` here (and below)? I'm trying to be a bit paranoid about overflow in the parser. (Although I don't have any tests for it...)

regex-syntax/src/parser.rs
BurntSushi: Yeah, I can't recall it showing up much in papers. It's more like "I'm hand writing a proof and I'm going to gratuitously abuse shorthand."

regex-syntax/src/parser.rs
killercup: Your have your overflow paranoia, I have my parens. ;)

How much can I get rid of? Does `+` bind stronger than `<` which binds stronger than `&&`?

BurntSushi: Yup. Exactly.

BurntSushi: I'm usually not one to rely on memorization of precedence, but in these cases, I think if the precedence was anything else, you'd end up with a compile error.

killercup:
@BurntSushi, I've added a commit using more checked adds and less parens.

Btw, there are a lot of small functions in there. I hope they all get inlined.

BurntSushi:
Excellent! Thank you!

> Btw, there are a lot of small functions in there. I hope they all get inlined.

Yeah, not so concerned. There are other things I'm more concerned about affecting performance (like simplifying the AST and canonicalizing character classes). For another day!

BurntSushi:
OK, crates.io updated! (`regex-syntax 0.1.1`)

PR #102, opened 2015-07-22
Fix $$ -> $ when there are no expansions.

Ryanl92:
The docs mention that `$$` will produce a single `$` in the output however if there are no expansions in the replace string you still get `$$` in the output. Some test cases:

```
let re = Regex::new("world").unwrap();
println!("{}", re.replace_all("Hello world!", "$10"));  // Hello !
println!("{}", re.replace_all("Hello world!", "$$10"));  // Hello $$10!
println!("{}", re.replace_all("Hello world!", "\\$10"));  // Hello \!
let re = Regex::new(r"(!)").unwrap();
// This case does not change with this patch because of the expansions
println!("{}", re.replace_all("Hello world!", "$1 $$expand $1"));  // Hello world! $expand !
```

This patch should fix the second println without effecting the others.

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -599,10 +599,7 @@ impl Regex {
         if rep.no_expand().is_some() {
             // borrow checker pains. `rep` is borrowed mutably in the `else`
             // branch below.
-
-            // A replacement for a literal $ will be on this branch if that is
-            // the only replacement, so replace the $$ with a $.
-            let rep = rep.no_expand().unwrap().replace("$$", "$");
+            let rep = rep.no_expand().unwrap();
             for (i, (s, e)) in self.find_iter(text).enumerate() {
                 if limit > 0 && i >= limit {
                     break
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks for the bug report and patch!

I agree this is a problem, but I'm not sure this is the right way to fix it. I see two problems:

1. You've added an allocation on the fast path. (However, you're copying the replacement string which is usually small, so that's probably OK.)
2. This will break uses of `NoExpand`. Namely, `NoExpand("$$")` should result in `$$`, but this code will change it to `$`.

I think the bug here is that `&str` is not satisfying the contract of `no_expand` in the `Replacer` trait. My guess is that [this code](https://github.com/rust-lang/regex/blob/master/src/re.rs#L713) should be changed to scan for a `$` (instead of doing a regex search). If one exists, then return `None`. Otherwise, we can be confident that no special replacements are needed.

Ryanl92:
Right both points make sense, I'll address them both in the morning, I when
I ran `cargo test` I don't think there were any so I'll probably look into
some unit tests for this as well to catch any regressions as there seem to
be a few corner cases that I missed.

As I now understand we should consider $$ -> $ an expansion in `no_expand`?

BurntSushi:
More tests would be lovely. :-) And you're right, some test definitely *should* have failed as a result of this patch.

> As I now understand we should consider $$ -> $ an expansion in `no_expand`?

Yes. But to simplify things, we can return `None` if a `$` exists *anywhere* in the string. A `$` doesn't necessarily imply an expansion, but that's OK. The key is that if there is no `$` in the string, then there are no possible expansions of any kind. Namely, the replacement string can be used as-is.

Ryanl92:
I've updated the patch with your comments, and added a couple of tests for the cases we've mentioned in this pull request.

BurntSushi:
Poifect. I buy it. Thanks!

PR #104, opened 2015-07-31
Fix #31: Document the requirement to add regex_macros to Cargo.toml...

### LINKED ISSUE #31: Regex docs don't point people to crates.io dependency

TheJakeSchmidt:
Document the requirement to add regex_macros to Cargo.toml in order to use regex!().

Issue #31 was closed in February, but I think mistakenly, because it doesn't seem to have been fixed yet.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,8 +71,17 @@
 //! more optimization work put it into it currently, so it is faster than
 //! the `regex!` macro in most cases.
 //!
-//! To use the `regex!` macro, you must enable the `plugin` feature and import
-//! the `regex_macros` crate as a syntax extension:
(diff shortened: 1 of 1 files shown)


BurntSushi:
Excellent. Thank you!

Issue #107, opened 2015-08-03
Allow \<newline> to escape the newline away


golddranks:
It's often desirable to use raw strings when declaring regular expressions. When raw strings are used, the standard Rust escapes are not in effect.

The regex syntax defines its own escapes, which are especially catered for escaping the regex puncuation characters, and some other characters that are difficult to include in the source code as-they-are. Raw strings are useful, because we don't have to double-escape the regex escapes, and can express characters like " freely.

However, because we can't use Rust escapes, we can't escape newlines either, like we can with normal strings:

> Line-break characters are allowed in string literals. Normally they represent themselves (i.e. no translation), but as a special exception, when a U+005C character \ occurs immediately before the newline, the U+005C character, the newline, and all whitespace at the beginning of the next line are ignored. Thus a and b are equal:

As far as I know, the regex crate doesn't have syntax that'd allow splitting the expression to multiple lines with escapable newlines: if there are newlines, they stand for themselves.

It would be nice to have an escape that allows that; backslash followed by newline character. It would help splitting raw string regexes to multiple lines for clarity. It would also help to follow coding conventions with regards to line length.

What do you think?

golddranks:
Btw. currently backslash followed by newline just throws a parsing error, so I think that implementing that escape would be non-breaking change.

BurntSushi:
Is there a reason why the `x` flag isn't suitable? See an example here: http://doc.rust-lang.org/regex/regex/index.html#example:-replacement-with-named-capture-groups and documented here: http://doc.rust-lang.org/regex/regex/index.html#grouping-and-flags

golddranks:
Ah, I see, my premise that there isn't a way to ignore newline was wrong. Thanks for the info!

The x flag seems to provide the functionality I was missing - and more.

I'd still argue that supporting the backslash newline escape would be good, because it's simple and intuitive to use (in the sense that Rust strings, and many other languages support the same feature). I'd consider it as an additional ergonomic bonus - it's immediately clear what it does, and it's syntactically lightweight.

BurntSushi:
That's not a bad point. I don't feel too strongly about it either way.

golddranks:
I made a pull request that implements the feature. No pressure to accept it if there's some controversy about it, or if it needs further thought, wider acceptance, review etc.

golddranks:
[...]

PR #108, opened 2015-08-03
Implemented escaping newlines.

golddranks:
Implementation of feature suggestion mentioned in https://github.com/rust-lang/regex/issues/107 Tests and documentation included.

I think this is a nice, ergonomic bonus; the semantics are obvious and lightweight (and similar with the Rust strings newline escape). @BurntSushi said he doesn't have any strong feelings either way. What do others think?

It allows a case that was previously illegal, so not a breaking change.

### DIFF
--- a/regex-syntax/src/parser.rs
+++ b/regex-syntax/src/parser.rs
@@ -185,9 +185,15 @@ impl Parser {
             return Ok(try!(self.lit(c)));
         }
         match c {
-            '\n' => { self.bump(); Ok(Build::Expr(Expr::Empty)) }
+            '\n' => { self.bump();
+                while self.cur().is_whitespace() { self.bump(); };
+                Ok(Build::Expr(Expr::Empty))
+            }
             '\r' => { self.bump();
-                    if self.bump_if('\n') { Ok(Build::Expr(Expr::Empty)) }
+                    if self.bump_if('\n') {
+                        while self.cur().is_whitespace() { self.bump(); };
+                        Ok(Build::Expr(Expr::Empty))
+                    }
                     else { Err(self.err(ErrorKind::UnrecognizedEscape(c))) } },
             'a' => { self.bump(); Ok(try!(self.lit('\x07'))) }
             'f' => { self.bump(); Ok(try!(self.lit('\x0C'))) }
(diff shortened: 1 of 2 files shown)


golddranks:
Rebased on newest master and added the compatibility for

> Rust's existing line continuation drops the \n and all proceeding whitespace up to the first non-whitespace character.

regex-syntax/src/parser.rs
```diff
@@ -185,6 +185,16 @@ impl Parser {
...
+            '\n' => { self.bump();
+                while self.cur().is_whitespace() { self.bump(); };
+                Ok(Build::Expr(Expr::Empty))
+            }
+            '\r' => { self.bump();
+                    if self.bump_if('\n') {
+                        while self.cur().is_whitespace() { self.bump(); };
+                        Ok(Build::Expr(Expr::Empty))
+                    }
+                    else { Err(self.err(ErrorKind::UnrecognizedEscape(c))) } },
```
BurntSushi: Could you please format this code so that it's consistent with the rest of the code in this library? e.g., Consistent indentation, new lines after braces, etc.

golddranks:
Oops, I forgot the existence of test_err! and did some things with the expect_panic, and used .expect() to elaborate the message, but that breaks Rust 1.3 compatibility. Reverted already the tests to use test_err!, but going to revert the .expect() too now.

golddranks:
Or... do we want to support Rust 1.3? I noticed that `.expect()` is used elsewhere too, just not in the tests.

BurntSushi:
@golddranks Please don't increase the minimum required Rust version in this PR. `expect` exists as a method on `Option<T>` but not on `Result<T, E>` (pre Rust 1.4).

golddranks:
Ah, indeed, the already existing `.expect()`s were on `Option<T>`. Pardon me! I'll revert it.

BurntSushi:
Sorry for letting this PR languish, but I don't think I ever got to a place where I was comfortable with it. I don't think it's a good idea, in general, to introduce a syntactic element that *silently* behaves differently than it does in most other regex engines. I understand there are benefits to `\<newline>`, but I feel like the `x` flag covers *most* of them.

I know it's a small issue, and I think reasonable people can disagree. But at the end of the day, I feel a bit more comfortable with the conservative choice.

PR #105, opened 2015-07-31
Avoid repeated memchr in SingleSearch

bluss:
Avoid repeated memchr in SingleSearch

Repeated memchr in BMH slows down the literal search dramatically in
certain corner cases.

PR #109, opened 2015-08-04
Revert "Avoid repeated memchr in SingleSearch"

bluss:
Revert "Avoid repeated memchr in SingleSearch"

This reverts commit 77a72f9f1bbaf19ef707b4e51a6a9d9670b3a4cf,
PR #105.

I went and looked some super simple ascii string searches (for my own code)
and they all degraded with PR #105, now that I evaluated it for regex too.
(While the original issue remains, that repeated memchr is 10x worse for certain
periodic patterns).

So this is not good, the "natural language" cases were all made worse, while the
silly pathologies were improved.

To be careful, I'd like to revert the change.

### DIFF
--- a/src/prefix.rs
+++ b/src/prefix.rs
@@ -203,6 +203,10 @@ impl SingleSearch {
                 return Some(i);
             }
             i += self.shift[b as usize];
+            i += match memchr(pat[0], &haystack[i..]) {
+                None => return None,
+                Some(i) => i,
+            };
         }
         None
     }


bluss:
See #105 for more discussion & benchmarks. cc @BurntSushi 

I just didn't like that I could find simple examples that were much worse off now, even though the periodic examples were 10 times improved. I wonder what the DNA strings most look like..

BurntSushi:
As a computational biologist in a former life, I can say that the number of times I ran a regex on DNA strings was precisely `0`. ;-)

BurntSushi:
Thanks! I'm sure we'll bounce around more in the future on these. They can be pretty fiddly.

The other benchmark I want to run is on text in other languages. Namely, if most characters are in language X and most characters in language X start with the same byte, then we're going to see the pathological behavior you've identified in very real use cases.

bluss:
> As a computational biologist in a former life, I can say that the number of times I ran a regex on DNA strings was precisely 0. ;-)

awww. Reality is boring.

bluss:
@BurntSushi I think those are also cases where BMH struggles

PR #114, opened 2015-08-24
minor tweak

SeanRBurton:
Change a type signature to discard unused information.


BurntSushi:
We may actually need that info for a DFA implementation. But I'm fine with removing it for now.

alexcrichton:
r? @BurntSushi 

I'm also fine one way or the other, but I'll defer to you

BurntSushi:
Thanks for the ping! Merging under the "don't write code you don't need yet" principle. :P

PR #115, opened 2015-08-27
Abstract instruction dispatch into a trait.

SeanRBurton:
The nfa and backtracking implementations were reimplementing logic for
dispatching an instruction against a position in the input stream. I
factored this out into a `Dispatch` trait which abstracts the common
logic.

 This actually makes the code a little bigger, but it improves
orthogonality and will make it easier to switch to a bytecode (or
whatever), if desired.

### DIFF
--- a/src/backtrack.rs
+++ b/src/backtrack.rs
@@ -233,7 +233,7 @@ impl<'r, 't, 'c> Backtrack<'r, 't, 'c> {
                     }
                 }
                 Ranges(ref inst) => {
-                    if inst.matches(at.char()).is_some() {
+                    if inst.matches(at.char()) {
                         pc += 1;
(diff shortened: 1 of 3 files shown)


BurntSushi:
Thanks!

I'm not completely convinced this is worth doing at this point in time. Adding a layer of indirection in the matching engines makes it harder to understand IMO, and there's really not that much benefit that I can see. With that said, I *think* I could get past that. What has me concerned is that we still have at least a few more matching engines to build (DFA, one pass NFA, reverse NFA and/or DFA), and it's not clear to me that they will fit into this `Dispatch` trait. I think I'd rather hold off on this kind of abstraction until we have a more complete picture of the interfaces of matching engines.

PR #116, opened 2015-08-28
More small refactorings.

SeanRBurton:
Nothing too big here, mostly superficial tweaks. I did slightly change the data layout of the `Threads` data structure from Nfa, but the interface is still almost the same.

### DIFF
--- a/src/nfa.rs
+++ b/src/nfa.rs
@@ -78,7 +78,7 @@ impl<'r, 't> Nfa<'r, 't> {
     ) -> bool {
         let mut matched = false;
         q.clist.empty(); q.nlist.empty();
-'LOOP:  loop {
+        loop {
             if q.clist.size == 0 {
                 // Three ways to bail out when our current set of threads is
                 // empty.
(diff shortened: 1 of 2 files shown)


src/nfa.rs
```diff
@@ -234,17 +235,13 @@ pub struct NfaThreads {
 
 #[derive(Debug)]
 struct Threads {
-    dense: Vec<Thread>,
+    pcs: Vec<usize>,
+    cap_bucket_size: usize,
+    caps: Vec<Option<usize>>,
```
BurntSushi: What motivated this change in representation? Is there a performance improvement?

src/nfa.rs
```diff
@@ -90,7 +90,7 @@ impl<'r, 't> Nfa<'r, 't> {
                 //    soon as the last thread dies.
                 if matched
                    || (!at.is_beginning() && self.prog.anchored_begin) {
-                    break;
+                    return true;
```
BurntSushi: Does this pass tests? You can't just return `true` here because there might not be a match.

PR #121, opened 2015-09-26
"raw strings" link fix

Seeker14491:
The "raw strings" link points to the raw byte string literals section of the docs, but it should point to the raw string literals section.


BurntSushi:
Nice! Thanks! :-)

PR #131, opened 2015-11-14
Add Index<usize> and Index<&str> to Captures

sbstp:
I often use regexps with non-optional capture groups and end up having to use `caps.at(3).unwrap()` or `caps.name("name").unwrap()`. This change is a quality of life improvement that allows the indexing of `Captures`, like this: `caps[3]` and `caps["name"]`. If the group does not exist, a panic is triggered.

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -378,6 +378,11 @@ impl Regex {
     /// assert_eq!(caps.at(1), Some("Citizen Kane"));
     /// assert_eq!(caps.at(2), Some("1941"));
     /// assert_eq!(caps.at(0), Some("'Citizen Kane' (1941)"));
+    /// // You can also access the groups by index using the Index notation.
+    /// // Note that this will panic on an invalid index.
+    /// assert_eq!(&caps[1], "Citizen Kane");
+    /// assert_eq!(&caps[2], "1941");
+    /// assert_eq!(&caps[0], "'Citizen Kane' (1941)");
     /// # }
     /// ```
     ///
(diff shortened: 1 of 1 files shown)


BurntSushi:
Hmm. I hadn't implemented these in order to err on the conservative side, but perhaps the time is right to do it. I also do indeed use a lot of regexes where the captures are guaranteed to exist if the regex overall matches, so this is a nice convenience.

@alexcrichton Do you have any objections?

BurntSushi:
@sbstp Could you maybe add an example to the documentation of the `captures` method that demonstrates this functionality? It would also be good to add a little blurb that states when it might be useful (e.g., when the captures are guaranteed to exist if a match happened).

sbstp:
Sure thing. Also I'm not super familiar with the Index trait, but I noticed that it returns `str` instead of `&str`, even though the `index` method returns a reference. It it possible to make it return a `&str`?

BurntSushi:
@sbstp I'm not sure I follow. Both of your impls have `fn index<'a>(&'a self, i: usize) -> &'a str`, which says that `index` returns an `&str`.

sbstp:
Here's an example: http://is.gd/n7pbTt
It fails to compile because abc expects a `&str` but gets a `str`.

BurntSushi:
@sbstp Oh oh, I see. Yes, indexing automatically derefs the return value because it's usually what you want. In this case, you'll have to do `&a[0]`, which turns it back into a reference. Note that it isn't always necessary, for example, if you call a method. e.g., `a[0].chars()` would work to return an iterator over characters.

sbstp:
Ok thanks. The build failed on `TARGET=i686-pc-windows-msvc`, but worked on the other versions. It is something wrong with my changes?

BurntSushi:
@sbstp Nah, you're good, IIRC the Windows nightlies are busted.

sbstp:
@BurntSushi added some docs, let me know what you think.

BurntSushi:
@sbstp That looks great, thank you! Just want to run it by @alexcrichton first and this should be good to go.

BurntSushi:
@sbstp OK, I'm just going to merge this. :-)

sbstp:
Thanks!

BurntSushi:
This is on crates.io in `0.1.43`. Thanks again @sbstp!

PR #132, opened 2015-11-20
Document names_iter and made it no-copy

### LINKED ISSUE #125: Allow accessing meta info for capture groups
When using user-supplied regexes, I need information about the only dynamic property of those regexes: the number and names of capture groups.

This is useful to allow users to e.g. specify a subset of possible groups.

flying-sheep:
fixes #125

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -632,15 +632,16 @@ impl Regex {
         }
     }
 
-    #[doc(hidden)]
-    pub fn names_iter<'a>(&'a self) -> NamesIter<'a> {
+    /// Returns an iterator over the capture names.
+    pub fn capture_names<'r>(&'r self) -> CaptureNames<'r> {
         match *self {
-            Regex::Native(ref n) => NamesIter::Native(n.names.iter()),
-            Regex::Dynamic(ref d) => NamesIter::Dynamic(d.cap_names.iter())
+            Regex::Native(ref n) => CaptureNames::Native(n.names.iter()),
+            Regex::Dynamic(ref d) => CaptureNames::Dynamic(d.cap_names.iter())
         }
     }
 
-    fn names_len(&self) -> usize {
+    /// Returns the number of captures.
+    pub fn captures_len(&self) -> usize {
         match *self {
             Regex::Native(ref n) => n.names.len(),
             Regex::Dynamic(ref d) => d.cap_names.len()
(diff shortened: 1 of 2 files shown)


BurntSushi:
@flying-sheep Thanks! I think you'll need to modify this line to call `to_ownd` on each name (to get a `String` from a `&str`): https://github.com/rust-lang-nursery/regex/blob/master/regex_macros/src/lib.rs#L87 That should fix the build.

flying-sheep:
oh! why does `cargo test` not test regex_macros? because i’m on stable rust?

BurntSushi:
@flying-sheep Because it's a separate crate. You have to cd into `regex_macros` and run `cargo test`. (This repository actually has three crates in it!)

flying-sheep:
aha! i did your change, now fingers crossed that it works. else i’ll have to install multirust and all that jazz just to test this

BurntSushi:
@flying-sheep *crosses fingers* :-)

(Seriously though, multirust is awesome. You should try it out at some point! I'm sure we can squirm by here without it though.)

src/re.rs
```diff
@@ -632,8 +632,8 @@ impl Regex {
         }
     }
 
-    #[doc(hidden)]
-    pub fn names_iter<'a>(&'a self) -> NamesIter<'a> {
+    /// Returns an iterator over the capture names.
+    pub fn names_iter<'r>(&'r self) -> NamesIter<'r> {
```
BurntSushi: @flying-sheep Sorry I didn't realize this before, but can you change this to `capture_names` (add `capture_` prefix and drop `_iter` suffix)?

Correspondingly, rename the type `NamesIter` to `CaptureNames`. These names seem clearer IMO.

flying-sheep:
> @flying-sheep *crosses fingers* :-)

it worked! :laughing: 

and i promise to try multirust haha

src/re.rs
flying-sheep: sure, i like those better as well. makes the doc comments redundant, though, so i’ll improve those as well

flying-sheep:
the names are changed, the documentation amended, and i exported `names_len` (which i renamed to `captures_len`)

src/re.rs
```diff
@@ -655,20 +656,32 @@ impl Regex {
...
-    type Item=Option<String>;
+impl<'r> Iterator for CaptureNames<'r> {
+    type Item=Option<&'r str>;
+
+    fn next(&mut self) -> Option<Option<&'r str>> {
+        match *self {
+            CaptureNames::Native(ref mut i) =>
+                i.next().map(|o| *o),
+            CaptureNames::Dynamic(ref mut i) =>
+                i.next().as_ref().map(|o| o.as_ref().map(|s| s.as_ref())),
```
flying-sheep: i could also do

```rust
i.next().map(|o| (*o).as_ref().map(|s| s.as_ref()))
```

what do you prefer?

BurntSushi:
@flying-sheep Looks great! Thanks so much. :-)

src/re.rs
BurntSushi: Ooops, I missed this before I merged. It's fine how it is I guess. I don't think I have a strong preference either way.

flying-sheep:
with pleasure

PR #133, opened 2015-11-20
Reexport CaptureNames

flying-sheep:
Sorry, missed that!


BurntSushi:
Ah nice catch, thanks.

PR #135, opened 2015-11-27
Update to work with the latest rust nightly (rustc 1.6.0-nightly (1727dee16 2015-11-26))

fhartwig:


regex_macros/src/lib.rs
```diff
@@ -20,6 +20,7 @@
 extern crate regex;
 extern crate syntax;
 extern crate rustc;
```
hobofan: This line can actually be removed since rustc is only required for the plugin registry.

fhartwig: You're right, fixed. Thanks!

untitaker:
Equivalent to #136

BurntSushi:
Great, thank you!

PR #137, opened 2015-12-01
Fix MSVC-32 CI

alexcrichton:
Unfortunately panicking isn't implemented just yet, so these need to be ignored
there for now.


alexcrichton:
(may want to wait for the green light on CI before merging)

BurntSushi:
Ah, right, I knew that but completely forgot! Thanks!

PR #143, opened 2015-12-18
Loosen lifetime bounds on accessing named captures via `Index`.

jld:
The current definition equates the lifetime of the strings being
matched/captured and the string used as the capture name.  The latter
is usually `'static` (string literals), which leads to confusing
[borrow-check errors][err] on reasonable-looking code.

This change declares the two lifetimes as separate; this should be a
strict increase in the set of `impl` instances, so it's a minor version
bump.

[err][]: https://gist.github.com/jld/476a13a00ad05bd99a63

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "regex"
-version = "0.1.48"  #:version
+version = "0.1.49"  #:version
 authors = ["The Rust Project Developers"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
--- a/src/re.rs
+++ b/src/re.rs
@@ -951,6 +951,15 @@ impl<'t> Captures<'t> {
 
 /// Get a group by index.
 ///
+/// `'t` is the lifetime of the matched text.
+///
+/// The text can't outlive the `Captures` object if this method is
+/// used, because of how `Index` is defined (normally `a[i]` is part
+/// of `a` and can't outlive it); to do that, use [`at()`][]
+/// instead.
+///
+/// [`at()`]: #method.at
+///
 /// # Panics
 /// If there is no group at the given index.
 impl<'t> Index<usize> for Captures<'t> {
@@ -964,6 +973,16 @@ impl<'t> Index<usize> for Captures<'t> {
 }
 
 /// Get a group by name.
+///
+/// `'t` is the lifetime of the matched text and `'i` is the lifetime
+/// of the group name (the index).
+///
+/// The text can't outlive the `Captures` object if this method is
+/// used, because of how `Index` is defined (normally `a[i]` is part
+/// of `a` and can't outlive it); to do that, use [`name()`][]
+/// instead.
+///
+/// [`name()`]: #method.name
 ///
 /// # Panics
 /// If there is no group named by the given value.
--- a/src/re.rs
+++ b/src/re.rs
@@ -967,11 +967,11 @@ impl<'t> Index<usize> for Captures<'t> {
 ///
 /// # Panics
 /// If there is no group named by the given value.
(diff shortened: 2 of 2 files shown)


src/re.rs
```diff
@@ -970,11 +970,11 @@ impl<'t> Index<usize> for Captures<'t> {
 ///
 /// # Panics
 /// If there is no group named by the given value.
-impl<'t> Index<&'t str> for Captures<'t> {
+impl<'t, 'i> Index<&'i str> for Captures<'t> {
```
BurntSushi: Could you please add a bit of documentation here that briefly describes what the `'i` lifetime parameter stands for? Thanks! (The `'t` lifetime should also be documented, e.g., http://doc.rust-lang.org/regex/regex/struct.FindMatches.html)

BurntSushi:
Can't believe I missed this! I have one doc nit, but otherwise looks good to me. Thank you!

src/re.rs
jld: Good point; thanks.

While I'm editing the docs: I noticed that the `str` can't outlive the `Captures` if it's obtained through either `Index` impl; should that be a separate PR?  It's a little surprising, so a warning and advice to use methods instead might be nice.  (This is part of the definition of the `Index` trait — normally `a[i]` is part of `a` and can't outlive it — so it can't be fixed without changing the `Output` type to `&'t str` and breaking compatibility.)

BurntSushi: @jld So sorry I missed this comment! You raise a good point. Including those docs in this PR seems fine to me!

jld:
Really sorry for dropping the ball on this.  (Also sorry for the rebase / force-push, but I had to fix the version bump.)  This version should take care of the doc comment issues.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "regex"
-version = "0.1.48"  #:version
+version = "0.1.49"  #:version
```
BurntSushi: Could you remove the version bump please? I typically do them in separate commits with tags using a script.

src/re.rs
```diff
@@ -951,6 +951,15 @@ impl<'t> Captures<'t> {
...
 /// Get a group by index.
 ///
+/// `'t` is the lifetime of the matched text.
+///
+/// The text can't outlive the `Captures` object if this method is
+/// used, because of how `Index` is defined (normally `a[i]` is part
+/// of `a` and can't outlive it); to do that, use [`at()`][]
+/// instead.
+///
+/// [`at()`]: #method.at
```
BurntSushi: Could you remove the links here? I've generally avoided putting links to things in the docs because they're liable to become rotten.

src/re.rs
```diff
@@ -965,13 +974,23 @@ impl<'t> Index<usize> for Captures<'t> {
...
 ///
+/// `'t` is the lifetime of the matched text and `'i` is the lifetime
+/// of the group name (the index).
+///
+/// The text can't outlive the `Captures` object if this method is
+/// used, because of how `Index` is defined (normally `a[i]` is part
+/// of `a` and can't outlive it); to do that, use [`name()`][]
+/// instead.
+///
+/// [`name()`]: #method.name
```
BurntSushi: Same here for links.

BurntSushi:
@jld OK, now I'm the one who has dropped the ball! I left a few more nits. If you like, I can pull your branch and make the changes myself---I feel bad about the back and forth!

BurntSushi:
OK, I just went ahead and merged it in d8fb7f10a3c0010b0a364e969c586d1b72f8d14a. Thanks!

BurntSushi:
Released on crates.io in `0.1.55`.

PR #155, opened 2016-01-31
Remove last_match fields from FindCaptures and FindMatches

defuz:
I removed `last_match` fields from `FindCaptures` and `FindMatches` structs and simplified algorithm of `next` method.

I think that behavior of iterators should not change (at least, all tests are passed), but please double-check that.

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -1078,7 +1078,7 @@ impl<'r, 't> Iterator for FindCaptures<'r, 't> {
         // i.e., no infinite loops please.
         if e == s {
             self.last_end += self.search[self.last_end..].chars()
-                                 .next().map(|c| c.len_utf8()).unwrap_or(0);
+                                 .next().map(|c| c.len_utf8()).unwrap_or(1);
         } else {
             self.last_end = e;
         }
--- a/src/re.rs
+++ b/src/re.rs
@@ -350,6 +350,7 @@ impl Regex {
             re: self,
             search: text,
             last_end: 0,
+            skip_next_empty: false
         }
     }
 
@@ -454,6 +455,7 @@ impl Regex {
             re: self,
             search: text,
             last_end: 0,
+            skip_next_empty: false
         }
     }
 
@@ -1058,6 +1060,7 @@ pub struct FindCaptures<'r, 't> {
     re: &'r Regex,
     search: &'t str,
     last_end: usize,
+    skip_next_empty: bool
 }
 
 impl<'r, 't> Iterator for FindCaptures<'r, 't> {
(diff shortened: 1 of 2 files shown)


src/re.rs
```diff
@@ -1079,16 +1076,12 @@ impl<'r, 't> Iterator for FindCaptures<'r, 't> {
 
         // Don't accept empty matches immediately following a match.
         // i.e., no infinite loops please.
-        if e == s && Some(self.last_end) == self.last_match {
-            if self.last_end >= self.search.len() {
-                return None;
-            }
+        if e == s {
```
BurntSushi: I think the issue here is if a second empty match occurs at the end of the string, then this branch is taken, `self.last_end` is incremented by `1` and ` Some(...)` value is returned. In contrast, the code currently returns a `None`. This error appears to be triggered by the `searcher_many_zero_length_matches` test, where your patch produces a new `Match(6, 6)` that shouldn't be there.

BurntSushi:
@defuz Cool! Are you sure you ran all the tests? It looks like some of them are failing, e.g., `tests::searcher_one_zero_length_matches`. I think I commented on the bit of code that causes it.

I would definitely welcome simplifications to this code though. I'm not happy with it and it is convoluted and tricky and non-obvious what is going on. The test coverage also clearly isn't great. It looks like only two tests failed, and they aren't even regex specific tests (they're testing the implementation of the `Pattern`/`Searcher` traits.

defuz:
@BurntSushi thanks for the feedback! This problem brought me back to thinking about how it should work with empty captures at all. It seems that my previous code works well only in certain cases (not mixing empty and non-empty captures). It always returns redundant empty capture right after the non-empty captures.

I tried to get rid of the problem by adding boolean state between iterations. It seems that it works well for me. I also added 3 test that might detect problems when they occur.

BurntSushi:
@defuz Right, OK! I think I buy it. I'm still hoping for making the code clearer, but I think is at least an improvement. The added tests are a nice bonus. :-) Thank you!

BurntSushi:
@defuz Just wanted to let you know that I had to revert this change in my local branch. I came up with additional test cases that fail with this patch but pass under the old code. I tried fixing what you had here, but it ended up no longer being a simplification unfortunately. (The problem happens when there are non-adjacent successive empty width matches.) I think it ends up being clearer to test the actual last match location directly.

Here are the failing tests:

```
---- tests::match_multi_3 stdout ----
        thread 'tests::match_multi_3' panicked at 'For RE '(?m)^' against '"abc\ndef\nxyz"', expected '[(0, 0), (4, 4), (8, 8)]' but got '[(0, 0), (4, 4), (4, 4), (8, 8), (8, 8)]'', regex_macros/tests/tests.rs:380

---- tests::match_multi_4 stdout ----
        thread 'tests::match_multi_4' panicked at 'For RE '(?m)$' against '"abc\ndef\nxyz"', expected '[(3, 3), (7, 7), (11, 11)]' but got '[(3, 3), (3, 3), (7, 7), (7, 7), (11, 11), (11, 11)]'', regex_macros/tests/tests.rs:382
```

defuz:
@BurntSushi It's more like a bug in the algorithm. I found this solution:

```rust
// Don't accept empty matches immediately following a match.
// i.e., no infinite loops please.
if e == s {
    self.last_end = e + self.search[e..].chars()
                         .next().map(|c| c.len_utf8()).unwrap_or(1);
```

It seems that this fix passes all tests.

PR #157, opened 2016-02-01
Simplify RegexSplitsN

defuz:
I replaced inner fields `cur` and `limit` to single field `n` that performs a countdown to zero.


BurntSushi:
@defuz Nice! You're on a roll. :-)

PR #161, opened 2016-02-08
Implement fmt::Debug for Captures (fix #160)

### LINKED ISSUE #160: Implement Debug for Captures
src/detail/discover.rs:309:40: 309:44 error: the trait `core::fmt::Debug` is not implemented for the type `regex::re::Captures<'_>` [E0277]
    src/detail/discover.rs:309                 println!("caps: {:?}", caps);
                                                                      ^~~~

Be nice if it was.

dhardy:
Pretty ad-hoc formatting but I'm not sure if there is any expectation for this sort of thing.

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -980,6 +980,21 @@ impl<'t> Index<&'t str> for Captures<'t> {
 
 }
 
+impl<'t> fmt::Debug for Captures<'t> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        try!(writeln!(f, "Captures of {}:", self.text));
+        for i in 0..self.locs.len() {
+            try!(writeln!(f, "{}:\t{:?}", i, self.at(i)));
+        }
+        if let Some(ref h) = self.named {
+            for (n, i) in h.iter() {
+                try!(writeln!(f, "{}:\t{:?}", n, self.at(*i)));
+            }
+        }
+        Ok(())
+    }
+}
+
 /// An iterator over capture groups for a particular match of a regular
 /// expression.
 ///


src/re.rs
```diff
@@ -980,6 +980,21 @@ impl<'t> Index<&'t str> for Captures<'t> {
...
+            try!(writeln!(f, "{}:\t{:?}", i, self.at(i)));
+        }
+        if let Some(ref h) = self.named {
+            for (n, i) in h.iter() {
+                try!(writeln!(f, "{}:\t{:?}", n, self.at(*i)));
+            }
+        }
+        Ok(())
+    }
+}
```
BurntSushi: I think the convention is to show the raw type as much as possible. In this case, I suspect it's safe to drop the `text` field (since that could be quite large and distracting) but show the `locs` and `named` fields as-is.

I urge you to use the debug formatting helpers: http://doc.rust-lang.org/std/fmt/struct.DebugStruct.html :-)

BurntSushi:
Thanks for this! Definitely an oversight on my part. We should probably just show the underlying type as faithfully as we can. I think the ideal display would probably be to intelligently combine the positional and named captures somehow, but the current representation makes that a bit tricky (or impossible) to do.

dhardy:
Thanks for the comments. Examples are always key to making this sort of thing look right. I just implemented something to help debug a problem I had.

What you ask for sounds similar to what `#[derive(Debug)]` would do. I refrained from going that route because the internals looked a little confusing until I had them sussed out.

TBH I just thought taking five minutes out of my debugging routine to aid other "debuggers" might help. But I guess if you throw a couple of examples my way I might try to "improve" (subjective) on it for you.

BurntSushi:
@dhardy Thanks for the PR. I went ahead and implemented something that will hopefully be a little nicer (at the cost of some allocation, but I think that's OK since it's a `Debug` impl). Example:

```rust
let re = Regex::new(r"(?P<foo>\D+)(\d+)(?P<bar>\D+)(\d+)").unwrap();
println!("{:?}", re.captures(b"abc123xyz456"));
```

Outputs:

```
Some(Captures({0: Some("abc123xyz456"), "foo": Some("abc"), 2: Some("123"), "bar": Some("xyz"), 4: Some("456")}))
```

dhardy:
Nice.

PR #163, opened 2016-02-13
Better version of FindMatches and FindCaptures

defuz:
Ok, this PR fix #155 and makes the algorithm more intuitive (at least I think so).

Also, this behavior is more efficient because it avoid redundant actions after the recursive call and never execute a regex at the same position. Old code can do so in some cases.

:warning: This PR push changes to **dfa** branch.

@BurntSushi

### DIFF
--- a/src/re.rs
+++ b/src/re.rs
@@ -351,7 +351,7 @@ impl Regex {
             re: self,
             search: text,
             last_end: 0,
-            last_match: None,
+            skip_next_char: false
         }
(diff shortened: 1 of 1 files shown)


BurntSushi:
I think I'm just going to stick with what we have now. I still think it could be improved, but I don't think this patch is an obvious improvement. The code is (still) tricky to read, and it looks like there's an extra branch in the common case.

PR #173, opened 2016-02-22
Add regex sets.

### LINKED ISSUE #156: add multi-regex matching
Lately I've been thinking a lot about providing a "multi regex" similar to RE2's "regex set" functionality. The problem they solve is, "I have multiple regexes that I want to run over some large search text once, and I want to see every match." The poor man's way of doing this is to combine them in a single regex of alternations, e.g., `re1|re2|re3|...`. Two problems with that though:

1. The current search machinery reports non-overlapping matches. That is, it's impossible for one alternation in a regex to share a match with another alternation in the same regex.
2. To check which expressions matched, one adds capture groups and then inspect them after a match. Requiring capture groups for this functionality is bad because it incurs a performance penalty and simply isn't needed.

We can start relatively simple by providing an API that answers these three questions:

1. Do any of the given regexes match anywhere? (analogous to `is_match`)
2. If so, which of those regexes match? Where do they match? (analogous to `find`)
3. Can you show me all matches? (analogous to `find_iter`)

Adding capture groups to this API seems possible, but is tricky, so I suggest doing that after an initial implementation is done.

BurntSushi:
Regex sets permit matching multiple (possibly overlapping) regular
expressions in a single scan of the search text. This adds a few new
types, with `RegexSet` being the primary one.

All matching engines support regex sets, including the lazy DFA.

This commit also refactors a lot of the code around handling captures
into a central `Search`, which now also includes a set of matches that
is used by regex sets to determine which regex has matched.

We also merged the `Program` and `Insts` type, which were split up when
adding the lazy DFA, but the code seemed more complicated because of it.

Closes #156.


BurntSushi:
@alexcrichton This PR includes new additions to the API, but no breaking changes. The additions are completely separate from the primary `Regex` API, and can be found in `src/set.rs`. I tried to be conservative when possible. For example, it might be better to have a distinct `RegexSetBuilder` type, so that callers can incrementally add regexes (where some my fail to parse), but I think that could be added later relatively easily.

alexcrichton:
Nice! I couldn't immediately come up with a use case for these, but sounds plausible to me!

BurntSushi:
URL router, user agent matcher. Generally any time you have lots of patterns you need to match. RE2 has it. :)

alexcrichton:
Sounds like a plan to me

src/set.rs
```diff
@@ -0,0 +1,315 @@
...
+/// A regex set corresponds to the union of two or more regular expressions.
+/// That is, a regex set will match text where at least one of its
+/// constituent regular expressions matches. A regex set as its formulated here
+/// provides a touch more power: it will also report *which* regular
+/// expressions in the set match. Indeed, this is the key difference between
+/// regex sets and a single `Regex` with many alternates, since only one
+/// alternate can match at a time.
+///
+/// For example, consider regular expressions to match email addresses and
+/// domains: `[a-z]+@[a-z]+\.(com|org|net)` and `[a-z]+\.(com|org|net)`. If a
```
rkjnsn: Maybe it would make sense to use a different example here? It seems like there are already plenty of incomplete/inaccurate regexes around the internet the purport to match e-mail addresses without adding more. It seems like the docs should at least call out the fact that these are grossly simplified and will fail to match many valid e-mail addresses, so that nobody copy-pastes them without thinking.

BurntSushi: I'm happy to indulge alternative examples.

PR #176, opened 2016-02-23
Add ExprBuilder, which can set the default values of flags when parsing.

### LINKED ISSUE #172: Configure how `regex_syntax` treats `.`, case-sensitivity, etc
Currently it seems that `.` is translated to `AnyCharNoNL`, but for my use case I would prefer `AnyChar`. I imagine I could probably achieve this by tweaking the input string to something like `(?s:...)`, but it seems like something that ought to be configurable in a more "first-class" way (among other things, that will mess up the indices in my error messages, unless of course I account for it). Another similar example is being able to control if the regex is case sensitive or not.

BurntSushi:
Closes #172.

### DIFF
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -82,10 +82,7 @@ use unicode::case_folding;
 use self::Expr::*;
 use self::Repeater::*;
 
-pub use parser::is_punct;
-
-/// The maximum number of nested expressions allowed.
-const NEST_LIMIT: usize = 200;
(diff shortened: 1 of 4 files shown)


BurntSushi:
cc @nikomatsakis @jneem

BurntSushi:
N.B. CI passes the msvc Windows build but not gnu, not sure why but it appears unrelated to this change.

BurntSushi:
Published in crates.io in `0.2.5`.

PR #180, opened 2016-03-03
Set `bench = false` for the main library

bluss:
Set `bench = false` for the main library

Without this setting, the regex crate is compiled twice for cargo bench
-- once as a library, once with --test for benchmarking. Since there are
no benchmarks in the library itself it's faster to just skip this
compile.

I found this out recently, so this PR is a way to tell you about this little gotcha.
Saving time is always nice.


BurntSushi:
Oooo, sneaky! I like! Thank you. :-)

PR #182, opened 2016-03-07
Fix regex_macros

### LINKED ISSUE #181: regex_macros is broken
`Inst::Match` was given a new field, and now the crate doesn't compile on master.

```
$ cargo build
   Compiling regex_macros v0.1.31 (file:///home/manishearth/Mozilla/Git/regex/regex_macros)
src/lib.rs:437:17: 437:28 error: this pattern has 0 fields, but the corresponding variant has 1 field [E0023]
src/lib.rs:437                 Inst::Match => quote_expr!(self.cx, {
                               ^~~~~~~~~~~
src/lib.rs:437:17: 437:28 help: run `rustc --explain E0023` to see a detailed explanation
error: aborting due to previous error
Could not compile `regex_macros`.
```

(shouldn't the CI catch this?)

mcarton:
Fix #181.

### DIFF
--- a/regex_macros/src/lib.rs
+++ b/regex_macros/src/lib.rs
@@ -434,7 +434,7 @@ fn exec<'t>(
     fn step_insts(&self) -> P<ast::Expr> {
         let arms = self.prog.iter().enumerate().map(|(pc, inst)| {
             let body = match *inst {
-                Inst::Match => quote_expr!(self.cx, {
+                Inst::Match(_) => quote_expr!(self.cx, {
                     for (slot, val) in caps.iter_mut().zip(thread_caps.iter()) {
                         *slot = *val;
                     }


BurntSushi:
Thanks!

BurntSushi:
Published on crates.io in `0.1.32`.

mcarton:
Thanks!

tinyplasticgreyknight:
:+1: Now I can progress :-)

BurntSushi:
@tinyplasticgreyknight FYI, `regex!` is quite a bit slower than using `Regex::new` these days. Unless you have very specialized needs, I'd recommend going with `Regex::new`.

tinyplasticgreyknight:
Thanks, but it's actually a transitive dependency of something else I am trying to compile. :-)  Now successfully!

Issue #168, opened 2016-02-17
add second "capture" lifetime to SubCaptures and SubCapturesNamed


defuz:
I think this code should work:

``` rust
#[test]
fn test_iter_lifetimes() {
    fn get_first_subcapture<'t>(regex: &Regex, text: &'t str) -> Option<&'t str> {
        let captures = regex.captures(text).unwrap();
        captures.iter().next().unwrap()
    }

    let regex = Regex::new("(a+)(b+)").unwrap();
    let subcapture = get_first_subcapture(&regex, "aabbab");

    assert_eq!(subcapture, Some("aabb"));
}
```

But today it does not compile:

``` rust
src/re.rs:956:9: 956:17 error: `captures` does not live long enough
src/re.rs:956         captures.iter().next().unwrap()
                      ^~~~~~~~
src/re.rs:954:82: 957:6 note: reference must be valid for the lifetime 't as defined on the block at 954:81...
src/re.rs:954     fn get_first_subcapture<'t>(regex: &Regex, text: &'t str) -> Option<&'t str> {
src/re.rs:955         let captures = regex.captures(text).unwrap();
src/re.rs:956         captures.iter().next().unwrap()
src/re.rs:957     }
src/re.rs:955:54: 957:6 note: ...but borrowed value is only valid for the block suffix following statement 0 at 955:53
src/re.rs:955         let captures = regex.captures(text).unwrap();
src/re.rs:956         captures.iter().next().unwrap()
src/re.rs:957     }
```

The problem is that subcaptures can not outlive `Captures` struct, which is wrong. To fix this, we need to add new lifetime:

``` rust
pub fn iter<'c>(&'c self) -> SubCaptures<'c, 't>
```

And:

``` rust
pub struct SubCaptures<'c, 't: 'c> {
    idx: usize,
    caps: &'c Captures<'t>,
}
```

There is a similar problem with `iter_named` and `SubCapturesNamed`. Possibly related to #158

BurntSushi:
Man, what a bonehead move on my part. The current definition is trivially wrong:

```rust
pub struct SubCaptures<'t> {
    idx: usize,
    caps: &'t Captures<'t>,
}
```

How presumptuous of me to assume that the search text might not live as long as the capture value! In this case, I believe `'t` is inferred to be the shorter of the two lifetimes, which certainly can't be reconciled with the lifetime in your function declaration.

Nice find!

BurntSushi:
[...]

PR #183, opened 2016-03-10
Add regex matching for &[u8].

### LINKED ISSUE #85: implement matching on byte strings
Currently as far as I see regexes here are only for Unicode text. But they can be used to parse binary files as well (to to parse a mixture of binary and text).

For example, how can one implement `strings --all` using regexes in Rust?

BurntSushi:
This commit enables support for compiling regular expressions that can
match on arbitrary byte slices. In particular, we add a new sub-module
called `bytes` that duplicates the API of the top-level module, except
`&str` for subjects is replaced by `&[u8]`. Additionally, Unicode
support in the regular expression is disabled by default but can be
selectively re-enabled with the `u` flag. (Unicode support cannot be
selectively disabled in the standard top-level API.)

Most of the interesting changes occurred in the `regex-syntax` crate,
where the AST now explicitly distinguishes between "ASCII compatible"
expressions and Unicode aware expressions.

This PR makes a few other changes out of convenience:

1. The DFA now knows how to "give up" if it's flushing its cache too
often. When the DFA gives up, either backtracking or the NFA algorithm
take over, which provides better performance.
2. Benchmarks were added for Oniguruma.
3. The benchmarks in general were overhauled to be defined in one place
by using conditional compilation.
4. The tests have been completely reorganized to make it easier to split
up the tests depending on which regex engine we're using. For example,
we occasionally need to be able to write tests specifically for
`regex::Regex` or specifically for `regex::bytes::Regex`.
5. Fixes a bug where NUL bytes weren't represented correctly in the byte
class optimization for the DFA.

Closes #85.


BurntSushi:
cc @alexcrichton @vi @flying-sheep @birkenfeld @nikomatsakis @jneem

BurntSushi:
Important notes:

1. This *doubles* the size of the public API since the new `bytes` sub-module is essentially a replica of the top-level crate API, except with `s/&str/&[u8]` for the search strings. (e.g., `is_match` takes a `&[u8]` instead of a `&str`.)
2. The API isn't quite identical, but the key changes made were to address #168 and #151. (Both of which I'd like to make to the top-level API for 1.0.)
3. There should be no breaking changes in `regex` (there are breaking changes in `regex-syntax`).

flying-sheep:
wow, amazing!

birkenfeld:
Looks very nice.

alexcrichton:
Holy cow, amazing work as always @BurntSushi! 

Out of curiosity, does this have much impact on compile times of this crate? It'd be kinda unfortunate if most users don't use byte-based regexes but end up doubling compile times (and may be indicative of perhaps another crate should exist?), but if it's small already probably doesn't matter too much.

Other than that the concept seems good to me, having a separate API for bytes-based regexes as it does seems like they'll only ever be disjointly used.

vi:
If comple time is a problem, the feature may be under feature gate in cargo.

BurntSushi:
For debug compile times:

```
$ touch src/lib.rs 
$ time cargo build
   Compiling regex v0.1.55 (file:///home/andrew/data/projects/rust/regex)

real    0m5.770s
user    0m5.487s
sys     0m0.173s

$ $EDITOR src/*.rs # remove `bytes` module
$ touch src/lib.rs 
$ time cargo build
   Compiling regex v0.1.55 (file:///home/andrew/data/projects/rust/regex)

real    0m5.506s
user    0m5.283s
sys     0m0.130s
```

For `--release` times:

```
$ touch src/lib.rs 
$ time cargo build --release
   Compiling regex v0.1.55 (file:///home/andrew/data/projects/rust/regex)

real    0m16.714s
user    0m16.517s
sys     0m0.090s
$ $EDITOR src/*.rs # remove `bytes` module
$ touch src/lib.rs 
$ time cargo build --release
   Compiling regex v0.1.55 (file:///home/andrew/data/projects/rust/regex)

real    0m16.346s
user    0m16.103s
sys     0m0.140s
```

So it looks like the difference is negligible. I'm actually not too surprised by this since the matching engines themselves *always* match on `&[u8]`, even when using Unicode-only regexes. (It wasn't always this way, but it became more convenient for a variety of reasons with the arrival of the DFA, which really really demands `&[u8]` matching.) As a result, the only "additional" code this introduces is the API itself, which all told isn't *that* much. :-)

> Other than that the concept seems good to me, having a separate API for bytes-based regexes as it does seems like they'll only ever be disjointly used.

Those were my thoughts as well. :-)

alexcrichton:
Awesome, :+1: from me!

PR #196, opened 2016-04-09
Add ASCII word boundaries to the lazy DFA.

BurntSushi:
In other words, `\b` in a `bytes::Regex` can now be used in the DFA.
This leads to a big performance boost:

```
sherlock::word_ending_n                  115,465,261 (5 MB/s)  3,038,621 (195 MB/s)    -112,426,640  -97.37%
```

Unfortunately, Unicode word boundaries continue to elude the DFA. This
state of affairs is lamentable, but after a lot of thought, I've
concluded there are only two ways to speed up Unicode word boundaries:

1. Come up with a hairbrained scheme to add multi-byte look-behind/ahead
   to the lazy DFA. (The theory says it's possible. Figuring out how to
   do this without combinatorial state explosion is not within my grasp
   at the moment.)
2. Build a second lazy DFA with transitions on Unicode codepoints
   instead of bytes. (The looming inevitability of this makes me queasy
   for a number of reasons.)

To ameliorate this state of affairs, it is now possible to disable
Unicode support in `Regex::new` with `(?-u)`. In other words, one can
now use an ASCII word boundary with `(?-u:\b)`.

Disabling Unicode support does not violate any invariants around UTF-8.
In particular, if the regular expression could lead to a match of
invalid UTF-8, then the parser will return an error. (This only happens
for `Regex::new`. `bytes::Regex::new` still of course allows matching
arbitrary bytes.)

Finally, a new `PERFORMANCE.md` guide was written.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -11,33 +11,32 @@ expressions using finite automata: https://swtch.com/~rsc/regexp/
 ## Architecture overview
 
 As you probably already know, this library executes regular expressions using
-finite automata. In particular, a design goal is to make searching linear with
-respect to both the regular expression and the text being searched. Meeting
-that design goal on its own is not so hard and can be done with an
-implementation of the NFA algorithm, as described in:
-https://swtch.com/~rsc/regexp/regexp2.html --- This library contains such an
-implementation in src/nfa.rs. (N.B. PCRE's documentation also claims to
-implement the NFA algorithm by citing Jeffrey Friedl's book, "Mastering Regular
-Expressions." The book unfortunately conflates terminology.)
+finite automata. In particular, a design goal is to make searching linear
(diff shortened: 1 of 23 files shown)


BurntSushi:
> This
state of affairs is lamentable, but after a lot of thought, I've
concluded there are only two ways to speed up Unicode word boundaries:

Here's another idea: what if the DFA ran even when there were Unicode word boundaries? Instead of actually trying to match them, we could just ignore them. If the DFA reports a match, then it could be filtered through a (slower) engine that does support Unicode word boundaries. The key insight here is that word boundaries are assertions, and as such, never extend or shorten a match. The problem with this approach is that it could result in quadratic behavior, since if the filter rejects the DFA's match, the search would have to start again immediately after the *start* of the previous search. *sigh*

PR #198, opened 2016-04-09
Fix typo Aho-Coraisck => Aho-Corasick

matthieu-m:
Checked correct spelling with [Wikipedia](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm).


BurntSushi:
Thanks! Totally just a typo. :-)

PR #208, opened 2016-04-22
Revert "Replace mempool with thread_local"

### LINKED ISSUE #206: 0.1.66 breaks semver
0.1.66 includes cdee1e7, which introduces a **breaking change**.  Existing code that compiles under Rust 1.3.0-1.5.0 stops compiling due to the transitive dependency on `void`.

BurntSushi:
This caused regex to require Rust 1.6, which really should require a
semver bump since it will break existing code. For now, we'll stick with
mempool and move to the faster thread_local version on the next semver
bump (hopefully 1.0).

Fixes #206

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.6.0
+  - 1.3.0
   - stable
   - beta
   - nightly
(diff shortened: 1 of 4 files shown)


BurntSushi:
cc @Amanieu I'll undo this revert on the next semver bump. Sorry about the churn.

Amanieu:
Don't bother, I'm pushing an update to thread_local to make it work on 1.3.

BurntSushi:
@Amanieu Thank you! :-)

alexcrichton:
r=me but if a new version of `thread_local` comes out seems like that works too!

alexcrichton:
Aha, https://github.com/rust-lang-nursery/regex/pull/209 does that

PR #209, opened 2016-04-22
Bump thread_local version and test on Rust 1.3.

### LINKED ISSUE #206: 0.1.66 breaks semver

BurntSushi:
Fixes #206.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.6.0
+  - 1.3.0
   - stable
   - beta
   - nightly
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ aho-corasick = "0.5.1"
 # For skipping along search text quickly when a leading byte is known.
 memchr = "0.1.9"
 # For managing regex caches quickly across multiple threads.
-thread_local = "0.2.0"
+thread_local = "0.2.4"
 # For parsing regular expressions.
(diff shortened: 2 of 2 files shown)


Amanieu:
I don't think bumping the version number is necessary since Cargo will automatically pick up the new version.

BurntSushi:
I think it's necessary to specify a minimum bound, otherwise it's not *guaranteed* that Cargo will pick up the new version? (e.g., If there's some other constraint on the version of `thread_local` in a build.) I might be misunderstanding though.

alexcrichton:
Yeah this way regex say it's *only* compatible with 0.2.4 and above (which I believe is the correct case)

BurntSushi:
@alexcrichton So, I hadn't merged this yet because I don't think it will build on Rust 1.3 on Windows. :P So I guess we technically support Rust 1.3+ on Linux and Rust 1.4+ on Windows.

alexcrichton:
Gah oops! Sorry about that :(

Out of curiosity, is there one place these version dependencies are coming from? Or are they just for convenience?

BurntSushi:
@alexcrichton I think I'm just sympathetic to the "your build just broke because a semver bump required a newer Rust." I don't think we have any policy on it yet, but maybe it's time to talk about it? :-)

alexcrichton:
Yeah seems fine to me, I'll add an item to our agenda.

PR #210, opened 2016-04-22
Fixes a performance bug in bytes::Regex::replace.

BurntSushi:
This was using `Vec::extend` to accumulate bytes in a buffer, but this
compiles down to less efficient code than, say, `Vec::extend_from_slice`.
However, that method is newly available as of Rust 1.6, so we do a small
backport to regain performance.

This bug was noticed by @llogiq here: https://github.com/TeXitoi/benchmarksgame-rs/issues/31
In particular, this increases the performance of bytes::Regex two-fold on
that benchmark.

### DIFF
--- a/examples/shootout-regex-dna-bytes.rs
+++ b/examples/shootout-regex-dna-bytes.rs
@@ -14,7 +14,7 @@ use std::thread;
 macro_rules! regex { ($re:expr) => { ::regex::bytes::Regex::new($re).unwrap() } }
 
 fn main() {
-    let mut seq = Vec::with_capacity(50 * (1 << 20));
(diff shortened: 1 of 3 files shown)


BurntSushi:
@alexcrichton I know I'm really pushing it with this Rust 1.3 business. :-)

An alternative (which would let us omit the `unsafe`) would be to add a feature and enable it with a `build.rs`, and then use `Vec::extend_from_slice` when the feature is enabled and fall back to something slower if not.

alexcrichton:
Looks good to me!

llogiq:
@BurntSushi Great!

PR #212, opened 2016-04-23
Drop PartialEq impl for Regex (fixes #178)

### LINKED ISSUE #178: drop the PartialEq impl
I'm now of the opinion that merging #81 was a mistake. I think having a `PartialEq` impl on `Regex` is incredibly misleading, even though it is clearly documented. In particular, regular expressions do actually have a formal notion of equality, and it is quite expensive to compute. We simply shouldn't expose any `PartialEq` or `Eq` impls.

If users need a `PartialEq` impl (e.g., to store regexes as keys in a HashMap?), then I think requiring a newtype is fine.

wesleywiser:


wesleywiser:
First time contributor to `regex` so if I've messed something up, please let me know :)

BurntSushi:
@wesleywiser This looks great! Thank you! So there are two problems. One is that #178 is part of the `1.0` milestone and is also a breaking change, so we can't quite land it yet. The other problem is that my set of issues for the `1.0` milestone roughly corresponds to things that I want to address in the regex 1.0 RFC, which means it hasn't actually been decided yet whether we're going through with said changes.

I'm really sorry about the confusion---I didn't really think too much about how this organization might look to new contributors! How about we leave this PR open, and if we do decide to take it, I'll merge it in then?

wesleywiser:
No problem! If you want to leave this open or close it, either is fine.

BurntSushi:
Great! I'll leave it open. Thanks again!

PR #214, opened 2016-04-26
Fix lifetimes for subcapture iterators in str regexes

birkenfeld:
So that strings extracted from a Captures::iter[_named]() can be used
after the captures struct is dropped.

Note: I started with the `re_unicode` API, introducing new lifetimes. Then I discovered that it was already correct in `re_bytes`, but I think the new lifetime names are more consistent, so I changed to `'c` there.

The `'c'` lifetime in the `iter_pos` definitions is elidable, I can do it that way if you prefer.

### DIFF
--- a/src/re_bytes.rs
+++ b/src/re_bytes.rs
@@ -682,21 +682,21 @@ impl<'t> Captures<'t> {
 
     /// Creates an iterator of all the capture groups in order of appearance
     /// in the regular expression.
-    pub fn iter<'a>(&'a self) -> SubCaptures<'a, 't> {
+    pub fn iter<'c>(&'c self) -> SubCaptures<'c, 't> {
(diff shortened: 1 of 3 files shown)


BurntSushi:
@birkenfeld Thanks! This is issue #168. When I wrote the bytes API, I fixed this issue. However, the addition of a type parameter to a public type is a breaking change, so I'm holding off on making this until the next semver bump.

birkenfeld:
Ah right. I was only thinking about the methods, where there is no breakage, but the type change itself is indeed incompatible.

Well, I guess you'll make a 0.2 (or 1.0?) bump eventually.

BurntSushi:
@birkenfeld Yeah, #168 is on the 1.0 milestone. :-)

Issue #215, opened 2016-04-26
Finding matches with capturing subgroups anchored at beginning depends on length of haystack?


birkenfeld:
Please refer to the following gist for the test code: https://gist.github.com/birkenfeld/ffc2f0da8b93da33f9fbd5a08a8790f3

In short, I'm experimenting with porting Pygments to Rust. The main task is implementing a scanner/lexer using lists of regexes for each state of the scanner. The scanner is an `Iterator`, and on each iteration, it goes through the regex list in order and tries to match every regex against the remaining text (all regexes are anchored at the beginning with `\A`, in Python `re.match()` is used). When a match is found, the remaining string is updated, and the search begins at the beginning of the list on the next iteration. The scanner uses `captures()` instead of `find()` because I want to use subgroups to assign different token types to them within one match, without having to use states for that.

Now, I've noticed that making the haystack longer, e.g. by appending copies of the test input, makes the matching slower and slower. You can see that in the benchmark output in the gist, where the timing is for scanning through the same string every time, but within a slice that is 1x, 2x, etc. this base string. I.e. the timings should stay constant (and they do when using `.find()`).

It seems strange to me that when the patterns are anchored at the start of the haystack, and are structured like the ones in the example (i.e. there are no patterns that have to scan through to the end of the haystack) that the time to match should not depend on the haystack length. Is this a problem of optimization (e.g. looking in the whole haystack for literals without applying the special constraints due to the `\A` assertion)?

Sorry for the long issue and code, hope it's understandable.

BurntSushi:
OK. This is an awesome (and valid) bug report. Thank you.

I'll start off by baking your noodle. I take it you decided to stop benchmarking at 20x, because clearly, a pattern had formed! But, alas, you were deceived! If you continued, you'd see this:

```
test bench_001x ... bench:     445,333 ns/iter (+/- 1,398)
test bench_002x ... bench:     563,531 ns/iter (+/- 5,479)
test bench_003x ... bench:     670,828 ns/iter (+/- 3,312)
test bench_005x ... bench:     901,192 ns/iter (+/- 2,839)
test bench_010x ... bench:   1,566,748 ns/iter (+/- 4,172)
test bench_020x ... bench:   3,538,074 ns/iter (+/- 5,703)
test bench_100x ... bench:     669,923 ns/iter (+/- 708)
test bench_200x ... bench:     671,756 ns/iter (+/- 8,083)
[...]

PR #217, opened 2016-04-27
Add known upper limit to capture search.

BurntSushi:
The DFA will report the end location of a match, so we should pass that
along to capture detection. In theory, the DFA and the NFA report the
same match locations, so this upper bound shouldn't be necessary---the
NFA should quit once it finds the right match. It turns out though
bounding the text has two important ramifications:

1. It will enable the backtracking engine to be used more often. In
particular, the backtracking engine can only be used on small inputs and
this change decreases the size of the input by only considering the
match.
2. The backtracking engine must start every search by zeroing memory
that is proportional to the size of the input. If the input is smaller,
then this runs more quickly.

We are also careful to bound the match to one additional "character"
past the end of the match, so that lookahead operators work correctly.

See also: #215.

### DIFF
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -10,6 +10,7 @@
 
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::cmp;
 use std::sync::Arc;
 
 use thread_local::CachedThreadLocal;
@@ -27,6 +28,7 @@ use re_bytes;
 use re_trait::{RegularExpression, Slot};
 use re_unicode;
 use set;
+use utf8::next_utf8;
 
 /// Exec manages the execution of a regular expression.
 ///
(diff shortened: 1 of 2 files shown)


src/exec.rs
```diff
@@ -439,9 +431,16 @@ impl<'c> RegularExpression for ExecNoSync<'c> {
...
+                        // We need the +1 here to account for lookahead
+                        // operators.
+                        let e = if self.ro.nfa.uses_bytes() {
+                            cmp::min(e + 1, text.len())
+                        } else {
+                            cmp::min(next_utf8(text, e), text.len())
+                        };
                         self.captures_nfa(
-                            MatchNfaType::Auto, slots, text, s)
+                            MatchNfaType::Auto, slots, &text[..e], s)
```
birkenfeld: Should this change also be applied above in the `MatchType::Literal` case?

birkenfeld: (And below in the `DfaAnchoredReverse` case?)

BurntSushi: Good catch on the literal case. For anchored reverse it won't help because if there's a match, it's guaranteed to end at the input (since the case only happens when the regex is anchored to the end of the text).

birkenfeld: I see, thanks!

Issue #166, opened 2016-02-15
expose more knobs


BurntSushi:
`Regex` currently has two constructors, `Regex::new` and `Regex::with_size_limit`. The latter is the same as the former, except it allows one to bound the size of the compiled program. The idea is here to control how much memory is used if a regex is compiled from an untrusted source.

There are other knobs that seem useful for exposing to the user:
- The recursion limit used in `regex-syntax` for recursively simplifying a regex. (It'd be nice to move this to a stack on the heap, but it seems tricky.)
- The cache size used in the lazy DFA. For regexes that create lots of distinct states, it's possible to realize big gains if you're willing to spend the memory to do it. Currently, it is set to a constant of ~2MB.

There are other knobs that maybe shouldn't be exposed, but could:
- Control the budget for extracting literal prefixes.
- Control whether the DFA is used. (There exists regexes and inputs where avoiding the lazy DFA is actually faster, but it's probably hard to know when this is without some experience with the internals of this crate.)
- Control whether common suffixes are factored out of compiled programs. This could reduce compilation times at the expense of bigger programs.

It's not clear what, if any, of these things should be exposed. I feel like the knobs that control memory bounds should be accessible to callers, because when you need them, you _really need them_.

BurntSushi:
I've given this some thought, and this is where I'm currently at:

1. There should be a struct with no public fields that permits setting various knobs that can be applied to the construction of a `Regex`. Knobs can be set using methods. This permits expanding the set of the knobs in the future in an API compatible way.
2. We should probably start with just two knobs: the existing compile size limit and a new knob that controls how much cache the lazy DFA is allowed to use.
3. This probably means that `Regex` will continue to need an alternate constructor, much like how `Vec` or `HashMap` have alternate constructors. For example, `Regex::with_config` or something.

PR #218, opened 2016-04-29
Add rure, a C API.

BurntSushi:
This commit contains a new sub-crate called `regex-capi` which provides
a C library called `rure`.

A new `RegexBuilder` type was also added to the Rust API proper, which
permits both users of C and Rust to tweak various knobs on a `Regex`.
This fixes issue #166.

Since it's likely that this API will be used to provide bindings to
other languages, I've created bindings to Go as a proof of concept:
https://github.com/BurntSushi/rure-go --- to my knowledge, the wrapper
has as little overhead as it can. It was in particular important for the
C library to not store any pointers provided by the caller, as this can
be problematic in languages with managed runtimes and a moving GC.

The C API doesn't expose `RegexSet` and a few other convenience functions
such as splitting or replacing. That can be future work.

Note that the regex-capi crate requires Rust 1.9, since it uses
`panic::catch_unwind`.

This also includes tests of basic API functionality and a commented
example. Both should now run as part of CI.


BurntSushi:
@alexcrichton Some thoughts:

1. I'm not much of a C programmer, so don't assume I know what I'm doing. :-)
2. Does my use of `panic::catch_unwind` look right/reasonable?
3. How about the structure of `regex-capi` and naming? I mostly got this structure from [`tendril`](https://github.com/servo/tendril) and it seems reasonable.

regex-capi/src/lib.rs
```diff
@@ -0,0 +1,515 @@
...
+use std::panic::{self, AssertUnwindSafe};
+use std::process;
+use std::ptr;
+use std::slice;
+use std::str;
+
+use libc::{c_char, uint32_t};
+use regex::bytes;
+
+macro_rules! protect {
```
huonw: This could theoretically incorporate more of the FFI set-up, e.g.

```rust
macro_rules! ffi_fn {
     (fn $name: ident($($arg: ident: $arg_ty: ty),*) -> $ret: ty $body: block) => {
          #[no_mangle]
          pub extern fn $name($($arg: $arg_ty),*) -> $ret { ... }
     }
}

ffi_fn! {
    fn rust_compile_must(pattern: *const c_char) -> *const Regex { ... }
}
```

regex-capi/src/lib.rs
```diff
@@ -0,0 +1,515 @@
...
+                } else if let Some(s) = err.downcast_ref::<String>() {
+                    s.to_owned()
+                } else {
+                    "UNABLE TO SHOW RESULT OF PANIC.".to_owned()
+                };
+                let _ = writeln!(
+                    &mut io::stderr(),
+                    "panic unwind caught, aborting: {:?}",
+                    msg);
+                process::exit(1)
```
huonw: I would think that a true abort is more appropriate, or at least using not-`1`. `exit` seems a purposeful user-interface thing (e.g. a command-line utility fails with a different code for various possible problems), while this sort of error is more catastrophic than that.

BurntSushi: @huonw I see. How can I cause that to happen?

huonw: Well, not using `1` could be happened by using something like `2` or `100` instead. :stuck_out_tongue_closed_eyes: 

More seriously, I *think* `libc::abort` does that sort of stuff (at least, if someone's turned core dumps on).

BurntSushi: Interesting. I didn't think at all about the distinction between `process::exit(...)` and `abort`. Happy to switch!

regex-capi/src/lib.rs
BurntSushi: Nice, thank you. I did this. It'd be cool to just have one `ffi` block and stuff all the extern fns in it, but I think that makes the macro much more complex.

regex-capi/src/lib.rs
BurntSushi: Done!

PR #220, opened 2016-04-30
fix mismatched/redundant integer types in c interface

baby-bell:
technically, `usize` is `uintptr_t` in C, not `size_t`, although they are identical for x86(-64). I changed some function signatures and added a few casts here and there to uniformly use `size_t` in Rust and C. As well, I removed usage of `uint32_t`, since `u32` serves just as well.

### DIFF
--- a/regex-capi/src/lib.rs
+++ b/regex-capi/src/lib.rs
@@ -158,7 +158,7 @@ ffi_fn! {
         options: *const Options,
         error: *mut Error,
     ) -> *const Regex {
-        let pat = unsafe { slice::from_raw_parts(pattern, length as usize) };
+        let pat = unsafe { slice::from_raw_parts(pattern, length) };
         let pat = match str::from_utf8(pat) {
             Ok(pat) => pat,
             Err(err) => {
@@ -222,8 +222,8 @@ ffi_fn! {
         start: size_t,
     ) -> bool {
         let re = unsafe { &*re };
-        let haystack = unsafe { slice::from_raw_parts(haystack, len as usize) };
-        re.is_match_at(haystack, start as usize)
+        let haystack = unsafe { slice::from_raw_parts(haystack, len) };
+        re.is_match_at(haystack, start)
     }
 }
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
Hmm... Under what circumstances are `uintptr_t` and `size_t` of a different size? OK, I found this: http://stackoverflow.com/questions/1464174/size-t-vs-intptr-t

Internally, Rust's regex engine uses `usize`. Doesn't this imply that the C API has to change to use `uintptr_t`? But that seems strange, since I understand `uintptr_t` to be for holding pointer values, not array indexes.

I guess with the PR as is, there are casts from `size_t` to `usize`. It feels like those should be OK in practice, since that SO post seems to suggest that the maximum array index might be smaller than the maximum pointer value (if they're not equal in the first place).

But there are also casts from `uintptr_t` (`usize`) to `size_t`, and it feels like those are the ones that will fail on platforms where `size_t` is not the same as `uintptr_t`.

So, what do we do? My inclination is to accept your PR because:

1. `size_t` is the correct C type to use.
2. We *probably* don't care about platforms where `sizeof(uintptr_t) != sizeof(size_t)`. I'd be happy to be corrected on that.

cc @alexcrichton

BurntSushi:
@baby-bell Thanks for this PR and bring it to my attention. :-)

alexcrichton:
The equivalence of `size_t` and `usize` is baked into Rust in **many** other places, and all current platforms Rust supports has `size_t` equivalent to `usize`. It's probably not a bad idea to use C types in the FFI declarations in Rust just for some extra safety in any case, though.

BurntSushi:
Ah, OK. The C API does indeed already use `size_t`. I think that means it's fine to use `usize` in the Rust function definitions, as I'm not sure what using `size_t` explicitly buys in this case. (It seems roughly equivalent to the cast introduced in this PR anyway?)

alexcrichton:
Yeah I think `size_t` in the C API makes sense, and by using `size_t` in Rust as well you'll just naturally work on platforms where `size_t` is the same as `usize`, and then you'll get compile failures for other platforms (which would require explicit support anyway.

BurntSushi:
After talking with @eddyb and @alexcrichton on IRC, I think I understand what you're saying now. The key I was missing is that if the FFI definitions switch to `size_t`, then that's right, but we should *remove* the casts to `usize`. Since `libc` makes `size_t` an alias to `usize`, this will cause compile time failures on any platform that defines `size_t` to be some type that is not `usize`.

@baby-bell could you make that change? Then we can merge. Thank you. :-)

baby-bell:
I've removed the casts as requested

BurntSushi:
@baby-bell Awesome, thank you!

PR #221, opened 2016-04-30
Rename some methods on SparseSet to better match sets in the standard…

gereeter:
… library.

This also removes a comment that seems to be outdated.


BurntSushi:
Looks good, thank you!

BurntSushi:
And you're spot on about the comment. Definitely outdated. :-)

PR #223, opened 2016-05-01
Fix #186.

### LINKED ISSUE #186: DFA misses an opportunity to short circuit with an anchored RegexSet
While trying to implement a lexer, I noticed that ```RegexSet::matches``` exhibits very bad performance when used on long strings.
The following test case runs in 1 minute, 16 seconds (cargo build --release):
```rust
extern crate regex;
fn main() {
    let mut test_input = String::from("42 ");
    for _ in 0 .. 100_000 { test_input.push_str("XXXXXXXXXXXXXXXXXXXX"); } // Delete me

    let set = regex::RegexSet::new(&[r"^[0-9]+", r"^THIS_SHOULD_NEVER_MATCH"]).unwrap();
    for _ in 0 .. 10_000 { println!("{:?}", set.matches(&test_input)); }
}
```
Without the line marked with ```// Delete me```, the program runs in about 50 ms.

If I understand correctly, ```matches``` should try to match only at the very beginning of the string, thanks to the ^ anchor. However, it would seem that ```matches``` traverses the whole string on each call.

BurntSushi:
This enables RegexSets to short-circuit when:

1. All patterns are anchored to the beginning of the input.
2. All patterns have either matched or will never match.

We make this happen by checking whether all NFA states in a DFA state
are match states, when a DFA match is observed. If all NFA states are
match states, and since all match states are final states, we know that
the current set of matches will never change. Since we don't care about
reporting location information, we can quit.

N.B. If no matches can be found, then the DFA will short circuit using its
normal mechanism.

### DIFF
--- a/src/dfa.rs
+++ b/src/dfa.rs
@@ -581,6 +581,19 @@ impl<'a> Fsm<'a> {
                 self.last_match_si = next_si;
                 prev_si = next_si;
 
+                // This permits short-circuiting when matching a regex set.
+                // In particular, if this DFA state contains only match states,
+                // then it's impossible to extend the set of matches since
+                // match states are final. Therefore, we can quit.
+                if self.prog.matches.len() > 1 {
+                    let state = self.state(next_si);
+                    let just_matches = state.insts.iter()
+                         .all(|&ip| self.prog[ip as usize].is_match());
+                    if just_matches {
+                        return result;
+                    }
+                }
+
                 // Another inner loop! If the DFA stays in this particular
                 // match state, then we can rip through all of the input
                 // very quickly, and only recording the match location once
--- a/src/prog.rs
+++ b/src/prog.rs
@@ -289,6 +289,16 @@ pub enum Inst {
     Bytes(InstBytes),
 }
 
+impl Inst {
+    /// Returns true if and only if this is a match instruction.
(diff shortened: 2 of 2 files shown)


BurntSushi:
cc @dprien @birkenfeld

birkenfeld:
Thanks for the note! You'll be pleased to know that for my use case, the timings dropped from 
```
test test::benches::highlight_html_001x ... bench:   4,413,488 ns/iter (+/- 97,944)
```
to a nice
```
test test::benches::highlight_html_001x ... bench:     278,142 ns/iter (+/- 31,189)
```
Alas, there's still no speedup compared to sequentially matching all expressions, which clocks in at
```
test test::benches::highlight_html_001x ... bench:     272,105 ns/iter (+/- 10,361)
```
This is possibly quite dependent on the individual lexer (the number of regexes per state, for example), so I'll definitely continue to try RegexSet with different configurations as I add them.

BurntSushi:
That is great news that `RegexSet` now at least as comparable performance.

And you're right, it is dependent. In particular, on the number of mismatches before finding a match. For example, if you order your regexes by most likely to match to least likely, then depending on the distribution, one could definitely get comparable performance.

In any case, thank you so much for testing out this PR and confirming that it is, in fact, fixed. :-)

PR #225, opened 2016-05-04
Merge forward/backward DFA matching.

SeanRBurton:
I couldn't get the benchmark suite working, but I also have a branch using a macro rather than a function in case this is a performance regression (not that it should be).


BurntSushi:
I'm actually not much of a fan of this change unfortunately. The code duplication is unfortunate, but I think the complexity of the code resulting from merging them is even more unfortunate.

Issue #199, opened 2016-04-09
decrease memory usage of DFA with variable width delta encoding of instruction pointers


BurntSushi:
(This is a ticket that I feel has limited scope, and as such, I'd be happy to mentor it. I think I'd consider this "medium" if one has worked with finite automata before, and probably "hard" if not.)

The lazy DFA works by computing states as they are visited during search. At most one state is created per byte of the haystack. In the vast majority of cases, the time it takes to do state construction is negligible because they are reused throughout search. However, the DFA does have pathological cases that can lead to exponential blow up in the number states. In particular, the worst case is when every byte of input results in a new state being created. For this reason, the DFA puts a bound on how many states are stored. If that bound is reached, all existing states are deleted and the DFA continues on (possibly recomputing states that were previously deleted). Since creating states can be slow, if the DFA clears the state cache too frequently, then the DFA will give up and a slower matcher will run in its place.

Therefore, given some fixed memory requirement of `N` bytes, it makes sense to try and figure out how we can store as many states as possible in `N` bytes. The more states we can fit into `N` bytes, the less frequently we flush the cache, which in turn increases the number of regexes and haystacks that we can process with the DFA.

Let's take a look at how [`State` is defined](https://github.com/rust-lang-nursery/regex/blob/2ab7be3d043a1ef640dc58ec4a4038d166ba1acd/src/dfa.rs#L222):

```rust
struct State {
    next: Box<[StatePtr]>,
    insts: Box<[InstPtr]>,
    flags: StateFlags,
}
```

`next` corresponds to the state's transitions. `insts` corresponds to the [set of NFA states that make up this single DFA state](https://en.wikipedia.org/wiki/Powerset_construction#Intuition). `flags` indicates some facts about the state, such as whether it is a match state, whether it observes word characters (to support word boundary assertion lookaround) and whether any of its NFA states correspond to empty assertions.

One somewhat straight-forward way to reduce the memory requirements is to optimize this representation. `flags` is probably as small as it's going to get: it's a single byte. `next` is also unfortunately as small as it will get since it must support fast random access. Notably, it is used inside the DFA's inner loop. `insts` however is not used in the DFA fast path and is only used during state construction. State construction is permitted to be somewhat slow, which means we should be able to optimize for space at the expense of time.

[...]

PR #226, opened 2016-05-05
Implement compression of DFA states.

SeanRBurton:
Implements the suggestion from issue #199.

### DIFF
--- a/src/dfa.rs
+++ b/src/dfa.rs
@@ -241,7 +241,9 @@ impl<T> Result<T> {
 /// it is packed into a single byte; Otherwise the byte 128 (-128 as an i8)
 /// is coded as a flag, followed by 4 bytes encoding the delta.
 #[derive(Clone, Eq, Hash, PartialEq)]
-struct State{data: Box<[u8]>}
+struct State{
+    data: Box<[u8]>,
+}
 
 /// InstPtr is a 32 bit pointer into a sequence of opcodes (i.e., it indexes
 /// an NFA state).
@@ -250,12 +252,12 @@ struct State{data: Box<[u8]>}
 /// `u32` here for the DFA to save on space.
 type InstPtr = u32;
 
-//Used to construct new states.
+// Used to construct new states.
 fn push_inst_ptr(data: &mut Vec<u8>, prev: &mut InstPtr, ip: InstPtr) {
-	let delta = (ip as i32) - (*prev as i32);
+    let delta = (ip as i32) - (*prev as i32);
     if delta.abs() <= 127 {
         data.push(delta as u8);
-		*prev = ip;
+        *prev = ip;
         return;
     }
     let delta = delta as u32;
@@ -264,12 +266,12 @@ fn push_inst_ptr(data: &mut Vec<u8>, prev: &mut InstPtr, ip: InstPtr) {
     let b = (delta & (0xFF << 1 * 8)) >> 1 * 8;
     let c = (delta & (0xFF << 2 * 8)) >> 2 * 8;
     let d = (delta & (0xFF << 3 * 8)) >> 3 * 8;
-	data.push(128);
+    data.push(128);
     data.push(a as u8);
     data.push(b as u8);
     data.push(c as u8);
     data.push(d as u8);
-	*prev = ip;
+    *prev = ip;
 }
 
 struct InstPtrs<'a> {
@@ -1185,9 +1187,9 @@ impl<'a> Fsm<'a> {
         // are conditional, we need to make them part of a state's key in the
         // cache.
 
-		// Reserve 1 byte for flags.
+        // Reserve 1 byte for flags.
         let mut insts = vec![0];
-		let mut prev = 0;
+        let mut prev = 0;
         for &ip in q {
             let ip = usize_to_u32(ip);
             match self.prog[ip as usize] {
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks! This was assigned to @killercup...

src/dfa.rs
```diff
@@ -225,24 +225,23 @@ impl<T> Result<T> {
...
-    /// epsilon transitions are necessarily followed! Namely, epsilon
-    /// transitions that correspond to empty assertions are only followed if
-    /// the flags set at the current byte satisfy the assertion.
-    insts: Box<[InstPtr]>,
-    /// Flags for this state, which indicate whether it is a match state,
-    /// observed an ASCII word byte and whether any instruction in `insts`
-    /// is a zero-width assertion.
-    flags: StateFlags,
-}
+struct State{data: Box<[u8]>}
```
BurntSushi: Could you please use the same style of formatting as the rest of the code? Probably `struct State(Box<[u8]>)` would be nice here.

src/dfa.rs
```diff
@@ -251,6 +250,74 @@ struct State {
 /// `u32` here for the DFA to save on space.
 type InstPtr = u32;
 
+//Used to construct new states.
```
BurntSushi: Space after the `//`, please.

src/dfa.rs
```diff
@@ -251,6 +250,74 @@ struct State {
...
+fn push_inst_ptr(data: &mut Vec<u8>, prev: &mut InstPtr, ip: InstPtr) {
+	let delta = (ip as i32) - (*prev as i32);
```
BurntSushi: Fix indentation.

src/dfa.rs
```diff
@@ -251,6 +250,74 @@ struct State {
...
+    if delta.abs() <= 127 {
+        data.push(delta as u8);
+		*prev = ip;
```
BurntSushi: indentation

src/dfa.rs
```diff
@@ -251,6 +250,74 @@ struct State {
...
+        return;
+    }
+    let delta = delta as u32;
+    // Write 4 bytes in little-endian format.
+    let a = (delta & (0xFF << 0 * 8)) >> 0 * 8;
+    let b = (delta & (0xFF << 1 * 8)) >> 1 * 8;
+    let c = (delta & (0xFF << 2 * 8)) >> 2 * 8;
+    let d = (delta & (0xFF << 3 * 8)) >> 3 * 8;
+	data.push(128);
```
BurntSushi: indentation... and so on

src/dfa.rs
```diff
@@ -1644,8 +1688,7 @@ impl Byte {
 impl fmt::Debug for State {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("State")
-         .field("flags", &self.flags)
-         .field("insts", &self.insts)
+         .field("data", &self.data)
```
BurntSushi: Could you make this impl show something a bit more useful? It would be nice if it kept roughly the same information content as it was before.

BurntSushi:
OK, I have a few nits but this otherwise looks good! Thank you.

In the future, could you please get in touch or make a comment on the issue tracker before working on a non-trivial PR?

killercup:
Oh, yeah, I was working on that. But that's okay, your code is basically what I have written as well and then some more.

Just FYI, I originally had the idea to represent the two different sizes (i8 and i32) with a postfixed bit on the first byte, to allow i7 in one byte and i31 in 4 bytes, but I'm sure that while maybe looking more clever it would've been more error prone and probably slower to read in practice.

Now that I think about it, InstPtr is an u32, right? Can we correctly represent all possible deltas with i32s? (Or maybe I'm just missing something, it is pretty early here :))

BurntSushi:
> Now that I think about it, InstPtr is an u32, right? Can we correctly represent all possible deltas with i32s? (Or maybe I'm just missing something, it is pretty early here :))

Do you mean *encode* every delta as an `i32`? I think if you do that, you'll lose the compression benefits? I might be misunderstanding though.

BurntSushi:
Thanks @SeanRBurton!

BurntSushi:
> probably slower to read in practice

I think if the performance of reading these instruction pointers becomes a factor, then the DFA has already lost. :-) (One can only hope it will realize it, quit and fall back to the slow-but-faster-in-this-case Pike VM.)

SeanRBurton:
I guess he means that if a delta > i32::MAX then we can't represent it in 4 bytes. Maybe we should just say that insts.len() has to be <= i32::MAX to use the dfa.

SeanRBurton:
RE: https://github.com/rust-lang-nursery/regex/pull/226#issuecomment-217410925

I guess he means that if a delta > i32::MAX then we can't represent it in 4 bytes. Maybe we should just say that insts.len() has to be <= i32::MAX to use the dfa.

killercup:
I was only thinking about the upper bound, where this uses 4 u8's to build a i32. It's probably irrelevant in praxis, but since InstrPtrs are u32, there could in theory be a jump from u32::MIN to u32::MAX which the i32 delta can't represent. (Maybe just add a debug_assert? Are there overflow checks?)

Oh, and BTW, are there any benchmarks? Andrew mentioned in the original issue this could improve the class negation benchmark (IIRC).

BurntSushi:
> I guess he means that if a delta > i32::MAX then we can't represent it in 4 bytes. Maybe we should just say that insts.len() has to be <= i32::MAX to use the dfa.

Yup, that's done: https://github.com/rust-lang-nursery/regex/blob/master/src/dfa.rs#L71 --- It was necessary before since we limited ourselves to 32 bits anyway.

> there could in theory be a jump from u32::MIN to u32::MAX which the i32 delta can't represent. (Maybe just add a debug_assert? Are there overflow checks?)

Perhaps we should limit the DFA to `i32::MAX` instead.

> Oh, and BTW, are there any benchmarks? Andrew mentioned in the original issue this could improve the class negation benchmark (IIRC).

I haven't run them yet, but I will before the next release to make sure there are no regressions.

It will only impact the class negation benchmark if it brings memory usage down from 4-5MB or so to under 2MB. (When I last thought about, I didn't think this optimization was capable to doing that much.)

PR #230, opened 2016-05-07
regex 1.0

BurntSushi:
This is a series of commits corresponding to my _proposed_ changes for regex 1.0. It's still a work in progress and shouldn't be merged until an RFC is accepted.


alexcrichton:
🚀 Nice!

Would it make sense to draw up a short summary of the breaking changes from the current 0.1 release train?

BurntSushi:
@alexcrichton Oh yes, that will be in the RFC. I can add the relevant pieces to the commit messages too. It's a WIP. :-)

Amanieu:
You should also consider raising the minimum Rust version. For example, regex currently requires Rust 1.4 on Windows.

RustPowers:
@BurntSuchi API is ugly for me, could you wait a bit to 1.0? To begin with I would like to ask a few questions:
1. Is there a difference in performance at `.find("abc")` and `.bind_iter("abc").next().unwrap()`? Is it necessary to have a separate method to find the first occurrence?
2. What do you think about the integration to `std::str::pattern`? Then could work as follows:`string.contains(&regex)`, `string.matches(&regex)`, string.match_indeces(&regex) etc. The problem is that `Pattern` works only with type `str`, but in this case it is possible to change it to `Pattern<T>` (`Pattern<str>` and `Pattern<[u8]>`).

BurntSushi:
@RustPowers Please comment on the RFC instead: https://github.com/rust-lang/rfcs/pull/1620

(The `Pattern` trait is unstable.)

liigo:
Ping. https://github.com/rust-lang/rfcs/pull/1620 was proved one month ago. Any plan on regex 1.0?

BurntSushi:
Closing in favor of #310.

upsuper:
Is regex going to be part of std, or would it remain an independent crate that developers would still need to specify in `Cargo.toml` and download and build locally?

BurntSushi:
@upsuper There are no plans to move `regex` into std.

PR #232, opened 2016-05-15
Fix SIMD Fallback Import

killercup:
Fixes `cargo doc` and some other stuff.

r? @BurntSushi

### DIFF
--- a/src/literals.rs
+++ b/src/literals.rs
@@ -15,6 +15,7 @@ use memchr::{memchr, memchr2, memchr3};
 use syntax;
 
 use freqs::BYTE_FREQUENCIES;
+
 use simd_accel::teddy128::Teddy;
 
 /// A prefix extracted from a compiled regular expression.
--- a/src/simd_accel/teddy128.rs
+++ b/src/simd_accel/teddy128.rs
@@ -47,7 +47,7 @@ number of short patterns is far more likely.
 Faro and Kulekci published another paper[4b] that is conceptually very similar
 to [4a]. The key difference is that it uses the CRC32 instruction (introduced
 as part of SSE 4.2) to compute fingerprint values. This also enables the
-algorithm to work effectively on substrings as short at 7 bytes with 4 byte
+algorithm to work effectively on substrings as short as 7 bytes with 4 byte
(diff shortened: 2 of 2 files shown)


killercup:
I also added some details to the references since I was looking through the papers anyway. (It follows the APA style used by Google Scholar.)

src/literals.rs
```diff
@@ -15,8 +15,13 @@ use memchr::{memchr, memchr2, memchr3};
 use syntax;
 
 use freqs::BYTE_FREQUENCIES;
+
+#[cfg(feature = "simd-accel")]
```
BurntSushi: I think you don't actually need this `cfg` switch, right? The `cfg` switch is done in `lib.rs` so that `simd_fallback` should never actually be explicitly used.

killercup: You still import it as `simd_fallback`, though. Might be a good idea to provide a shim module in `lib.rs` depending on the feature flag (the shim in `simd_accel` only deals with the target arch feature).

BurntSushi: Oh, I see. I thought I was using `path`, but maybe it was somewhere else. Maybe this should be in `lib.rs`?

```rust
#[cfg(not(feature = "simd-accel"))]
#[path = "simd_fallback/mod.rs"]
mod simd_accel;
```

I think that would work? Or something equivalent would be fine. I was just trying to stop the use of `cfg` switching everywhere.

killercup: Yeah, that should work. Defining the shim in `lib.rs` is far better than using cfg here. I'll change this.

killercup:
Rebased.

BurntSushi:
@killercup Could you rebase this against master? Sorry for the churn. Thanks!

killercup:
@BurntSushi no problem! Rebased on master.

BurntSushi:
Awesome, thanks so much! :D

BurntSushi:
@killercup Hmm, it looks like this got merged into the `simd-teddy` branch. Whoops. Could you resubmit against `master` please?

killercup:
Oh, Github didn't change the target branch :( just one second!
Andrew Gallant <notifications@github.com> schrieb am Do., 19. Mai 2016 um
01:39:

PR #238, opened 2016-05-20
Add initial fuzzing support

SeanRBurton:
I also had to fix the bad behaviour of the Expr arbitrary instance to avoid getting CompiledTooBig all over the place. I don't yet implement zero-width assertions, but it would be good to try and work them in.


BurntSushi:
Thanks!

Could you please describe what this PR is doing? Examples would be helpful.

regex-fuzz/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+fn random_char_in_range(low: char, high: char) -> char {
+    let (low, high) = (low as u32, high as u32);
+    let l0 = low;
+    let h0 = ::std::cmp::min(high, 0xD7FF);
+    let l1 = ::std::cmp::max(low, 0xE000);
+    let h1 = ::std::cmp::min(high, 0x10FFFF);
+    let s0 = if h0 < l0 {0} else {h0 - l0 + 1};
+    let s1 = if h1 < l1 {0} else {h1 - l1 + 1};
+    let mut rng = rand::thread_rng();
+    let x: u32 = if Range::new(0, s1 + s0).ind_sample(&mut rng) < s0 {
```
huonw: FWIW, since these `Range`s aren't being reused, I believe they can use `gen_range` with no penalty, e.g. `rng.gen_range(0, s1 + s2) < s0`.

SeanRBurton:
Just run `cargo test` in the regex-fuzz directory.

BurntSushi:
I'd like a description of the testing strategy please.

SeanRBurton:
So far I just generate a random regex and a corresponding matching string, and check that the string really does match.

BurntSushi:
Closing due to inactivity.

PR #249, opened 2016-06-15
Replace `^` (xor) on bools with `!=` and `==`.

SeanRBurton:
Just a tiny refactoring for readability and concision.


BurntSushi:
I agree, this is better. Thanks!

PR #253, opened 2016-06-16
modularize the C interface

baby-bell:
Splitting each part of the interface into its own module makes the interface slightly easier to comprehend.


BurntSushi:
Hmm. I guess I kind of like this better. Should the error type/functions be in `rure.go` though?

baby-bell:
Error handling in C is usually not very elegant, unfortunately. There might be a better way to expose errors in the interface, but I'm not sure that's within the scope of this PR.

BurntSushi:
Sorry, I meant to type `rure.rs`. I'm fine with this though. Thanks!

Issue #257, opened 2016-06-21
Option::unwrap() called on None-value in Regex::new()


lukaslueg:
AFL found this regex which causes `Regex::new()` in `regex-0.1.71` to call `Option::unwrap()` on a None-value, causing a panic

```
r"[^\D\d]"
```

BurntSushi:
@lukaslueg Thank you for filing all these bugs. I can't wait to fix them. :-)

BurntSushi:
This particular bug is a result of two things:

1. `[^\D\d]` is actually an empty character class, since `[\D\d]` says "match anything" and `[^\D\d]` says "match !anything."
2. The compiler assumes that empty character classes are impossible.

I'd like to fix this by more explicitly banning empty character classes. Other regex engines actually allow them and treat them as "never match anything." It seems more sensible (and more conservative) to just ban them outright.

Issue #251, opened 2016-06-16
Regex::split() crashes


lukaslueg:
AFL found the following crash in `regex-0.1.71` using `rustc-1.11.0-dev 2016-06-13`

```
regex::Regex::new("a$").unwrap().split("a").last();
```
```
...panicked at 'begin <= end (1 <= 0) when slicing `a`...
```

BurntSushi:
I think this is caused by the same underlying issue in #255.

Issue #250, opened 2016-06-16
Fuzzed regex causes assertion-failure


lukaslueg:
AFL found this regex which causes an assertion failure `!ranges.is_empty()` in `regex-0.1.71/src/compile.rs:421` while `Regex::new()` using `rustc-1.11.0-dev 2016-06-13`. In release-mode it causes an `index 18446744073709551615 out of range for slice of length 0`.

```
/=(?-u:\?(?:\[\[:\[\[[^|\]]*(?:\|(?-u:\?[\[[]*(?:\|(?:[^|\]]*))2:[^|\]]*))25[σף]]*(?:\|(?:[^|\]]*)0?:[^|\]]*))25[0])+
```

Beware the special characters. In Base64 it's Lz0oPy11Olw/KD86XFtcWzpcW1xbW158XF1dKig/Olx8KD8tdTpcP1tcW1tdKig/Olx8KD86W158
XF1dKikpMjpbXnxcXV0qKSkyNVvPg9ejXV0qKD86XHwoPzpbXnxcXV0qKTA/OltefFxdXSopKTI1
WzBdKX8r

BurntSushi:
Nice. This is a bug because Unicode literals should be disallowed in character classes when Unicode mode is disabled, but they weren't. A smaller reproducible example is `(?-u)[☃]`.

Issue #246, opened 2016-06-09
RegexBuilder should document that it does not incorporate flags into string-representation


lukaslueg:
AFAICS `regex::RegexBuilder` does not incorporate flags into the string representation of the final regex. 

```rust
let re = regex::RegexBuilder::new("foo").case_insensitive(true).compile().unwrap();
assert_eq!(re.as_str(), "(?i)foo");
```
panics because `"foo" != "(?i)foo"`.

Am I confused about something here? Otherwise this should at least be documented :-)

BurntSushi:
The string representation is the exact pattern given. The docs on the Display impl and as_str method say:

> Shows the original regular expression.

How can the documentation be improved?

lukaslueg:
I was under the impression that the string representation would by equivalent to the pattern and the flags set on RegexBuilder. That is, given `::new('foo')` and `.case_insensitive(true)`, the string representation is `(?i)foo`.

BurntSushi:
Right, but it's not. I'm asking: how might the docs be improved so that you don't get that impression?

lukaslueg:
Maybe a note in `RegexBuilder::compile()`:

````
Consume the builder and compile the regular expression.
Note that the resulting `Regex::as_str()` returns the verbatim pattern given to `RegexBuilder::new()` without incorporating the flags set thereafter.
```

PR #262, opened 2016-07-10
fix several small bugs found from fuzzing

### LINKED ISSUE #255: Endless loop in Regex::captures_iter()
AFL found this simple regex that causes `Regex::captures_iter()` in `regex-0.1.71` to enter into an endless loop.

```
regex::Regex::new("a$").unwrap().captures_iter("a").last()
```

BurntSushi:
Fixes #255, #257, #246, #251, #250 and #241 (many of which were found by @lukaslueg applying AFL to this library---many thanks!)

The commit messages contain the gory details.


BurntSushi:
These fixes are in `regex 0.1.72` on crates.io.

PR #278, opened 2016-09-04
Replace the slow loop in Teddy with Aho-Corasick.

BurntSushi:
This can have a significant impact on performance when the match rate is
low (common) and the number of patterns is high (e.g., for a case
insensitive search).


BurntSushi:
As a meta comment: it's kind of silly I didn't just do this in the first place. There's almost no downside (sans a very small additional amount of memory required) and the code is far simpler than it was before.

Issue #279, opened 2016-09-11
Named capture group introspection not exposed by the C API


davidblewett:
The C API exposes an iterator over all captures (`rure_iter_next_captures`), and can return the index of a given capture group name (`rure_capture_name_index`). The Rust Regex struct has a `capture_names` iterator, but that doesn't appear to be exposed at the C level.

I don't see a way to retrieve the capture group name while iterating over the set of captures (i.e., given the index of a particular capture give me the corresponding name, hypothetically `rure_capture_index_name`), or to return the list of capture group names (`rure_iter_capture_groups`, for example).

BurntSushi:
Ah yup, it looks like exposing the `capture_names` iterator in some form in the C API should fix this.

davidblewett:
I came upon this while working on Python bindings ( current WIP here: https://github.com/rust-lang-nursery/regex/compare/master...davidblewett:rure-python?expand=1 ). I wanted to implement the `groupdict` method, similar to Python's `re.MatchObject`: https://docs.python.org/2/library/re.html#re.MatchObject.groupdict . 

I'm new to Rust, and have never really used C. I can take a crack at it some time this week.

BurntSushi:
@davidblewett That'd be awesome! I'm not a C expert either but I'd be happy to help mentor you if you'd like. I'm on IRC in the evenings during the week.

BurntSushi:
Fixed in #282

PR #282, opened 2016-09-12
Capi capture names

davidblewett:
Refs #279.

This isn't ready, but I need some help. I'm stumped by how Rust is handling strings. In my tests with these changes, I see this when using the API:
```
Entering rure_iter_capture_names_next
Iterator last_index: 0
capture_names length: 1
last: 1
Selected name: last
CString cast: 'last'
Entering rure_iter_capture_names_next
Iterator last_index: 1
capture_names length: 1
> /home/dblewett/.virtualenvs/rure/site-packages/rure/regex.py(172)find_captures()
-> match = ffi.new('rure_match *')
(Pdb) all_names
['l          ']
```

I'm not sure if this implementation is correct and the Python FFI is not working, or if the implementation is flawed. What I get back from each call to `rure_iter_next_captures` is either an empty string (when using `ffi.new('char[]', [])`), a single character (when using `ffi.new('char[]', '')`), or a single character followed by padding (when using `ffi.new('char[]', ' ' * 20)`).

### DIFF
--- a/regex-capi/src/rure.rs
+++ b/regex-capi/src/rure.rs
@@ -5,7 +5,7 @@ use ::libc::{c_char, size_t};
 
 use ::std::collections::HashMap;
 use ::std::ops::Deref;
-use ::std::ffi::CStr;
+use ::std::ffi::{CStr, CString};
 use ::std::ptr;
 use ::std::str;
 use ::std::slice;
@@ -42,6 +42,11 @@ pub struct Iter {
     re: *const Regex,
     last_end: usize,
     last_match: Option<usize>,
+}
+
+pub struct IterCaptureNames {
+    capture_names: bytes::CaptureNames<'static>,
+    name_ptrs: Vec<*mut c_char>,
 }
 
 impl Deref for Regex {
@@ -226,6 +231,67 @@ ffi_fn! {
 }
 
 ffi_fn! {
+    fn rure_iter_capture_names_new(
+        re: *const Regex,
+    ) -> *mut IterCaptureNames {
+        let re = unsafe { &*re };
+        Box::into_raw(Box::new(IterCaptureNames {
+            capture_names: re.re.capture_names(),
+            name_ptrs: Vec::new(),
+        }))
+    }
+}
+
+ffi_fn! {
+    fn rure_iter_capture_names_free(it: *mut IterCaptureNames) {
+        unsafe {
+            let it = &mut *it;
+            while let Some(ptr) = it.name_ptrs.pop(){
+                CString::from_raw(ptr);
+            }
+            Box::from_raw(it);
+        }
+    }
+}
+
+ffi_fn! {
+    fn rure_iter_capture_names_next(
+        it: *mut IterCaptureNames,
+        capture_name: *mut *mut c_char,
+    ) -> bool {
+        if capture_name.is_null() {
+            return false;
+        }
+
+        let it = unsafe { &mut *it };
+        let cn = match it.capture_names.next() {
+            // Top-level iterator ran out of capture groups
+            None => return false,
+            Some(val) => {
+                let name = match val {
+                    // inner Option didn't have a name
+                    None => "",
+                    Some(name) => name
+                };
+                name
+            }
+        };
+
+        unsafe {
+            let cs = match CString::new(cn.as_bytes()) {
+                Result::Ok(val) => val,
+                Result::Err(err) => return false
+            };
+            let ptr = cs.into_raw();
+            it.name_ptrs.push(ptr);
+            *capture_name = ptr;
+        }
+        true
+
+    }
+}
+
+ffi_fn! {
     fn rure_iter_new(
         re: *const Regex,
     ) -> *mut Iter {
@@ -387,5 +453,3 @@ ffi_fn! {
         options.dfa_size_limit = limit;
     }
 }
-
-
--- a/regex-capi/Cargo.toml
+++ b/regex-capi/Cargo.toml
@@ -17,4 +17,4 @@ crate-type = ["staticlib", "dylib"]
 
 [dependencies]
 libc = "0.2"
-regex = { version = "0.1.69", path = ".." }
+regex = { version = "0.1.77", path = ".." }
--- a/regex-capi/ctest/test.c
+++ b/regex-capi/ctest/test.c
@@ -195,6 +195,60 @@ done:
     return passed;
 }
 
+bool test_iter_capture_name(char *expect, char *given) {
+    bool passed = true;
+    if (strcmp(expect, given)) {
+        if (DEBUG) {
+            fprintf(stderr,
+                    "[test_iter_capture_name] expected first capture name '%s' "
+                    "got '%s'\n",
+                    expect, given);
+        }
+        passed = false;
+    }
+    return passed;
+}
+
+bool test_iter_capture_names() {
+    bool passed = true;
+
+    char *name;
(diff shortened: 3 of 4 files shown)


BurntSushi:
Hmm, it's hard to tell what exactly is going wrong... Here's my advice:

* Forget about Python for the moment and write a C unit test. There are many existing C unit tests here: https://github.com/rust-lang-nursery/regex/blob/master/regex-capi/ctest/test.c (When you define a new one, make sure to add it to the `main` function at the bottom.)
* The C iterator should precisely match the semantics of the Rust [`CaptureNames`](https://doc.rust-lang.org/regex/regex/struct.CaptureNames.html) iterator. This means yielding an iteration even for unnamed capture groups. You could return true for those captures, but ensure the string given back to the caller is zero-length.
* You should be able to embed the `CaptureNames` Rust iterator type directly into your `IterCaptureNames` type (perhaps even using `CaptureNames` directly). That way, you don't have to re-implement the iterator.
* The API generally uses length based strings instead of C strings. C strings are undoubtedly more convenient though, and for this particular use case, I think it's probably the right choice.

Sorry I can't diagnose your precise problem, but I'd attack this by ignoring Python and trying to make the code a bit simpler.

Thank you for working on this!

davidblewett:
Okay, point taken about not using Python's cffi. I changed to calling the `CaptureNames` iterator, and added a unit test. What I don't get now is that the test fails on the 2nd invocation of `rure_iter_capture_names_next` (when trying to print the string returned).

```
kell% LD_LIBRARY_PATH="../target/debug:$LD_LIBRARY_PATH" ./test
PASSED: test_is_match
PASSED: test_shortest_match
PASSED: test_find
snowman
PASSED: test_captures
PASSED: test_iter
Entering rure_iter_capture_names_next
Iterator last_index: 0
capture_names length: 3
month: 2
day: 3
year: 1
Selected name:
CString cast: ''
[test_iter_capture_names] Received name '(null)'
Entering rure_iter_capture_names_next
Iterator last_index: 1
capture_names length: 3
month: 2
day: 3
year: 1
Selected name: year
CString cast: 'year'
[test_iter_capture_names] Third name? 'True'
zsh: segmentation fault (core dumped)  LD_LIBRARY_PATH="../target/debug:$LD_LIBRARY_PATH" ./test

```

davidblewett:
@BurntSushi : I *think* I'm finally grokking the lifetime aspect of this change. However, I think to fully support it is going to require more changes than I'm comfortable making:
* The `ffi_fun` macro doesn't appear to support the lifetime specifier:
```
   Compiling rure v0.1.1 (file:///home/dblewett/src/regex/regex-capi)
src/rure.rs:237:35: 237:36 error: no rules expected the token `<`
src/rure.rs:237     fn rure_iter_capture_names_new<'a>(
                                                  ^
```
* I tried just manually specifying the usual no_mangle, pub extern dance, but got this compiler warning:
```
   Compiling rure v0.1.1 (file:///home/dblewett/src/regex/regex-capi)
src/rure.rs:238:5: 247:6 warning: generic functions must be mangled, #[warn(no_mangle_generic_items)] on by default
src/rure.rs:238     pub extern fn rure_iter_capture_names_new<'a>(
```

BurntSushi:
That warning is there for good reason. Lifetime parameters are type parameters, and you can't expose generics in a C ABI. You have to remove it. (You might do it by using `bytes::CaptureNames<'static>`, which will require a transmute when you build it.)

I'm not sure about your other problems. I'll need to have a closer look for that, which I can do later today thanks to your C test. :-)

BurntSushi:
Just to say a bit more about `'static`... When it comes to exposing a C ABI, we completely lose all of the guarantees enforced by the Rust compiler, which means it is now up to the C programmer to maintain proper lifetimes. For example, one of the first things I looked for in your docs was something like this (and was very pleased to see it):

> An rure_iter_capture_names value may not outlive its corresponding rure, and should be freed before its corresponding rure is freed.

This is precisely what that `'a` lifetime enforced at compile time inside of Rust. By forcing it to be `'static`, we're explicitly throwing away all of Rust's guarantees and putting them on the C programmer.

(Technically, I don't think an `rure_iter_capture_names` needs to actually be freed before its corresponding `rure`, but it certainly can't be used after its corresponding `rure` is freed.)

regex-capi/src/rure.rs
```diff
@@ -221,11 +226,77 @@ ffi_fn! {
...
+        // Fill in pointer
+        unsafe {
+            let cs = match CString::new(cn.as_bytes()) {
+                Result::Ok(val) => {
+                    val
+                },
+                Result::Err(err) => return false
+            };
+            println!("CString cast: '{}'", cs.to_str().unwrap());
+            *capture_name = *cs.into_raw();
```
davidblewett: I think the problem is here; the `CString` is getting freed when the `unsafe` block completes, thus the C unit test triggers a segfault trying to read the memory (similar to https://users.rust-lang.org/t/correct-way-to-implement-a-function-which-returns-a-c-string/315/3 ). I think the solution is to use a `Box`, but I'm unsure of the details of how to do that.

davidblewett:
@BurntSushi : Figured the problem out; it all boiled down to 4702e69abe1f12d6bff58a9a87ee92894eaa28c7  it appears. The C unit test now passes correctly. Will test in Python's cffi next.

davidblewett:
I've rebased my rure-python branch on top of this, and it's working great. https://github.com/davidblewett/regex/tree/rure-python

davidblewett:
@BurntSushi : Would like to get this merged; it appears everything looks in place. Tests are passing, and I've cleaned up the debug statements I had in from earlier. Tried to condense the code back down so that it is easier to follow as well.

I have a functioning Python binding I would like to issue a PR for that is based on this.

BurntSushi:
@davidblewett This all looks good. Would you mind squashing this down to one commit and rebase it on to current master with a descriptive commit message? After that it should be good to go. Thanks so much for doing this! Very nice work. :-)

> I have a functioning Python binding I would like to issue a PR for that is based on this.

A PR to this repo? Or somewhere else?

davidblewett:
Yeah, I can squash it. Will try to do that tomorrow.

The branch I've been working on Python support is here: https://github.com/davidblewett/regex/compare/capi-capture-names...davidblewett:rure-python .

BurntSushi:
To be clear, I'm not sure I'm comfortable with adding other language bindings. Maintaining a canonical C API is one thing, but opening the door to every other language doesn't seem like a good idea. :(

davidblewett:
@BurntSushi : went ahead and squashed the commits to a single and updated the PR.

davidblewett:
> To be clear, I'm not sure I'm comfortable with adding other language bindings.

Yeah, I can understand that. I'm trying to find the best route to do this ( I asked on the web forum awhile back: https://users.rust-lang.org/t/binding-regex-from-python/7253 ).

The problem that I'm running into is that I couldn't figure out how to have a simple "wrapper" crate of the `rure` crate. Every time I tried, the C functions always returned garbage. By having it inside the `rure` crate itself, it simplified a lot of the Python packaging side. I suppose what I can do is instead of trying to use the `rust_setuptools` / `RustDistribution` stuff to try to automate building the crate, I can just dictate that you need to drop the `rure` static library in a given directory and we'll build the wheel based on that.

BurntSushi:
All right, all set! Thanks so much for your patience and pushing through this.

I'm not sure what the in-crate/out-crate issue is, but if you open a new issue with more details I'd be happy to help!

PR #283, opened 2016-09-13
Fix invariant in comment

aweinstock314:
`(end >= end)` is trivial, `(end >= start)` is likely intended.


BurntSushi:
Thanks @aweinstock314!

PR #288, opened 2016-10-19
Add RegexSet functionality to C API

tiehuis:
These functions implement a C interface to the RegexSet api.

Some notes:
- These do not include start offsets as the standard regex functions
  do. The reason being is down to how these are implemented in the core
  regex crate. The RegexSet api does not expose a public is_match_at
  whilst the Regex api does. I would actually like to change this if possible
  as the inconsistency this makes between `rure` and `rure_set` types is
  not great.
- This only tests a complete compile/match mainly for sanity. One or
  two more tests targeting the specific areas would be preferred.
- Set matches take a mutable array to fill with results. This is more
  C-like and allows the caller to manage the memory on the stack if
  they want.

Wanted to make a request before targeting these changes to clarify.

### DIFF
--- a/regex-capi/include/rure.h
+++ b/regex-capi/include/rure.h
@@ -468,7 +468,10 @@ void rure_options_dfa_size_limit(rure_options *options, size_t limit);
  */
 rure_set *rure_compile_set(const uint8_t **patterns,
                            const size_t *patterns_lengths,
-                           size_t patterns_count, rure_error *error);
+                           size_t patterns_count,
+                           uint32_t flags,
+                           rure_options *options,
+                           rure_error *error);
 
 /*
  * rure_set_free frees the given compiled regular expression set.
@@ -486,7 +489,8 @@ void rure_set_free(rure_set *re);
  * useful. UTF-8 is even more useful. Other text encodings aren't supported.
  * length should be the number of bytes in haystack.
  */
-bool rure_set_is_match(rure_set *re, const uint8_t *haystack, size_t length);
+bool rure_set_is_match(rure_set *re, const uint8_t *haystack, size_t length,
+                       size_t start);
 
 /*
  * rure_set_matches compares each regex in the set against the haystack and
@@ -507,7 +511,7 @@ bool rure_set_is_match(rure_set *re, const uint8_t *haystack, size_t length);
  * caring which, use rure_set_is_match.
  */
 bool rure_set_matches(rure_set *re, const uint8_t *haystack, size_t length,
-                      bool *matches);
+                      size_t start, bool *matches);
 
 /*
  * rure_set_len returns the number of patterns rure_set was compiled with.
--- a/regex-capi/include/rure.h
+++ b/regex-capi/include/rure.h
@@ -473,7 +473,7 @@ rure_set *rure_compile_set(const uint8_t **patterns,
 /*
  * rure_set_free frees the given compiled regular expression set.
  *
- * This must be called at most once.
+ * This must be called at most once for any rure_set.
  */
 void rure_set_free(rure_set *re);
 
@@ -490,8 +490,10 @@ bool rure_set_is_match(rure_set *re, const uint8_t *haystack, size_t length);
 
 /*
  * rure_set_matches compares each regex in the set against the haystack and
- * returns an array of bools which correspond to if a match was found for
- * the specified regex.
+ * modifies matches with the match result of each pattern. Match results are
+ * ordered in the same way as the rure_set was compiled. For example,
+ * index 0 of matches corresponds to the first pattern passed to
+ * `rure_compile_set`.
  *
  * haystack may contain arbitrary bytes, but ASCII compatible text is more
  * useful. UTF-8 is even more useful. Other text encodings aren't supported.
@@ -505,7 +507,12 @@ bool rure_set_is_match(rure_set *re, const uint8_t *haystack, size_t length);
  * caring which, use rure_set_is_match.
  */
 bool rure_set_matches(rure_set *re, const uint8_t *haystack, size_t length,
-                      const bool *matches);
+                      bool *matches);
+
+/*
+ * rure_set_len returns the number of patterns rure_set was compiled with.
+ */
+size_t rure_set_len(rure_set *re);
 
 /*
  * rure_error_new allocates space for an error.
--- a/regex-capi/include/rure.h
+++ b/regex-capi/include/rure.h
@@ -29,6 +29,13 @@ extern "C" {
 typedef struct rure rure;
 
 /*
+ * rure_set is the type of a set of compiled regular expressions.
+ *
+ * A rure can be safely used from multiple threads simultaneously.
+ */
+typedef struct rure_set rure_set;
+
+/*
  * rure_options is the set of non-flag configuration options for compiling
  * a regular expression. Currently, only two options are available: setting
  * the size limit of the compiled program and setting the size limit of the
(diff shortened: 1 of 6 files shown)


regex-capi/include/rure.h
```diff
@@ -165,7 +172,7 @@ rure *rure_compile(const uint8_t *pattern, size_t length,
 /*
  * rure_free frees the given compiled regular expression.
  *
- * This must be called at most once.
+ * This must be called at most once for any rure.
  */
```
tiehuis: This is slightly more clear I think. `rure_set_free` should match this.

regex-capi/include/rure.h
```diff
@@ -447,6 +454,60 @@ void rure_options_size_limit(rure_options *options, size_t limit);
...
+ *
+ * haystack may contain arbitrary bytes, but ASCII compatible text is more
+ * useful. UTF-8 is even more useful. Other text encodings aren't supported.
+ * length should be the number of bytes in haystack.
+ */
+bool rure_set_is_match(rure_set *re, const uint8_t *haystack, size_t length);
+
+/*
+ * rure_set_matches compares each regex in the set against the haystack and
+ * returns an array of bools which correspond to if a match was found for
```
BurntSushi: I think "to if a" -> "to a"

regex-capi/include/rure.h
```diff
@@ -447,6 +454,60 @@ void rure_options_size_limit(rure_options *options, size_t limit);
...
+ *
+ * matches must be greater than or equal to the number of patterns the
+ * rure_set was compiled with.
+ *
+ * Only use this function if you specifically need to know which regexes
+ * matched within the set. To determine if any of the regexes matched without
+ * caring which, use rure_set_is_match.
+ */
+bool rure_set_matches(rure_set *re, const uint8_t *haystack, size_t length,
+                      const bool *matches);
```
BurntSushi: Should this be `bool *matches` instead of `const bool *matches`? (Since the data pointed to by `matches` will be modified.)

BurntSushi:
> These functions implement a C interface to the RegexSet api.

Thanks for doing this! At a first glance, this looks pretty good with a few nits.

So one thing that we should consider before merging this is how to incorporate [`rure_options`](https://github.com/rust-lang-nursery/regex/blob/3cfef1e79d135a3e8a670aff53e7fabef453a3e1/regex-capi/include/rure.h#L31-L40) into `RegexSet` compilation. We probably also need to think about [flags](https://github.com/rust-lang-nursery/regex/blob/3cfef1e79d135a3e8a670aff53e7fabef453a3e1/regex-capi/include/rure.h#L42-L61) as well.

The core problem you'll run into is that neither of these things is actually exposed in the `regex` crate proper.

Maybe we can come up with a straw man proposal for the C API and convince ourselves that it can be added later, so that we can merge this sooner.

> These do not include start offsets as the standard regex functions
do. The reason being is down to how these are implemented in the core
regex crate. The RegexSet api does not expose a public is_match_at
whilst the Regex api does. I would actually like to change this if possible
as the inconsistency this makes between rure and rure_set types is
not great.

The inconsistency is in the `regex` crate as well.

The inconsistency exists for a reason. Getting the match offsets is hard.

> This only tests a complete compile/match mainly for sanity. One or
two more tests targeting the specific areas would be preferred.

This is probably OK. Mostly, I think the C tests should just make sure that FFI is working. Testing the semantics of RegexSet is better suited for the `regex` crate proper. But I won't stop you from adding more tests. :-)

> Set matches take a mutable array to fill with results. This is more
C-like and allows the caller to manage the memory on the stack if
they want.

Perfect.

tiehuis:
> So one thing that we should consider before merging this is how to incorporate rure_options into RegexSet compilation. We probably also need to think about flags as well.

> The core problem you'll run into is that neither of these things is actually exposed in the regex crate proper.

> Maybe we can come up with a straw man proposal for the C API and convince ourselves that it can be added later, so that we can merge this sooner.

I had a bit more of a look into `RegexSet` and would the following work for the actual crate implementation. Correct me if I've overlooked something:
 * Extend `RegexBuilder` with two new functions `from_set` and `compile_set`. `from_set` just takes a slice of patterns instead of a single. `compile_set` is similar but maps using `From::RegexSet` instead of `From::Regex`. This seems compatible since the types `Regex` and `RegexSet` just wrap an `Exec` type, with the bulk of the work already implemented within `Exec`.

And actually, just that seems like it would be sufficient for the lack of options to me. This can easily be exposed to the C API. This would be a backwards compatible API change, so would adjust the versioning however you are doing it. A corresponding `compile_must` for the `rure_set` type may be of use then at this point, too.

> The inconsistency is in the regex crate as well.

> The inconsistency exists for a reason. Getting the match offsets is hard.

If that is the case and I can just omit the start argument from the match functions.

> This is probably OK. Mostly, I think the C tests should just make sure that FFI is working. Testing the semantics of RegexSet is better suited for the regex crate proper. But I won't stop you from adding more tests. :-)

I'll just add a regex failure test for sanity.

Thanks for the quick response, let me know if the `RegexBuilder` modifications outline sounds fine and if so, I'll implement those changes hopefully tomorrow.

tiehuis:
I noted previously that we would need some extra functions in the regex crate. This seems like an incompatible change however (after reading the regex 1.0) RFC (at least in the short-term as you previously said). It looks like enough of the interface is public that we can just construct the `RegexOptions` ourselves and call use the `ExecBuilder` directly, though, so should still be feasible.

BurntSushi:
Sorry I'm dragging, but the past few days have been action packed (including a trip to the animal ER). Using `RegexOptions` and `ExecBuilder` is regrettable (they are technically exposed, but not really part of the public API), but I'm OK with it *as long as* there is a straight-forward way to expand the C API.

These options do need to get exposed in the `regex` crate proper though too. Your `from_set` and `compile_set` idea is one I've tossed around too, but my problem with it is that if you build with `from_set`, then it'd be nice to not call `compile`. I guess we could just return an error instead of ruling it out statically.

To be clear, I'm happy merging this PR without these options in the C API, as long as there's a reasonable path to adding them in the future.

tiehuis:
Not a problem.

I'd like to get something thought out in that would minimize any breaking future api changes. I'd also like to avoid modifying the regex crate if possible right now.

With these goals in mind, here is what I suggest.

 * Add options to the c api regex set by using `RegexOptions` directly. This would be an implementation detail and can be adjusted to use a proper interface when one lands in the core crate.

 * Add a dummy `start` variable to the `rure_set` match functions. This is supported internally but no public way of acessing this is present. This unused field can be updated when support lands in the core crate. For the moment, it would require callers of this to use a `0` value to ensure future compatibility.

This should be flexible enough for future improvements without any breaking api changes.

Also, as an aside regarding the `RegexBuilder` and its compatibility with `RegexSet`. The cleanest way I see is to just create a seperate builder for a `RegexSet` itself. That is a topic for another time though.

I'll be offline till Monday but that day (which is thankfully a public holiday for me) I'll implement these and see how it all fits.

BurntSushi:
Sounds good. Two things:

1. You'll need to add `RegexOptions` to the `internal` module, which is in `src/lib.rs`.
2. You can implement the `start` functionality by using `Exec` and calling `searcher`.

In fact, I think both of these things means that you end up working around the use of `RegexSet` proper entirely.

tiehuis:
Just a note that I've finished adding these features.

BurntSushi:
@tiehuis Thank you!

PR #293, opened 2016-11-05
Use an out-of-tree implementation of Teddy128.

jneem:
This switches from the in-tree implementation of Teddy to the one in the "teddy" crate, which is a bit faster (e.g. it supports AVX2). I've attached two outputs of "cargo benchcmp," one compiled with AVX2 support and one without. Even people without AVX2 support get a little boost :)

[benchcmp_ssse3.txt](https://github.com/rust-lang-nursery/regex/files/573103/benchcmp_ssse3.txt)
[benchcmp.txt](https://github.com/rust-lang-nursery/regex/files/573104/benchcmp.txt)


BurntSushi:
Thanks @jneem! I look forward to the perf improvements. :-)

I have three concerns. The first one is that this fails on Rust 1.3 as-is. That means it can't be part of regex `0.1`. I hope to be releasing `0.2` soon, after which we could move forward. (Or if it's easy to make `teddy` work on Rust 1.3 or newer, that'd be fine too.)

The second concern is: does your `teddy` crate work with only SSSE3? Or does it now also require SSE4.1?

With that said, my primary concern here is that the Teddy implementation is a core piece of matching functionality, and I expect its importance to only increase as SIMD gets stabilized. While I have a lot of confidence in Teddy, I think nobody (besides maybe the Hyperscan folks) has a lot of experience with it yet. Because of that and its critical role in the regex crate, I as the maintainer feel slightly uncomfortable that I can't fix bugs as they arise. With that in mind, I have two questions for you:

1. What are your long term plans for the `teddy` crate?
2. Would you be amenable to sharing commit access and crates.io ownership with myself and the Rust Project Developers team? (i.e., The same owners of the `regex` crate, see `cargo owner -l`.)

On an unrelated note, I am also somewhat saddened to see that there is no attribution to my work anywhere in your repo. :-/

jneem:
Thanks for the comments!

1. Rust 1.3 support looks possible, but a little annoying. Since rust 1.3 chokes on `extern "platform-intrinsic"`, I would have to split `teddy_simd.rs` into a separate crate, which would be a conditional dependency. Given that teddy currently compiles with rust 1.4, I'm inclined to give up on rust 1.3 and just wait for regex 0.2.

2. SSSE3 is enough ([this](https://github.com/jneem/teddy/blob/7f1cef60ed51122fcfe9c3a79f57e0d3a32d8b3b/src/x86/teddy_simd.rs#L98) is the only place SSE4.1 is used, if available)

3. I'd be more than happy to share (or even just give up) ownership of the crate. The main reason I forked it was so that it would be easier to iterate on, and not to establish any sort of personal fiefdom :)

My long-term plan for the `teddy` crate is to "finish" it. That means support for stable rustc, AVX-512, and maybe ARM. The main blocker, of course, is stable support for SIMD in rust.

PS: I'm really sorry about the lack of attribution. I added a bit to the README, and I'm happy to do so elsewhere also...

BurntSushi:
> Rust 1.3 support looks possible, but a little annoying. Since rust 1.3 chokes on extern "platform-intrinsic", I would have to split teddy_simd.rs into a separate crate, which would be a conditional dependency. Given that teddy currently compiles with rust 1.4, I'm inclined to give up on rust 1.3 and just wait for regex 0.2.

I agree. My plan is that `0.2` is imminent. I'll merge when I do that.

> SSSE3 is enough (this is the only place SSE4.1 is used, if available)

Awesome thanks!

> I'd be more than happy to share (or even just give up) ownership of the crate. The main reason I forked it was so that it would be easier to iterate on, and not to establish any sort of personal fiefdom :)

Haha, I'm not too worried about personal fiefdoms! Just my own ability to iterate. If you're happy with it, then I think just adding the owners of `regex` to the `teddy` crate and repos would be great!

> PS: I'm really sorry about the lack of attribution. I added a bit to the README, and I'm happy to do so elsewhere also...

Thanks! No worries. :-)

jneem:
> I think just adding the owners of regex to the teddy crate and repos would be great!

Do you mean the members of rust-lang-nursery? I'm happy to do that, but having a little trouble. On github, I can figure out how to add individual collaborators, but I can't figure out how to add a whole organization. On crates.io, I'm running into [this](http://doc.crates.io/crates-io.html#github-permissions) permissions problem. Anyway, I've added you to both.

BurntSushi:
Oh interesting, thank you! I managed to add `github:rust-lang-nursery:regex-owners` but couldn't add `github:rust-lang:regex`. I think that's fine for now. We can straighten out the owners once we move to `rust-lang` proper.

BurntSushi:
I've had more time to review the `teddy` crate in depth, and there are a few things I'd like to see changed before `regex` starts using it:

1. The documentation needs work. At the very least, the crate should be compiled with `![deny(missing_docs)]`. The docs should also make it very clear that nothing works at all without at least SSSE3 enabled.
2. The `teddy` crate uses both the `asm` and `associated_consts` nightly features, which AFAIK have no clear path to stabilization at present. We're working on getting SIMD stabilized, but I'd otherwise prefer to focus on things that stable Rust supports, even if we wind up with something slightly sub-optimal.
3. I feel like the code is very complicated, partially because of its heavy use of generics. It's hard for me to understand. I would rather violate DRY personally, especially for something internal.

I started doing some of this myself so that I could submit a PR, but I realized that I was on a path to rewriting the entire crate, which didn't seem like a good idea without speaking with you first.

I'm sorry I'm being so picky, but given this is a core part of regex matching, I have to be 100% confident that I can maintain this code.

I'm going to close this PR. I think we should continue discussion on the teddy repo.

jneem:
Thanks for taking a look!

I can certainly fix 1.

For 2, the associated consts can probably be replaced by functions returning constants and nothing will change. Removing the `asm` will make the AVX2 version a bit slower, because LLVM doesn't do a good job on the shuffles. It shouldn't be too dramatic, though.

For 3, I'd be happy to try to reduce (or better document) the complex parts, but I'm less enthusiastic about actually writing everything out. Suppose we say the "basic step" is reading some input, doing the shuffles, and checking if a fingerprint matches. In the in-tree Teddy, this is written out three times, one for each fingerprint size. If you want aligned loops, each fingerprint size needs the basic step written out twice. If you also want to unroll the loop once, each fingerprint size needs the basic step written out three times. That's 9 repetitions, multiplied by two instruction sets makes 18. There are at least two more instruction sets (AVX-512 and ARM NEON) that might be worthwhile supporting in the forseeable future, which makes 36 times that the basic step needs to be written out.

If you do decide you want to go down that road, it might be better to start from in-tree Teddy.

BurntSushi:
For (3), if the only alternative is writing it out 18 times, then that is indeed something I missed and that sounds bad. I think that means I need to spend more time with the code. Some additional comments in the code would be a big help. :-)

jneem:
Well, 18 was the most pessimistic. If you were generic over the SIMD type
but not the `State`, it would "only" be 9.

I'll try to get some better comments written this weekend.

BurntSushi:
@jneem I see. 9 is still quite bad too.

jneem:
Ok, I've made some simplifications and improved the documentation. Please
take a look!

PR #297, opened 2016-11-18
Fix typo in error messages.

mattdeboard:
While working on [rure-python](https://github.com/davidblewett/rure-python) this typo was helpful in finding why Python was telling me what it was telling me. However it has now outlived its usefulness and must die.


BurntSushi:
Looks like the build failed due to a [bug in `rustc`](https://github.com/rust-lang/rust/issues/37828), which appears to be [almost fixed](https://github.com/rust-lang/rust/pull/37797). (Alternatively, we could set `codegen-units = 1`.)

@mattdeboard In any case, thank you! :-) Glad it was useful for a time!

PR #299, opened 2016-12-02
dylib -> cdylib

steveklabnik:
The newer cdylib crate type is specifically for the "expose to other languages" case.

### DIFF
--- a/regex-capi/Cargo.toml
+++ b/regex-capi/Cargo.toml
@@ -13,7 +13,7 @@ A C API for Rust's regular expression library.
 
 [lib]
 name = "rure"
-crate-type = ["staticlib", "dylib"]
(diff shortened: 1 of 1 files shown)


BurntSushi:
Oh ya, I remember this. Didn't realize it landed. Do you happen to know which version of Cargo? (If not, I can dig for it.) I only ask because it might mean a `0.2` release (which is fine for `regex-capi`).

Drats. Is Cargo still broken?

steveklabnik:
>  Do you happen to know which version of Cargo?

rustc 1.11.

> Drats. Is Cargo still broken?

Yes. There's one more build on last night's nightly, it's been going 14 hours. If it succeeds, hooray, if not, ugh, tomorrow.

BurntSushi:
yay ci is back! Thanks!

PR #300, opened 2016-12-03
Update to thread_local 0.3

joshtriplett:
This reduces the number of versions of thread_local and thread-id present in many application crates.

This also changes the minimum Cargo version required, so it shouldn't go in until that makes sense.


BurntSushi:
The build failure on Rust 1.3 is precisely why this hasn't already been done. I'll do it as part of the `0.2` release.

(For context, I'm the one that [pushed for thread_local 0.3's existence](https://github.com/Amanieu/thread_local-rs/pull/4) so that I could do precisely this. :-))

PR #304, opened 2016-12-08
Verify character class still non-empty after converting to byte class

scooter-dangle:
For `[^\x00-\xff]`, while it is still treated as a full Unicode character class, it is not empty. For instance `≥` would still be matched.

However, when `CharClass::to_byte_class` is called on it (as is done when using `regex::bytes::Regex::new` rather than `regex::Regex::new`), it _is_ now empty, since it excludes all possible bytes.

This commit adds a test asserting that `regex::bytes::Regex::new` returns `Err` for this case (in accordance with https://github.com/rust-lang-nursery/regex/issues/106) and adds an `is_empty` check to the result of calling `CharClass::to_byte_class`, which allows the test to pass.

Fixes https://github.com/rust-lang-nursery/regex/issues/303

### DIFF
--- a/regex-syntax/src/parser.rs
+++ b/regex-syntax/src/parser.rs
@@ -596,7 +596,17 @@ impl Parser {
         Ok(Build::Expr(if self.flags.unicode {
             Expr::Class(class)
         } else {
-            Expr::ClassBytes(class.to_byte_class())
+            let byte_class = class.to_byte_class();
+
+            // If `class` was only non-empty due to multibyte characters, the 
+            // corresponding byte class will now be empty.
+            //
+            // See https://github.com/rust-lang-nursery/regex/issues/303
+            if byte_class.is_empty() {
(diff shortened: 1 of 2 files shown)


scooter-dangle:
I wasn't sure if you'd prefer moving the original `class.is_empty()` check down into the true branch of the `if self.flags.unicode` conditional to avoid making the call twice when we're parsing a byte class.

I didn't look into the `to_byte_class` code at all to see if it might rely at some point on the input not being empty.

BurntSushi:
I finally had a chance to actually dig into the problem you described and I agree with your analysis and the fix. Thank you so much!

I will add a test to `regex-syntax` for this as well.

scooter-dangle:
Thanks for looking into it and merging! (Also, thanks for being patient while I was verifying where/how we were seeing the issue.)

PR #305, opened 2016-12-09
Minor documentation fixes in teddy128.rs

ravron:
- Fix minor typos
- Fix bitmasks in Teddy explanation
- Fix min/max swap in documentation comment

I landed here from the [ripgrep blog post][1], and noticed a few errors while reading through the file. Please double-check my fixes.

[1]: http://blog.burntsushi.net/ripgrep/


BurntSushi:
Thank you for this! Sorry about the late review, but I had to make time to run through the example to make sure your edit was correct. Indeed, it is. :-) So sorry about that mistake, I know mistakes in examples like that can be really frustrating!

PR #306, opened 2016-12-15
Use separate example for RegexSet and bytes::RegexSet

### LINKED ISSUE #298: All the examples in regex::bytes are invalid uses of the API
The documentation for `regex::re_set::bytes` and `regex::re_set::unicode` are identical, because they're generated with a macro that is just hardcoded to output the documentation for unicode either way. I'm surprised this isn't caught by doctests (they don't seem to be disabled, although I'm not sure exactly how the configuration for tests in cargo works).

Also, question: why don't the both the `regex::*` and `regex::bytes::*` versions take `AsRef<[u8]>`? Then the documentation would be correct for both. You convert the input to `&[u8]` anyway. I guess it would allow passing invalid UTF8 to `regex::*` but you could at least use those bounds for `regex::bytes::*`.

bluss:
Make the doc example a macro parameter (doc comments turn into _several_
of #[doc] attrs).

Fixes #298

### DIFF
--- a/src/re_set.rs
+++ b/src/re_set.rs
@@ -9,7 +9,8 @@
 // except according to those terms.
 
 macro_rules! define_set {
-    ($name:ident, $exec_build:expr, $text_ty:ty, $as_bytes:expr) => {
+    ($name:ident, $exec_build:expr, $text_ty:ty, $as_bytes:expr,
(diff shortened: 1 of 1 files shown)


BurntSushi:
My hero! Thank you!

PR #315, opened 2017-01-02
Fix replace_all when only match is at start and replacement is empty.

### LINKED ISSUE #314: basic regex rule fails to match after updating to 0.2
I have a very simple function used to strip ansi formatting, which used to work fine with regex 0.1:

```rust
extern crate regex;

fn main() {
    let input = "\x1b[34mfoo";
    assert_eq!(strip_formatting(input), "foo");
}

fn strip_formatting<'t>(input: &'t str) -> String {
    let re = regex::Regex::new(r#"\x1B\[.+?m"#).unwrap();
    re.replace_all(input, "")
}
```

After updating to regex 0.2, I did a small change to the function to adapt to the new API (no syntax change):

```rust
extern crate regex;
use std::borrow::Cow;

fn main() {
    let input = "\x1b[34mfoo";
    assert_eq!(strip_formatting(input), "foo");
}

fn strip_formatting<'t>(input: &'t str) -> Cow<'t, str> {
    let re = regex::Regex::new(r#"\x1B\[.+?m"#).unwrap();
    re.replace_all(input, "")
}
```

... but now it fails to match the regex rule:

```
   Compiling generic v0.0.1 (file:///Users/mX/projects/rust/tmp/generic)
    Finished debug [unoptimized + debuginfo] target(s) in 2.9 secs
     Running `target/debug/generic`
thread 'main' panicked at 'assertion failed: `(left == right)` (left: `"\u{1b}[34mfoo"`, right: `"foo"`)', src/main.rs:6
```

The strange thing is that it works if I replace with something instead of `""`, e.g. this works:

```rust
extern crate regex;
use std::borrow::Cow;

fn main() {
    let input = "\x1b[34mfoo";
    assert_eq!(strip_formatting(input), "something foo");
}

fn strip_formatting<'t>(input: &'t str) -> Cow<'t, str> {
    let re = regex::Regex::new(r#"\x1B\[.+?m"#).unwrap();
    re.replace_all(input, "something ")
}
```

utkarshkukreti:
Fixes #314.

All old tests and the new test I added pass, and I also confirmed manually that the test case reported in #314 passes with this change. Let me know if you can think of any case that my change doesn't handle!

### DIFF
--- a/src/re_bytes.rs
+++ b/src/re_bytes.rs
@@ -493,7 +493,7 @@ impl Regex {
                 new.extend_from_slice(&rep);
                 last_match = m.end();
             }
-            if new.is_empty() {
+            if last_match == 0 {
                 return Cow::Borrowed(text);
             }
             new.extend_from_slice(&text[last_match..]);
--- a/src/re_unicode.rs
+++ b/src/re_unicode.rs
@@ -569,7 +569,7 @@ impl Regex {
                 new.push_str(&rep);
                 last_match = m.end();
             }
-            if new.is_empty() {
+            if last_match == 0 {
                 return Cow::Borrowed(text);
             }
             new.push_str(&text[last_match..]);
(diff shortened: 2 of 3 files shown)


BurntSushi:
Thanks for fixing this! This was a pretty silly mistake on my part!

maximih:
will a new version including this fix be pushed automatically to crates.io or someone has to trigger that manually?

thanks

BurntSushi:
@maximih I'm working on it. I'll issue a release once #323 is merged.

BurntSushi:
@maximih 0.2.1 is out!

maximih:
@BurntSushi thanks!

PR #335, opened 2017-01-29
Update docs to reflect `at` -> `get` rename

reddraggone9:
I noticed this while perusing the documentation and ran
```sh
rg '`at`'
```
in the repo to find these four instances. Feels kinda meta.


BurntSushi:
Nice, thank you! :-)

PR #339, opened 2017-02-16
docs: Fix unescaped `[` in character class

robinst:
Compiling the regex as it was documented before results in:

> Syntax("Error parsing regex near \'-/:-@[-`{-\' at character offset 7: Use of unescaped \'[\' in character class is not allowed.")


BurntSushi:
Another nice catch, thanks!

PR #346, opened 2017-02-22
Support nested character classes and intersection with `&&`

### LINKED ISSUE #341: Implement (at least part of) UTS#18 RL1.3 - Operators in character sets
I'm working on a syntax highlighting engine in Rust that requires an Oniguruma-compatible regex engine. I'm trying to port it from the `onig` crate to [fancy-regex](https://github.com/google/fancy-regex), but there's some features it doesn't support yet (see https://github.com/trishume/syntect/pull/34).

One of these features is the `&&` operator and nesting in character sets, for example `[a-w&&[^c-g]z]`. I was thinking this would be added to fancy-regex but @robinst pointed out [this comment](https://github.com/rust-lang/regex/blob/52fdae7169ec619530985a019184319ac4bbee5a/regex-syntax/src/lib.rs#L1408-L1410) which suggests that you plan for them to be in the `regex` crate.

It would be nice if the `regex` crate supported [UTS#18 RL1.3](http://www.unicode.org/reports/tr18/tr18-19.html#Subtraction_and_Intersection) in full, but the `&&` operator and nesting are all that Oniguruma-compatibility of fancy-regex requires.

I imagine this would take some changes to `regex-syntax` and then a pass to convert the fancy character sets down to basic character sets. I haven't thought enough about it to know if there are any unicode-related issues that might make this more complex, perhaps by making a tiny fancy character set compile to an enormous basic character set.

@BurntSushi do you have any insight on how difficult you think this would be to add for a contributor not familiar with the internals of `regex`?

robinst:
This implements parts of UTS#18 RL1.3, namely:

* Nested character classes, e.g.: `[a[b-c]]`
* Intersections in classes, e.g.: `[\w&&\p{Greek}]`

They can be combined to do things like `[\w&&[^a]]` to get all word
characters except `a`.

Fixes #341

### DIFF
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -680,7 +680,7 @@ impl CharClass {
         self.canonicalize()
     }
 
-    /// Canonicalze any sequence of ranges.
+    /// Canonicalize any sequence of ranges.
     ///
     /// This is responsible for enforcing the canonical format invariants
     /// as described on the docs for the `CharClass` type.
@@ -701,6 +701,41 @@ impl CharClass {
             ordered.ranges.push(candidate);
         }
         ordered
+    }
+
+    /// Calculate the intersection of two canonical character classes.
+    ///
+    /// The returned intersection is canonical.
+    fn intersection(&self, other: &CharClass) -> CharClass {
+        if self.ranges.is_empty() || other.ranges.is_empty() {
+            return CharClass::empty();
+        }
+
+        let mut intersection = CharClass::empty();
+
+        let mut iter_a = self.ranges.iter();
+        let mut iter_b = other.ranges.iter();
+        let mut a = iter_a.next().unwrap();
+        let mut b = iter_b.next().unwrap();
+        loop {
+            if let Some(i) = a.intersection(&b) {
+                intersection.ranges.push(i);
+            }
+
+            // If the range with the smaller end didn't match this time,
+            // it won't ever match, so move on to the next one.
+            let (iter, item) = if a.end < b.end {
+                (&mut iter_a, &mut a)
+            } else {
+                (&mut iter_b, &mut b)
+            };
+            match iter.next() {
+                Some(v) => *item = v,
+                None => break, // no more ranges to check, done
+            }
+        }
+
+        intersection.canonicalize()
     }
 
     /// Negates the character class.
@@ -799,6 +834,18 @@ impl ClassRange {
     /// since ranges are inclusive, `a-c` and `d-f` are overlapping!
     fn overlapping(self, other: ClassRange) -> bool {
         max(self.start, other.start) <= inc_char(min(self.end, other.end))
+    }
+
+    /// Returns the intersection of the two ranges if they have common
+    /// characters, `None` otherwise.
+    fn intersection(&self, other: &ClassRange) -> Option<ClassRange> {
+        let start = max(self.start, other.start);
+        let end = min(self.end, other.end);
+        if start <= end {
+            Some(ClassRange::new(start, end))
+        } else {
+            None
+        }
     }
 
     /// Creates a new range representing the union of `self` and `other.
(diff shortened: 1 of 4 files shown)


regex-syntax/src/lib.rs
```diff
@@ -703,6 +703,43 @@ impl CharClass {
         ordered
     }
 
+    /// Calculate the intersection of two canonical character classes.
+    ///
+    /// The returned intersection is canonical.
+    fn intersection(&self, other: &CharClass) -> CharClass {
```
robinst: Not sure if this should be `pub` or not.

regex-syntax/src/lib.rs
```diff
@@ -703,6 +703,43 @@ impl CharClass {
...
+                (&mut iter_a, &mut a)
+            } else {
+                (&mut iter_b, &mut b)
+            };
+            if let Some(v) = iter.next() {
+                *item = v;
+            } else {
+                // No more ranges to check, done.
+                break;
+            }
```
robinst: Not sure if this is the most idiomatic way to do this :). Maybe using old-fashioned indexing would work better in this case.

regex-syntax/src/parser.rs
```diff
@@ -526,65 +526,132 @@ impl Parser {
...
+            // corresponding byte class will now be empty.
+            //
+            // See https://github.com/rust-lang/regex/issues/303
+            if byte_class.is_empty() {
+                // e.g., (?-u)[^\x00-\xFF]
+                return Err(self.err(ErrorKind::EmptyClass));
+            }
+
+            Expr::ClassBytes(byte_class)
+        }))
```
robinst: This is the same as before, just moved.

regex-syntax/src/parser.rs
```diff
@@ -2222,6 +2312,250 @@ mod tests {
...
+        bytes.remove(b'c');
+        bytes.remove(b'C');
+        assert_eq!(pb(r"(?i-u)[a-b[^c]]"), Expr::ClassBytes(bytes));
+    }
+
+    #[test]
+    fn class_nested_class_brackets_hyphen() {
+        // This is really confusing, but `]` is allowed if first character within a class
+        // It parses as a nested class with the `]` and `-` characters
+        assert_eq!(p(r"[[]-]]"), Expr::Class(class(&[('-', '-'), (']', ']')])));
```
robinst: Is the decision to allow `]` unescaped as the first character in a class final? Thinking about this and how it interacts with `&&` was one of the most confusing parts about this change.

I'd wish the rule for `]` was that it needs to be escaped inside a character class, no matter where it is. The same for `-`. Is it too late to change this? It would simplify the parsing a bit, and would make it easier to explain how things work.

regex-syntax/src/parser.rs
```diff
@@ -526,65 +526,132 @@ impl Parser {
...
+            match self.cur() {
+                ']' => {
+                    // end of the class
+                    self.bump();
+                    break;
+                }
+                '&' => {
+                    // intersection with `&&`
+                    self.bump();
+                    self.bump();
```
trishume: Shouldn't this throw a syntax error (or just treat it as a literal `&` in the character class, not sure what spec says) if the second character you `bump` over isn't also an `&`?

**Edit:** noticed later that you only exit `parse_class_set` if you look ahead and see `&&`. Maybe add a comment on the second bump saying this was verified to be `&` in parse_class_set.

regex-syntax/src/parser.rs
```diff
@@ -596,27 +663,50 @@ impl Parser {
...
+    // End:   `x`
+    fn parse_class_escape(&mut self, class: &mut CharClass) -> Result<()> {
+        match try!(self.parse_escape()) {
+            Build::Expr(Expr::Class(class2)) => {
+                class.ranges.extend(class2);
             }
-
-            Expr::ClassBytes(byte_class)
-        }))
+            Build::Expr(Expr::ClassBytes(class2)) => {
```
trishume: What are all these extra cases for? I don't see how they relate to the rest of the changes in this PR. Do they add support for more types of escapes in character classes? Which tests test these new branches?

regex-syntax/src/parser.rs
trishume: Maybe if it is too late to break compatibility with existing regexes we can at least ban `]` and possibly `-` as first characters of a nested character class. As that test demonstrates, it can get really confusing when nested character classes are involved.

robinst: Maybe. The downside of this is that it would break the useful rule of "everything that can be a top-level character class can also be a nested character class".

regex-syntax/src/parser.rs
robinst: Yeah, needs a comment, adding it.

regex-syntax/src/parser.rs
robinst: It's hard to see in the diff, but this is unchanged from before, just extracted into its own `parse_class_escape` method.

regex-syntax/src/parser.rs
```diff
@@ -526,65 +526,132 @@ impl Parser {
...
                     Some(class2) => class.ranges.extend(class2),
                     None => {
-                        return Err(self.err(
-                            ErrorKind::UnsupportedClassChar('[')));
-                    }
-                },
-                '\\' => match try!(self.parse_escape()) {
-                    Build::Expr(Expr::Class(class2)) => {
+                        // Nested set, e.g. `[c-d]` in `[a-b[c-d]]`
+                        let class2 = try!(self.parse_class_as_chars());
```
BurntSushi: I haven't had chance to do a thorough review yet (although, from what I see, this looks really really good), but this is a potential problem. In particular, this turns a parser with predictable stack growth into a parser with unpredictable stack growth. This is bad because if a program accepts regexes as user input and they do `[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[...` or some such, then this recursion will cause a stack overflow.

With that said, there are various parts of this crate that use recursion over the abstract syntax, but not before [checking the nesting limit](https://doc.rust-lang.org/regex/regex_syntax/struct.ExprBuilder.html#method.nest_limit). It's not quite a guarantee (since who knows how big the stack really is), but it does afford the caller some control over what happens if a very bad regex is provided.

This basically means you would need to maintain an explicit stack in the parser state of nested character classes, and this is what the `Build` enum is for. Right now, the `Build` enum only cares about groups and alternates, but maybe it's easy to add a variant for nested classes?

I'm sorry I didn't bring this point up earlier. I should have seen it coming, but I completely forgot about it until the code was in front of me.

How willing are you to make this change? If not, I might be able to finish it (I just don't know when I will).

regex-syntax/src/parser.rs
BurntSushi: I admit that this is exasperated by the presence of nested character classes, but AFAIK, this is pretty standard. Changing this would, no doubt in my mind, require a semver version bump. I really really don't like breaking changes in the regex syntax, so I would rather live with the implementation complexity over the breaking changes and inconsistencies with other regex engines.

regex-syntax/src/lib.rs
BurntSushi: I'd say hold off for now. Thanks.

regex-syntax/src/parser.rs
robinst: Ok. Just tested with Java's Pattern, even its implementation allows it. So yeah, makes sense to keep it. (It's just one of those things that seem like they were originally done "not because it's a good idea, but because we can", and now everyone has to support it.)

BurntSushi: > (It's just one of those things that seem like they were originally done "not because it's a good idea, but because we can", and now everyone has to support it.)

I have no doubt that that is indeed the case. :-)

In fact, it's probably true for a lot of things in the regex syntax. :-)

regex-syntax/src/parser.rs
robinst: Very good point. I remember wondering about that problem while implementing it, but then forgot later.

I'll look into using `Build`, yeah. Maybe it could also be a separate stack. Have to wrap my head around it first though :).

regex-syntax/src/lib.rs
ncm: Maybe
`    match iter.next() {
           Some(v) => *item = v,
           _ => break  // no more ranges to check, done
     }`
?

robinst:
Rewrote it to use a stack instead of recursion. Also added a test for deeply nested character classes that didn't pass before but now it does :).

I decided to not use `Build` but instead have a separate stack as the two things are completely separate.

trishume:
@robinst Awesome. I looked over it briefly but I'm too tired to fully understand it at the moment, so I haven't done anything close to a careful review.

One thing I'm unsure about is if the important thing is just to not stack overflow, or to have limited memory usage. At the moment this should use heap proportional to at most the length of the regex as far as I can tell. But @BurntSushi said something about the [nesting limit](https://doc.rust-lang.org/regex/regex_syntax/struct.ExprBuilder.html#method.nest_limit), and I'm not sure if checking it is warranted in this case, or only when using recursion.

BurntSushi:
> But @BurntSushi said something about the nesting limit, and I'm not sure if checking it is warranted in this case, or only when using recursion.

Hmm, yes, the primary purpose of the nesting limit is to mitigate the risk of stack overflow. However, in the nested character case, all nestings are resolved (I think) to a single flattened set of Unicode codepoints.

Heap space may indeed be unbounded, but:

1. This is already true today (I think) with regexes like `(((((((((((((((((((((((((((((...`
2. The heap space is roughly proportional to the actual size of the regular expression pattern string, so some simple mitigation measures are possible by users.

robinst:
> all nestings are resolved (I think) to a single flattened set of Unicode codepoints.

Yes, they are.

> This is already true today (I think) with regexes like `(((((((((((((((((((((((((((((...`

Yes, the `stack` just keeps growing and there's no check before pushing expressions onto it.

Did you have a chance to look at the change yet? I was looking into a `quickcheck` test for `intersection` as well, I can add it if you want.

robinst:
> I was looking into a quickcheck test for intersection as well, I can add it if you want.

Added the quickcheck test as well now.

BurntSushi:
OK, I've finally had a chance to review this in more depth and I'm pretty much speechless. This is phenomenal work. I looked hard, but I don't see anything to complain about!

I'm not sure if #354 will cause conflicts, so you might wind up needing to rebase, but let's see what happens. (Bors seems to be stuck at the moment...)

@bors r+

robinst:
> OK, I've finally had a chance to review this in more depth and I'm pretty much speechless. This is phenomenal work. I looked hard, but I don't see anything to complain about!

Wow, thank you :)! It was surprisingly simple to do this change, the codebase is very readable!

Rebased it now!

PR #349, opened 2017-03-20
Fix panics parsing regex with whitespace in extended mode

robinst:
The added tests fail without the fix like this:

    ---- parser::tests::ignore_space_escape_hex2 stdout ----
    	thread 'parser::tests::ignore_space_escape_hex2' panicked at 'called `Result::unwrap()` on an `Err` value: Error { pos: 10, surround: "x 5 3", kind: InvalidBase16(" 5 3") }', src/libcore/result.rs:860

    ---- parser::tests::ignore_space_escape_hex stdout ----
    	thread 'parser::tests::ignore_space_escape_hex' panicked at 'called `Result::unwrap()` on an `Err` value: Error { pos: 12, surround: "{ 5 3 }", kind: InvalidBase16(" 5 3") }', src/libcore/result.rs:860

    ---- parser::tests::ignore_space_ascii_classes stdout ----
    	thread 'parser::tests::ignore_space_ascii_classes' panicked at 'called `Result::unwrap()` on an `Err` value: Error { pos: 5, surround: "(?x)[ [ : ", kind: UnsupportedClassChar('[') }', src/libcore/result.rs:860

    ---- parser::tests::ignore_space_escape_octal stdout ----
    	thread 'parser::tests::ignore_space_escape_octal' panicked at 'valid octal number', src/libcore/option.rs:785

    ---- parser::tests::ignore_space_escape_unicode_name stdout ----
    	thread 'parser::tests::ignore_space_escape_unicode_name' panicked at 'called `Result::unwrap()` on an `Err` value: Error { pos: 15, surround: "Y i }", kind: UnrecognizedUnicodeClass(" Y i") }', src/libcore/result.rs:860

    ---- parser::tests::ignore_space_repeat_counted stdout ----
    	thread 'parser::tests::ignore_space_repeat_counted' panicked at 'called `Result::unwrap()` on an `Err` value: Error { pos: 15, surround: ", 1 0 }", kind: InvalidBase10("1 0") }', src/libcore/result.rs:860

The reason for the panics is that `bump_get` would ignore space when
walking the characters, but then keep the spaces in the returned String.

Found using cargo-fuzz.

### DIFF
--- a/regex-syntax/src/parser.rs
+++ b/regex-syntax/src/parser.rs
@@ -423,7 +423,7 @@ impl Parser {
     //
     // Start: `1`
     // End:   `,` (where `until == ','`)
-    fn parse_decimal<B: Bumpable>(&mut self, until: B) -> Result<u32> {
+    fn parse_decimal<F: FnMut(char) -> bool>(&mut self, until: F) -> Result<u32> {
         match self.bump_get(until) {
             // e.g., a{}
             None => Err(self.err(ErrorKind::MissingBase10)),
@@ -809,14 +809,23 @@ impl Parser {
 
     fn eof(&self) -> bool { self.chars().next().is_none() }
 
-    fn bump_get<B: Bumpable>(&mut self, s: B) -> Option<String> {
-        let n = s.match_end(self);
+    fn bump_get<F: FnMut(char) -> bool>(&mut self, mut f: F) -> Option<String> {
+        let mut s = String::new();
+        let n = {
+            let bumpable = |c| {
+                if f(c) {
+                    s.push(c);
+                    true
+                } else {
+                    false
+                }
+            };
+            bumpable.match_end(self)
+        };
         if n == 0 {
             None
         } else {
             let end = checkadd(self.chari, n);
-            let s = self.chars[self.chari..end]
-                        .iter().cloned().collect::<String>();
             self.chari = end;
             Some(s)
         }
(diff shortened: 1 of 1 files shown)


regex-syntax/src/parser.rs
```diff
@@ -2375,6 +2384,46 @@ mod tests {
     }
 
     #[test]
+    fn ignore_space_escape_unicode_name() {
+        assert_eq!(p(r"(?x)\ p { Y i }"), Expr::Class(class(YI)));
+    }
+
+    #[test]
+    fn ignore_space_escape_octal() {
+        assert_eq!(p(r"(?x)\ 1 2 3"), lit('S'));
```
robinst: Seems a bit weird that it's allowed to add space between digits of a number, but that seems to be the closest to the current behavior.

BurntSushi:
@robinst Thanks for finding this! Sorry it slipped out of my queue, but your blog post caught my attention. :-) Nice work!

I'm not sure the fix is right either. Does this also apply to thinks like `\p{G r e     ek}`? We should probably look at how other regex engines handle verbose mode.

robinst:
> I'm not sure the fix is right either. Does this also apply to thinks like `\p{G r e ek}`?

Yes, and things like `[ [ : u p p e r : ] ]`, see the test cases. I'm not sure about it either.

Maybe whitespace should only be allowed between logical groups of characters. For example, it should not be allowed within a number or within a text identifier. Here's what other engines do:

Oniguruma: `(?x) \ x53`, `(?x) \x 53`, `(?x) \x5 3` all compile but don't match `S`

Perl behaves the same way, checked with `perl -e 'print "matches" if "S" =~ /(?x) \ x53/'`

So at least for `\x` they don't seem to allow any space in the whole sequence. Even `\ d` doesn't match a digit, whereas `regex-syntax` does.

BurntSushi:
Thinking about this a bit more, it feels like we shouldn't allow arbitrary whitespace in arbitrary syntax. Maybe things like `\p{G r e e k}` should produce a syntax error?

robinst:
> Thinking about this a bit more, it feels like we shouldn't allow arbitrary whitespace in arbitrary syntax.

Agreed. I've prepared a different pull request here: https://github.com/rust-lang/regex/pull/354

BurntSushi:
I decided to go with #354 over this one. Thanks so much!

PR #354, opened 2017-04-07
Fix panics with whitespace in extended mode by being more strict

robinst:
Instead of ignoring space in all the bump/peek methods (as proposed in
pull request #349), have an explicit `ignore_space` method that can be
used in places where space/comments should be allowed.

This makes parsing a bit stricter than before as well.

### DIFF
--- a/regex-syntax/src/parser.rs
+++ b/regex-syntax/src/parser.rs
@@ -104,7 +104,11 @@ impl Parser {
     // Starts at the beginning of the input and consumes until either the end
     // of input or an error.
     fn parse_expr(mut self) -> Result<Expr> {
-        while !self.eof() {
+        loop {
+            self.ignore_space();
+            if self.eof() {
+                break;
+            }
             let build_expr = match self.cur() {
                 '\\' => try!(self.parse_escape()),
                 '|' => { let e = try!(self.alternate()); self.bump(); e }
@@ -177,7 +181,7 @@ impl Parser {
             return Err(self.err(ErrorKind::UnexpectedEscapeEof));
         }
         let c = self.cur();
-        if is_punct(c) {
+        if is_punct(c) || (self.flags.ignore_space && c.is_whitespace()) {
             let c = self.bump();
             return Ok(try!(self.lit(c)));
         }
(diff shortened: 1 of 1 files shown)


regex-syntax/src/parser.rs
```diff
@@ -2424,6 +2489,14 @@ mod tests {
         ]));
     }
 
+    #[test]
+    fn ignore_space_escape_space() {
+        assert_eq!(p(r"(?x)a\  # hi there"), Expr::Concat(vec![
```
robinst: Allowing to escape whitespace is in line with what PCRE2 does, see here: http://www.pcre.org/current/doc/html/pcre2pattern.html#SEC5

> An escaping backslash can be used to include a white space or # character as part of the pattern.

And it seems less surprising than allowing `\ d` or the following to mean digits:

```rust
r"(?x)\ # comment
    d"
```

BurntSushi:
@robinst Thanks so much for working on this! This behavior is indeed a bit more in line with what I'd expect. I will say though that doing it this way leads to a lot more implementation complexity. :-( I'm not sure I can immediately think of a better way though.

> Allowing to escape whitespace is in line with what PCRE2 does

Hmmm, yes, I'll need to think about this. I'm a bit distracted at the moment.

robinst:
> I will say though that doing it this way leads to a lot more implementation complexity. :-( I'm not sure I can immediately think of a better way though.

FWIW, fancy-regex's parser does it the same way as I discovered recently, see the calls to `optional_whitespace`: https://github.com/google/fancy-regex/blob/master/src/parse.rs#L103

BurntSushi:
@robinst Sorry for taking so long to get around this. For the most part, I've avoid this because I've been internally waffling over this, but that's silly to do. Some thoughts:

1. I'm pretty unhappy with this solution, because it makes the parser a lot more complex. My guess is that new bugs will be introduced.
2. This does however fix a real bug.
3. I am in the process of rewriting `regex-syntax`. When I get to supporting `x`, I'll sit down and re-evaluate our approach here.

Thanks so much for finding and fixing this bug. It is much appreciated!

PR #365, opened 2017-05-14
Fix Bug in `rure_captures_len`

iwillspeak:
It looks like at some point in the past the captures were refactored
from being a vector of start and end positions into a list of location
structures. The C API still had a conversion of the lenght which
corrected for the captures being twice the length of the number of
captures.

This updates the length calculation in `rure.rs` to return the
correct length, and adds an assertion to the test case.

### DIFF
--- a/regex-capi/ctest/test.c
+++ b/regex-capi/ctest/test.c
@@ -102,6 +102,19 @@ bool test_captures() {
         }
         passed = false;
     }
+    size_t expect_captures_len = 3;
+    size_t captures_len = rure_captures_len(caps);
+    if (captures_len != expect_captures_len) {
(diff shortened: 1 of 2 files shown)


BurntSushi:
How embarrassing! Thank you for the fox and the test.

I'm on mobile so it's hard to tell, but it looks like the C code is formatted differently from other C code in the same file. Could you make the formatting consistent? (And please also squash. :-)) Thanks!

iwillspeak:
Tabs vs spaces I guess. Will fix up.

BurntSushi:
Awesome, thank you! @bors r+

iwillspeak:
\o/

Issue #357, opened 2017-04-21
Please, impl Into<&'t str> for Match<'t>


jan-hudec:
The `Match::as_str` takes `&self`, so it can't be directly mapped onto the `Option<Match<'t>>` returned by `get` and `name`. A version consuming the Match would be more convenient for the purpose. And standard conversion `From`/`Into` seem to fit here.

Basically, code that used `captures.name("foo")` now needs to use `captures.name("foo").as_ref().map(Match::as_str)` or `captures.name("foo").map(|m| m.as_str())`. With Into, it should be able to use `captures.name("foo").map(From::from)`.

Some code can be changed to use Match directly, but anything that wants to continue with something like `.unwrap_or("")` needs to convert to string.

Also note that `Index` does not seem to fit the bill, because code that needs Option can't use that.

BurntSushi:
I'm not quite sure what to make of this. I feel like writing `|m| m.as_str()` is pretty reasonable. On the other hand, I don't think there is much cost to this, so I think I'd be willing to add it.

BurntSushi:
Done inhttps://github.com/rust-lang/regex/pull/374

PR #373, opened 2017-05-22
Implement Into::into for Match

lukaslueg:
Concerning #357


killercup:
You could implement `From` instead and get the `Into` implementation for free.

BurntSushi:
Right, I agree with @killercup, we should prefer `From` impls.

PR #376, opened 2017-05-31
Fix clippy warnings

lukaslueg:
This fixes all clippy warnings, mostly lifetime elisions and useless borrows. Clippy now passes on nightly; on stable/without clippy, no lint warnings are given. All tests still pass.

I've not moved any code around, which is the reason for allowing `string_lit_as_bytes` in so many places.

Up for review


src/re_bytes.rs
```diff
@@ -785,6 +785,7 @@ pub struct Captures<'t> {
     named_groups: Arc<HashMap<String, usize>>,
 }
 
+#[allow(len_without_is_empty)]
```
lukaslueg: It's probably better to

```rust
fn is_empty(&self) -> bool {
    debug_assert!(self.len() > 0);
    false
}
```

as this will trigger const-folding and warnings should anyone expect captures to be empty

src/prog.rs
```diff
@@ -158,6 +158,12 @@ impl Program {
     }
 }
 
+impl Default for Program {
```
BurntSushi: Is this actually used? If not, please remove it.

src/re_bytes.rs
BurntSushi: I don't agree with adding a useless method to the public API.

src/re_bytes.rs
```diff
@@ -986,7 +987,7 @@ pub trait Replacer {
     /// be beneficial to avoid finding sub-captures.
     ///
     /// In general, this is called once for every call to `replacen`.
-    fn no_expansion<'r>(&'r mut self) -> Option<Cow<'r, [u8]>> {
+    fn no_expansion(&mut self) -> Option<Cow<[u8]>> {
```
BurntSushi: Please put the lifetimes back. I don't mind elision in simple cases, but I'd rather the lifetime parameter be explicit here, *especially* in the trait definition.

src/re_trait.rs
```diff
@@ -162,10 +163,10 @@ pub trait RegularExpression: Sized {
 
     /// Returns an iterator over all non-overlapping successive leftmost-first
     /// matches with captures.
-    fn captures_iter<'t>(
+    fn captures_iter(
         self,
-        text: &'t Self::Text,
-    ) -> CaptureMatches<'t, Self> {
+        text: & Self::Text,
```
BurntSushi: There's an extra space here.

src/re_unicode.rs
```diff
@@ -1198,7 +1199,7 @@ pub trait Replacer {
     /// be beneficial to avoid finding sub-captures.
     ///
     /// In general, this is called once for every call to `replacen`.
-    fn no_expansion<'r>(&'r mut self) -> Option<Cow<'r, str>> {
+    fn no_expansion(&mut self) -> Option<Cow<str>> {
```
BurntSushi: As with above, please put the lifetime back. Leaving it off the implementations is fine.

tests/api.rs
```diff
@@ -1,36 +1,48 @@
 #[test]
+#[cfg_attr(not(feature = "cargo-clippy"), allow(unknown_lints))]
+#[allow(trivial_regex, string_lit_as_bytes)]
```
BurntSushi: No, I'm sorry, I cannot abide peppering attributes everywhere for Clippy. These have to go. If that means we don't get clean Clippy output, then we'll just have to live with that.

src/compile.rs
```diff
@@ -1049,13 +1055,6 @@ impl ByteClassSet {
     }
 }
 
-fn u32_to_usize(n: u32) -> usize {
```
lukaslueg: Actually, I think this is the only really useful warning by clippy for the whole codebase: The comparison always evaluates to false, making the check superfluous.

Why is it there anyway? if `usize` is 32bit, `n` can't be larger. If `usize` is 64bit, `n` can't be larger as well.

BurntSushi: It's a guard against the case when `usize` is less than 32 bits. I'm not sure what the state of 16 bit usizes is in rustc, but IIRC people were interested in making it happen. It's certainly possible, so it seems worthwhile to be conservative.

lukaslueg: Shall I leave a comment about `std::convert::TryFrom` ?

BurntSushi: Sure.

lukaslueg:
Don't hesitate to let me know of needed changes, it's just a few hunk-jumps :-)

src/lib.rs
```diff
@@ -432,9 +432,9 @@ These classes are based on the definitions provided in
 <pre class="rust">
 \d     digit (\p{Nd})
 \D     not digit
-\s     whitespace (\p{White_Space})
+\s     whitespace (\p{`White_Space`})
```
BurntSushi: Oh hmm, I don't think you should be using back ticks here.

src/lib.rs
```diff
@@ -432,9 +432,9 @@ These classes are based on the definitions provided in
...
 \S     not whitespace
-\w     word character (\p{Alphabetic} + \p{M} + \d + \p{Pc} + \p{Join_Control})
+\w     word character (\p{Alphabetic} + \p{M} + \d + \p{Pc} + \p{`Join_Control`})
```
BurntSushi: Or here. :-)

Issue #364, opened 2017-05-12
implement PartialEq, Eq and Hash


ashfordneil:
Without these traits being implemented, it is difficult to store Regex structs, or types containing them, in collections such as HashSet and HashMap. It also prevents any type containing a Regex from automatically deriving these traits. Would it be okay to implement these traits for the Regex struct, perhaps just using the `as_str()` function and performing comparisons / hashes on the output of that?

BurntSushi:
`regex 0.1` had  this impl, but I removed it. It's way too misleading because finite state machines actually have a concept of equality that is very different from "do these regex strings match."

You should solve this by putting your `Regex` in a newtype and deriving the traits you want.

BurntSushi:
See https://github.com/rust-lang/regex/issues/178 and https://github.com/rust-lang/regex/issues/313

DenisKolodin:
Information for those who got here. A naive comparison can be implemented this way:

```rust
struct Wrapper {
    regex: Regex,
}

impl PartialEq for Wrapper {
    fn eq(&self, other: &Self) -> bool {
        self.regex.as_str() == other.regex.as_str()
    }
}
```

casey:
Not to necro an old issue, but I wonder if not having these implementations is worth it.

If I understand correctly, the benefit of not having these implementations is that if a user expects the `==` for `Regex` to compare the underlying abstract state machines of the regex, then they will be surprised when it doesn't.

(And it can't be implemented that way, because it's expensive to compute, and `==` is generally expected to be fast.)

However, I'm not sure if that's what users actually expect. Personally, I would expect `==` to just do string comparison, and would be surprised if `\d+ == \d\d*` was `true`.

Even if it was fast, comparing the underlying of the regex behavior might also be undesirable, since could you not have two regexes which compare true, since they match the same set of strings, but which have large efficiency differences? I.e., one is much more efficient than the other, and terminates much faster. In such a case, I'm not sure you would want `==` to return true, and obscure the differences between them.

The downside of not having these implementations, or at least the downside that I've personally run into, is that if you include a `Regex` in another type, and then want to derive `ParitalEq`, `Eq`, or `Hash`, you can't, because `Regex` doesn't implement them.

I ran into this while implementing a config file that contains a regex:

```
#[derive(Deserialize)]
struct Config {
  some_other_value: u32,
  #[serde_as(as = "DisplayFromStr")]
  regex: Regex,
}
```

I wanted to do this in a test:

```rust
assert_eq! {
[...]

Issue #313, opened 2017-01-01
PartialEq is not implemented in version 0.2


tredoe:
With the version 0.2, I get:
```
error[E0369]: binary operation `==` cannot be applied to type `std::option::Option<regex::Regex>`

note: an implementation of `std::cmp::PartialEq` might be missing for `std::option::Option<regex::Regex>`
```

BurntSushi:
Indeed it is not. See the changelog please. :-)

tredoe:
Please, could you add an example about how do it? For beginners
Thanks!

BurntSushi:
Oh absolutely! I'm on mobile right now, so it's a a little tough to write it the code, but I'll try to get the idea across and then clean it up later. :-)

The low tech solution is to just change `re1 == re2` to `re1.as_str() == re2.as_str()`. That might be all you need. But if you're, say, using regexes as keys in a btree map, then you'll need a wrapper type that implements `PartialEq` for you.

First, you need to define a wrapper type: `struct MyRegex(regex::Regex)`. For convenience, you'll want to implement the `std::ops::Deref` trait. This will make it so you can call all of the methods on Regex through your MyRegex type. Second, you'll need to implement the `std::cmp::PartialEq` on your MyRegex type, which can be done using the the as_str methods from the "low tech" solution above.

PR #378, opened 2017-06-01
Adds PartialEq impl for Regex.

remexre:
This simply compares the two regexes' strings, as the same string should always produce the same regex.


BurntSushi:
Thanks, but this was explicitly dropped in regex 0.2 because regex equivalence is not determined by their concrete syntax. See #178, #313 and #364

PR #381, opened 2017-06-05
Fix docs for renaming of `RegexBuilder::build`

michaelsproul:
There were some references to `compile` left over after the rename from `compile` to `build`.


BurntSushi:
Awesome, thanks! @bors r+

Issue #394, opened 2017-08-20
replace has problems with empty strings


fulara:
Seems like a replace has problems processing empty strings. example:
```
extern crate regex;
use regex::Regex;
fn main() {
    let regex = Regex::new("\\d*").unwrap();
    let s = regex.replace("", "replaced").into_owned();
    assert_eq!("replaced", s);
}
```
That fails. however:
```
extern crate regex;
use regex::Regex;
fn main() {
    let regex = Regex::new("\\d*").unwrap();
    let s = regex.find("");
    assert!(s.is_some());
}
```
That one passes.
Thanks to lapinot I've workaround here:
```
extern crate regex;
use regex::Regex;
fn main() {
    let regex = Regex::new("\\d*").unwrap();
    let s = regex.replace("", |caps: &regex::Captures| { "replaced".to_string() }).into_owned();
    assert_eq!("replaced", s);
}
```

BurntSushi:
This is probably a dupe of #393, but we can leave this open until a fix confirms.

BurntSushi:
Fixed by #395

PR #395, opened 2017-08-20
Fix `Regex::replacen` when replacing an empty match.

### LINKED ISSUE #393: replace does not work against regex '^'.
I was just trying different corner cases with rust regex and I came across this one:
Code:
```
extern crate regex;
use regex::Regex;
fn main() {
    let regex = Regex::new("^").unwrap();
    let s = regex.replace("_", "begin").into_owned();
    assert_eq!("begin_", s); // will fail, string still '_'
}
```
Above fails. and below succeeds:
```
extern crate regex;
use regex::Regex;
fn main() {
    let regex = Regex::new("$").unwrap();
    let s = regex.replace("_", "begin").into_owned();
    assert_eq!("_begin", s); // will pass.
}
```
Tested against 0.2.2.

I tried testing the same case with boost regex and sed and both of these yielded expected result.

Lapin0t:
Fixes #393, #394.

There was some logic error when deciding to return `Cow::Borrowed` vs `Cow::Owned` in the fast-path. I took the same solution as the slow-path, ie use `peekable` on the iterator. I'm not sure this is the most efficient way (maybe just add `mut did_replace = false` and set it to true in the loop), but it does the job.

### DIFF
--- a/src/re_bytes.rs
+++ b/src/re_bytes.rs
@@ -494,18 +494,19 @@ impl Regex {
         mut rep: R,
     ) -> Cow<'t, [u8]> {
         if let Some(rep) = rep.no_expansion() {
+            let mut it = self.find_iter(text).enumerate().peekable();
+            if it.peek().is_none() {
+                return Cow::Borrowed(text);
+            }
             let mut new = Vec::with_capacity(text.len());
             let mut last_match = 0;
-            for (i, m) in self.find_iter(text).enumerate() {
+            for (i, m) in it {
                 if limit > 0 && i >= limit {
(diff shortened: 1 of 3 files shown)


Lapin0t:
Note that i don't know why the CI checks are failing, something seemingly unrelated to this part of the code.

Lapin0t:
Bump 🐅... Not sure if I should do anything or not right now!

BurntSushi:
Right, sorry, forgot about this. LGTM. Thanks so much for the bug fix! @bors r+

PR #396, opened 2017-08-31
Remove sub-expressions from parsing error

andrea-prearo:
This PR skips printing sub-expressions in the error message.

This should avoid situations where the error message contains an AST with unreadable characters, as shown [here](https://github.com/BurntSushi/ripgrep/issues/395).

unreadable-chars

### DIFF
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -1589,8 +1589,7 @@ impl fmt::Display for ErrorKind {
             RepeaterExpectsExpr =>
                 write!(f, "Missing expression for repetition operator."),
             RepeaterUnexpectedExpr(ref e) =>
-                write!(f, "Invalid application of repetition operator to: \
-                          '{}'.", e),
+                write!(f, "Invalid application of repetition operator."),
             UnclosedCaptureName(ref s) =>
                 write!(f, "Capture name group for '{}' is not closed. \
                            (Missing a '>'.)", s),
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -1574,9 +1574,8 @@ impl fmt::Display for ErrorKind {
                 write!(f, "Invalid character class range '{}-{}'. \
                            Character class ranges must start with the smaller \
                            character, but {} > {}", start, end, start, end),
-            InvalidClassEscape(ref e) =>
-                write!(f, "Invalid escape sequence in character \
-                           class: '{}'.", e),
+            InvalidClassEscape(_) =>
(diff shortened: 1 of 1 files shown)


andrea-prearo:
@BurntSushi This PR is a follow up to our [recent discussion](https://github.com/BurntSushi/ripgrep/pull/590) on `regex-syntax` error messages.
It is intended to be a temporary workaround until we can have a better regex parser/pretty printer.

regex-syntax/src/lib.rs
```diff
@@ -1589,8 +1589,7 @@ impl fmt::Display for ErrorKind {
             RepeaterExpectsExpr =>
                 write!(f, "Missing expression for repetition operator."),
             RepeaterUnexpectedExpr(ref e) =>
```
andrea-prearo: I didn't remove the unused variable `e` because I assumed it can be put back into the error message again once the issues with the AST pretty printing are solved. But, if it makes sense, I can modify the code further to convert `RepeaterUnexpectedExpr(Expr)` to `RepeaterUnexpectedExpr` and get rid of the unused variable.

BurntSushi: @andrea-prearo Right, but this produces a compile warning, which causes the build to fail. You definitely don't want to change the error type itself, but using `RepeaterUnexpectedExpr(_)` is the right thing to do here.

There are other errors that try to print sub-expressions. Could you fix those too? e.g., `InvalidClassEscape` and maybe more.

andrea-prearo: @BurntSushi I just noticed that warnings are treated as errors by looking at the CI log. I'll fix that by using `RepeaterUnexpectedExpr(_)` as you suggested. I'll also work on fixing other sub-expressions.

andrea-prearo:
I went through all the errors and didn't see other instances that try to print out sub-expressions.

andrea-prearo:
The build is failing on `beta` and `nightly` for reasons unrelated to these changes.

screen shot 2017-08-31 at 10 50 22 am

screen shot 2017-08-31 at 11 27 46 am

BurntSushi:
@andrea-prearo Right, yeah, I'll submit a PR when I get a chance fixing that.

andrea-prearo:
@BurntSushi Sounds good. Thanks!

BurntSushi:
@andrea-prearo Could you rebase this on master? That should hopefully get CI passing!

andrea-prearo:
@BurntSushi I had some issues with the rebase, so I had to do some clean up. But tests are passing now!

BurntSushi:
@andrea-prearo Thanks! @bors r+

PR #400, opened 2017-09-07
regex-syntax: update to Unicode 10.0

### LINKED ISSUE #391: Update to more recent version of Unicode
This still uses Unicode 7.0.0 which causes serious discrepancies in the handling of e.g. `XID_Start` and `XID_Continue` properties. For instance, `^\p{XID_Continue}$` doesn't match `U+17828`, whereas [unicode-xid](https://crates.io/crates/unicode-xid)'s `UnicodeXID::is_xid_continue()` correctly returns `true` for this code point.

cuviper:
Fixes #391


BurntSushi:
@cuviper Nice! Thanks. @bors r+

PR #403, opened 2017-10-05
[doc] Fix get() function name

behnam:
It feels like it's a left-over after some codemod, since there's no `at()` method on this type.


behnam:
It's here on live docs: https://doc.rust-lang.org/regex/regex/struct.Regex.html#examples-1

BurntSushi:
Ah yup, nice find! @bors r+

Issue #348, opened 2017-03-15
on short strings that request captures, don't run the DFA


BurntSushi:
Currently, if a caller requests captures, then we always run the DFA to determine the *extent* of the match and then we run either the Pike VM or the bounded backtracker on only the extent of the match to fill in the capture locations. This works well when searching long strings because the DFA can save the NFA engines from doing a lot of work. But on short strings, the DFA probably doesn't pay for itself, so we should just run one of the NFA engines if the string is short enough. (More precisely, in cases where the *match* is roughly the same length as the entire string, then the DFA isn't helping us at all since the NFA engine will still run the length of the match. But, obviously, this case isn't possible to know up front, so we use "short strings" as a likely predictor of that case.)

There should be some experimentation to determine where this boundary lies, so that we can invent a heuristic for when the string is "short enough."

arielb1:
If a regex is "anchored" (starting with a `^` and ending with a `$`), then there is never a point in determining the extent of the match. Anchored regexes are quite a common use-case when regexes are used for parsing/validation.

BurntSushi:
@arielb1 That's a good observation! We should add that to the heuristic as well.

One possible downside I just thought of: the DFA is still useful in the non-match scenario, since it can detect a match failure much more quickly than the NFA engines can.

BurntSushi:
For anyone interested in working on this, the relevant code is in the core implementation of the `read_captures_at` method: https://github.com/rust-lang/regex/blob/d813518e2a199884cd38a4e32497a7453db79697/src/exec.rs#L513

In particular, right now, it will always run the DFA engine if the DFA engine was selected by the compiler. The optimization in question here would be to add a conditional [in the DFA case](https://github.com/rust-lang/regex/blob/d813518e2a199884cd38a4e32497a7453db79697/src/exec.rs#L545) that checks if a string is below a certain length, and if it is, just skip directly to [running the NFA](https://github.com/rust-lang/regex/blob/d813518e2a199884cd38a4e32497a7453db79697/src/exec.rs#L551).

To pick the right length, I'd suggest writing a micro benchmark that:

1. Uses a regex with several capture groups.
2. Searches strings of varying lengths.

You may find it useful to construct the regex and the search text such that the regex matches the entire text. In particular, there are cases where running the DFA *and* the NFA is actually faster because the DFA can determine "not a match" more quickly than the NFA can.

[...]

PR #410, opened 2017-10-27
Issue 348

ethanpailes:
I took a crack at implimenting #348, but ended up running into some squirrly results with the benchmark that I wrote. During some runs the optimization behaved as expected, but for most there was no real corrilation between input size and performance. I suspect I'm missing some benchmarking trick, so an insight would be appreciated. The steps I used to test the benchmark are outlined in a comment above the new benchmarks I added (in `bench/src/misc.rs`).

For now I've just done the anchor optimization.

As I was learning about the code base I also noticed a few documentation references that had drifted out of date, so I fixed those.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -270,9 +270,9 @@ N.B. To run tests for the `regex!` macro, use:
 
 The benchmarking in this crate is made up of many micro-benchmarks. Currently,
 there are two primary sets of benchmarks: the benchmarks that were adopted
-at this library's inception (in `benches/src/misc.rs`) and a newer set of
+at this library's inception (in `bench/src/misc.rs`) and a newer set of
 benchmarks meant to test various optimizations. Specifically, the latter set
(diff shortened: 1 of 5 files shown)


BurntSushi:
@ethanpailes Thanks so much for putting in the work for this optimization! It might take me a bit to review this, but I just wanted to chime in. :-)

ethanpailes:
Unfortunatly, I did end up in a place where you have more review work than I would have hoped because of those weird benchmark results, so it makes sense that it might take a little while to review.

BurntSushi:
I've rolled this PR into https://github.com/rust-lang/regex/pull/436. Thanks!

BurntSushi:
@ethanpailes There should definitely be a way to observe a performance difference on unanchored regexes, but the anchored optimization is definitely a step in the right direction. I don't quite have time to do the benchmark analysis for unanchored regexes though. From briefly looking at your regexes, my gut says that you need to make the match bigger.

PR #413, opened 2017-10-31
Tiny PR: update a the ByteInput comment

ethanpailes:
At some ByteInput switched to being backed by actual bytes
rather than a string. This commit just removes the dangling
reference to the old representation.

I just noticed this while reading through code, so I thought I would send the patch your way.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -223,10 +223,6 @@ impl<'t> Input for CharInput<'t> {
 }
 
 /// An input reader over bytes.
-///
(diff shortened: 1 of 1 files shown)


BurntSushi:
Excellent, thank you! @bors r+

ethanpailes:
I tried to reproduce that test failure by switching to beta and running both `cargo test` and `./ci/script.sh`, but I coudn't reproduce the failure.

PR #416, opened 2017-11-15
gcc::Config -> gcc::Build

ethanpailes:
gcc::Config has been depreciated since gcc 0.3.51.
The change was introduced by commit
dc7162fe40e6b31811e8879c9860fd9eafe6aa46 in the gcc
crate. This patch just changes the bench build script
to use the new name.

### DIFF
--- a/bench/build.rs
+++ b/bench/build.rs
@@ -41,7 +41,7 @@ fn main() {
             we!("re2 cannot be found by pkg-config");
             process::exit(1);
         }
-        gcc::Config::new()
+        gcc::Build::new()
             .cpp(true)
             .flag("-std=c++11")
             .file("src/ffi/re2.cpp")


ethanpailes:
Closing because I just went back to look at the 0.3 release and it looks like Build was not yet there. This shouldn't get switched over until bench depends on a new enough version of the GCC crate.

BurntSushi:
@ethanpailes Sorry, but this doesn't make any sense to me. This should, in theory, require a version bump in `Cargo.toml`, no? I think if you put `gcc = "0.3.51"` in the `Cargo.toml`, then this change should actually be OK. But yeah, in general, we should move to `cc`, but its minimum Rust version is `1.13`, while `regex` is still on `1.12`.

ethanpailes:
@BurntSushi Yeah I noticed that version issue right after I opened this PR, so I closed it quickly. It sounds like you were already aware of the issue.

BurntSushi:
@ethanpailes I wasn't actually. :P I was peripherally aware of the move from `gcc` to `cc` though.

Issue #408, opened 2017-10-15
investigate using boyer moore (for real) on longer query strings


BurntSushi:
See: https://github.com/BurntSushi/ripgrep/issues/617

ethanpailes:
I'm not sure that this will ever be worth it. I implemented tuned Boyer Moore in a [branch](https://github.com/ethanpailes/regex/tree/boyer-moore-research) on my fork to see if I could produce a speedup that could live behind a heuristic for pattern length, but the only case where I could get my TBM to win consistently was on the random data and large needle case. Whenever the text looked like natural language or the needle was small, TBM lost.

When I used a [needle size of 1000](https://github.com/ethanpailes/regex/blob/04c24870dcf415d9b5c529c49e78bde8408710f9/boyer-moore/bench/src/bench.rs#L184) TBM won on both random bytes and lorum ipsum, but it still lost on Sherlock.

I can't thing of any real use case where you are searching in random seeming data (maybe looking for a specific base64 encoded string in a big blob of base64 data). I would say genetic data is a use case, but according to J Moore BM and friends are (not the right choice)[http://www.cs.utexas.edu/users/moore/publications/sustik-moore.pdf] for that either.

Alternatively, I could have missed some optimizations that seem clearer to someone with more experience writing fast string search algorithms.

BurntSushi:
> I can't thing of any real use case where you are searching in random seeming data (maybe looking for a specific base64 encoded string in a big blob of base64 data).

One case I can think of is when you're searching for a hash in a file that contains a list of hashes. 

Another case that might be worth benchmarking is a needle that contains only common characters. There might be a win there with TBM that *doesn't* use memchr at all (because memchr has some amount of overhead, so if you call it frequently enough, it should be self defeating). But to be clear, this is only my intuition, which could be very wrong!

> Alternatively, I could have missed some optimizations that seem clearer to someone with more experience writing fast string search algorithms.

Just from glancing over the code, it looks pretty good to me. Very nice work. Do you want to work on trying to get it merged? We could start conservatively by only using BM when the needle is over a certain size. I'm not sure if I'd want it to be a separate crate just yet, but we could certainly work toward that.

ethanpailes:
> Another case that might be worth benchmarking is a needle that contains only common characters.

That's a good point. I will write a benchmark for that to see if it is a win. Another nice thing about that is that it would be pretty easy to have a factory method that looks at the pattern and switches to TBM if it does not have a sufficiently rare char.

[...]

PR #419, opened 2017-11-29
Add an implimentation of Tuned Boyer-Moore.

ethanpailes:
While the existing literal string searching algorithm
leveraging memchr is quite fast, in some case more
traditional approaches still make sense. This patch
provides an implimentation of Tuned Boyer-Moore as
laid out in Fast String Searching by Hume & Sunday.
Some refinements to their work were gleened from the
grep source.

See: #408 
See: [ripgrep-617](https://github.com/BurntSushi/ripgrep/issues/617)


bench/Cargo.toml
```diff
@@ -5,7 +5,7 @@ version = "0.1.0"
 authors = ["The Rust Project Developers"]
 license = "MIT/Apache-2.0"
 repository = "https://github.com/rust-lang/regex"
-documentation = "http://doc.rust-lang.org/regex/regex_syntax/index.html"
+documentation = "http://doc.rust-lang.org/regex/regex/index.html"
```
ethanpailes: This is just a random nit I had when reading the cargo file. It seems odd for the benchmark suite to point to the syntax crate for documentation. I'm not at all attached to this change.

src/lib.rs
```diff
@@ -514,7 +514,7 @@ another matching engine with fixed memory requirements.
 extern crate aho_corasick;
 extern crate memchr;
 extern crate thread_local;
-#[cfg(test)] extern crate quickcheck;
+#[macro_use] #[cfg(test)] extern crate quickcheck;
```
ethanpailes: Required to use the `quickcheck!` macro. If there is a reason to avoid pulling quickcheck macros I can refactor to use the `quickcheck` function.

src/literals.rs
```diff
@@ -392,7 +402,7 @@ impl SingleByteSet {
 ///
 /// TODO(burntsushi): Add some amount of shifting to this.
 #[derive(Clone, Debug)]
-pub struct SingleSearch {
+pub struct MemchrSearch {
```
ethanpailes: I renamed this guy because I thought the old name would be confusing in the presense of multiple search algorithms focused on finding a single literal.

ethanpailes:
@BurntSushi, do you have time to give this a quick "seems worthwhile" or "not right now"? It doesn't have to be a full review.

BurntSushi:
@ethanpailes Thanks for the ping! Looking over this, I think this actually looks pretty dang good. I don't have any specific comments so I think we should just bring it in. Thank you so much for doing this. :-) @bors r+

ethanpailes:
@BurntSushi, assuming tests pass, I'm now happy with this. The turnover length for patterns of uncommon chars seems to be around 7 or 8.

I want to take another crack at finding an absolute length turnover where TBM always wins even if the pattern chars are not that common, but that can go in another PR.

BurntSushi:
@ethanpailes Looks like quickcheck caught a bug?

ethanpailes:
So it does, I'll take another look.

ethanpailes:
I managed to blow away all my bug fixes when I updated the heuristic. I'm so good at git. Sorry about that.

ethanpailes:
@BurntSushi, I've had quickcheck running in a loop locally for a few hours, so it should be good to go now.

BurntSushi:
Thanks! Awesome work! @bors r+

PR #421, opened 2017-11-29
Introduce `use_std` default feature

vmx:
regex depends on memchr and aho-corasick which can optionally be compiled
without std. With the introduction of the `use_std` feature it is now possible
to compile regex with `default-features = false` which then uses a memchr
and aho-corasick that don't depend on std.

This makes regex available on platforms like wasm which haven't a libc
implemented.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,9 @@ appveyor = { repository = "rust-lang-libs/regex" }
 
 [dependencies]
 # For very fast prefix literal matching.
-aho-corasick = "0.6.0"
+aho-corasick = { version = "0.6.0", default-features = false }
 # For skipping along search text quickly when a leading byte is known.
-memchr = "2.0.0"
+memchr = { version = "2.0.0", default-features = false }
 # For managing regex caches quickly across multiple threads.
 thread_local = "0.3.2"
 # For parsing regular expressions.
@@ -40,10 +40,13 @@ quickcheck = { version = "0.4.1", default-features = false }
 rand = "0.3.15"
 
 [features]
+default = ["use_std"]
 # Enable to use the unstable pattern traits defined in std.
 pattern = []
 # Enable to use simd acceleration.
 simd-accel = ["simd"]
+# Allow compilation of dependencies without std
+use_std = ["aho-corasick/use_std", "memchr/use_std"]
 
 [lib]
 # There are no benchmarks in the library code itself


Cargo.toml
```diff
@@ -19,9 +19,9 @@ appveyor = { repository = "rust-lang-libs/regex" }
 
 [dependencies]
 # For very fast prefix literal matching.
-aho-corasick = "0.6.0"
+aho-corasick = { version = "0.6.0", default-features = false }
```
BurntSushi: I think this is blocked on aho-corasick getting a new release. `regex` would then need to depend on at least the newly released version.

vmx:
Please note that this pull request needs a new release of aho-corasick which just merged the corresponding commit.

Cargo.toml
```diff
@@ -40,10 +40,13 @@ quickcheck = { version = "0.4.1", default-features = false }
 rand = "0.3.15"
 
 [features]
+default = ["use_std"]
 # Enable to use the unstable pattern traits defined in std.
 pattern = []
 # Enable to use simd acceleration.
 simd-accel = ["simd"]
+# Allow compilation of dependencies without std
+use_std = ["aho-corasick/use_std", "memchr/use_std"]
```
BurntSushi: So now that I'm thinking about this (and this comment applies to the `aho-corasick` crate as well), I'm thinking that `use_std` might not be an appropriate name. It's surely appropriate for `memchr` since the absence of `use_std` actually means that `memchr` won't use `std` at all. But in the case of `aho-corasick` and `regex`, that's not what `use_std` means at all. Instead the absence of `use_std` in `regex` and `aho-corasick` seems to mean that it won't explicitly import from `libc` anywhere, which is quite a bit different than not using `std`.

I wonder if it makes sense to add a `use_libc` feature instead (and similarly for `aho-corasick`). That would then in turn have to imply the `memchr/use_std` feature.

andrew-d:
I would love to have proper `no_std` support for embedding `regex` into other things (e.g. compiling as a static library for embedding into other languages)!  However, it appears that there's a fair bit more to do; there appear to still be dependencies on other things that use `std`, both in dependencies and the library.  (for testing, I created the following project: https://github.com/andrew-d/regex-rustgo)

In third party libraries:

- `lazy_static` still depends on `std`, though this can be solved by using the `spin_no_std` feature
- `utf8-ranges` still depends on `std` - see [here][u8].
- `void` still depends on `std`, but you can solve this by using `default-features = false` (see [here][void]).

And then there's the obvious problem of the varied `use std::whatever` throughout the `regex` library itself 😄   Happy to try getting proper `no_std` support, but I think that it might be fairly tricky... thoughts?

[u8]: https://github.com/BurntSushi/utf8-ranges/blob/master/src/lib.rs#L94-L96
[void]: https://github.com/reem/rust-void/blob/master/Cargo.toml#L14

BurntSushi:
@andrew-d Uh, could you please open another issue? What you're talking about seems completely orthogonal to the problem that @vmx is trying to solve.

Spoiler alert: allocation is weaved into the very guts of regex, and I don't see that ever changing, so the very best you'll be able to do is `no_std` but with `alloc`. A true `no_std` or `core`-only regex library is a special beast that should, at the very least, be prototyped out-of-tree by someone else. :) --- I'm happy to discuss this more if you have more thoughts, but please open another issue.

alexcrichton:
Interesting! I think for something like wasm it's probably best to "work by default" rather than detect `use_std` or not though perhaps?

BurntSushi:
Could you unpack that? I think you're saying that memchr should make libc a dep for every target except wasm?

alexcrichton:
Oh sorry sure yeah. So in general if it can every crate I think should compile every platform by default. For example that means that memchr wouldn't have a libc dep on the wasm platform (but it would on Linux for example). That way you don't have to worry about fiddling with features for portability, but rather the crates choose a "reasonable default" and try to be portable by default.

Similarly the `regex` crate would in theory compile-by-default for wasm (for whatever definition that may entail).

It's tricky when a platform doesn't present some functionality though, but thankfully I think that's not the case for wasm?

BurntSushi:
@alexcrichton Ah OK, I think I just completely forgot about specifying target specific dependencies. I agree that that seems like what we should do here. In this case, `memchr` has fallbacks in the absence of libc so it's just a matter of setting the right cfgs I think.

@vmx So I think the solution to your problem here is:

1. Rollback the change to `aho-corasick`.
2. Make `libc` a target specific dependency.

For (2), the docs for that I think are here: http://doc.crates.io/specifying-dependencies.html#platform-specific-dependencies --- So I'd guess you'd want (assuming `not(target_arch = "wasm")` is actually the right thing here):

```toml
[target.'cfg(not(target_arch = "wasm"))'.dependencies]
libc = { version = "0.2.18", default-features = false, optional = true }
```

vmx:
@BurntSushi Thanks for the detailed explanation. I'll take a stab.

alexcrichton:
Oh to clarify I woudln't worry about target-specific dependencies in the sense that `libc` compiles on the wasm target it's just empty, so you'll just need to `#[cfg]` internally to not actually use `libc`

vmx:
There it is: https://github.com/BurntSushi/rust-memchr/pull/23
It works locally but needs a few new releases of various other crates.

vmx:
I'm closing this one in favour of #422, which is a lot more on-topic.

PR #439, opened 2017-12-31
Add support for benchmarking D's std.regex

### LINKED ISSUE #430: Add benchmarks for D's std.regex
It would be interesting to include benchmarks for D's std.regex.

I have initial work for it, but it is not ready to be a pull request, and I probably won't find the time to clean it up:

Branch: https://github.com/mrmonday/regex/tree/d-phobos-benchmarks
Diff:  https://github.com/rust-lang/regex/compare/master...mrmonday:d-phobos-benchmarks

General notes/notes for Rustaceans who don't read D:

 * Phobos is the name of D's standard library
 * There are 2*2 cargo features, which specify the compiler (`ldc` (LLVM) vs `dmd` (Digital Mars/Reference compiler)), and the regex mode - runtime (no extension), or generated at compile time (`-ct`). There's also a gdc (GNU based) compiler, but it's not in the arch repo's, so I didn't test it.
 * Various things are hardcoded/half-baked "just make it work"
 * There's a `src/ffi/build_d.sh` file with the commands needed to build things on the D side. These should be moved into the `build.sh`
 * Everything was built/tested using the ldc/dmd versions from the Arch Linux repositories
    ```
    # ldc
    LDC - the LLVM D compiler (1.6.0):
      based on DMD v2.076.1 and LLVM 5.0.0
      built with DMD64 D Compiler v2.077.0
    # dmd
    DMD64 D Compiler v2.077.1 
    ```
 * Most of the regex is rather hackily parsed directly out of the Rust files on the D side at compile time, with some of the regex hard coded on the D side where it's anything but a string literal. It's probably easier/quicker to hard code it, but that's not what I did.
 * All the misc benchmarks run, most of sherlock runs (still some work to do), I haven't tested regexdna
 * Documentation for D's std.regex is here: https://dlang.org/phobos/std_regex.html

I'm happy to answer any questions to help this along, if I can.

mrmonday:
This PR adds support for benchmarking the runtime version of the D
programming language's std.regex using the dmd and ldc compilers.

Sample results on a Core i7-7280HQ:
 * Rust: [rust.results.txt](https://github.com/rust-lang/regex/files/1595383/rust.results.txt)
 * Rust bytes: [rust-bytes.results.txt](https://github.com/rust-lang/regex/files/1595382/rust-bytes.results.txt)
 * DMD: [dmd.results.txt](https://github.com/rust-lang/regex/files/1595386/dmd.results.txt)
 * LDC: [ldc.results.txt](https://github.com/rust-lang/regex/files/1595384/ldc.results.txt)

Comparisons:
 * Rust vs LDC: [rust-vs-ldc.txt](https://github.com/rust-lang/regex/files/1595381/rust-vs-ldc.txt)
 * Rust Bytes vs LDC: [rust-bytes-vs-ldc.txt](https://github.com/rust-lang/regex/files/1595388/rust-bytes-vs-ldc.txt)
 * DMD vs LDC: [dmd-vs-ldc.txt](https://github.com/rust-lang/regex/files/1595385/dmd-vs-ldc.txt)

LDC is LLVM based, so should give the "most fair" results in terms of comparisons. I haven't checked that the exact flags/same optimisation levels are used.

This PR does not include the CTFE versions of the benchmarks - I may add these in a follow up PR. Older benchmarks did not show a definitive speed up - some benchmarks were faster (compared to runtime variants), some were slower. The compile times were noticably larger.

Closes #430


mrmonday:
The build failure seems to be a single Travis build timing out - it looks unrelated to my changes. I don't have access to retry it, otherwise I would.

BurntSushi:
CI is on the fritz. I'll look into it. Definitely unrelated. I pulled this PR down and tried it myself. It worked perfectly! Thanks so much for doing this! (I think this would have taken me personally 100 times as long to do as you!)

PR #444, opened 2018-01-26
Avoid depending on expression Debug stringifier in release builds

bholley:
The code weighs over a kilobyte, which doesn't seem worth it for an unreachable.

### DIFF
--- a/regex-syntax/src/parser.rs
+++ b/regex-syntax/src/parser.rs
@@ -1227,7 +1227,7 @@ impl Parser {
                             return Err(self.errat(
                                 chari, ErrorKind::UnclosedParen));
                         }
-                        e => unreachable!("{:?}", e),
+                        _ => unreachable!(),
                     }
                 }
                 Some(Build::Expr(e)) => { concat.push(e); }


BurntSushi:
@bholley I don't understand this change, could you explain it a bit more? If it's just about decreasing binary size, then could the same be accomplished by stripping the binary? Or else, even if I accept this, what's to stop something like this from appearing again?

bholley:
> If it's just about decreasing binary size, then could the same be accomplished by stripping the binary? 

Not sure what you mean by this. The issue is that panics are reachable in release-mode code, and can't be eliminated by the linker.

> Or else, even if I accept this, what's to stop something like this from appearing again?

Ideally by having people try to avoid {:?} in panic messages and release-mode logging. I do agree that it's pretty hard to avoid. I'd been hoping we could set up a continuous integration hook for Firefox, but I think we'd end up tripping it a lot with third-party crates. It's really a footgun in the language.

So on our end, we'll likely try to do our best with code review and accept that it won't be perfect. I think it's probably impractical to do too much for third-party crates, but I spent about 15 minutes with binutils tracking this one down, so I figured I might as well send it along and save a kilobyte for everybody using regex.

bholley:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1430814

BurntSushi:
OK, I guess I'll take this, but I wouldn't expect me to be particularly vigilant about this "footgun" going forward.

PR #447, opened 2018-02-09
Closes #446

### LINKED ISSUE #446: tuned boyer moore fails a match case
Here is a failing test case:

```rust
    #[test]
    fn bm_backstop_boundary() {
        let haystack = b"\
// aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
e_data.clone_created(entity_id, entity_to_add.entity_id);
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
".to_vec();
        let needle = b"clone_created".to_vec();

        let searcher = BoyerMooreSearch::new(needle);
        let result = searcher.find(&haystack);
        assert_eq!(Some(43), result);
    }
```

This was originally reported against ripgrep:

* https://github.com/BurntSushi/ripgrep/issues/780
* https://github.com/BurntSushi/ripgrep/issues/781

Both bugs are related to TBM. That is, shutting TBM off fixes the issue. The above unit test was derived by widdling down the example corpus given in https://github.com/BurntSushi/ripgrep/issues/781

I've tried to fix this myself, but the code is dense. As best I can tell, the md2 shift is wrong, or this is hitting up into a boundary condition with respect to the backstop (more likely?). That is, this particular case enters the `memchr` skip routine, which uses `_` as its guard char. This takes `window_end` from `12` to `44`, which is positioned at the `l` in `clone_created`. This then fails `check_match` (which seems correct), but this in turn causes the `md2_shift` (which is `12`) to get added to the `window_end` index, which causes the code to skip ahead too far and miss the match.

The part that doesn't make sense to me here is the relationship between `check_match` and the `md2_shift`. Namely, `check_match` appears to be used as a guard against whether `md2_shift` is used. If the match fails, then the code assumes we can skip `md2_shift` bytes. It almost seems like the case analysis is a little off. For example, this code seems to fix this particular bug:

```rust
                if self.check_match(haystack, window_end) {
                    return Some(window_end - (self.pattern.len() - 1));
                } else if window_end + self.md2_shift >= backstop {
                    break;
                } else {
                    window_end += self.md2_shift;
                    continue;
                }
```

but I didn't arrive to this code in a principled way, so I'm not confident in it. It feels like to me the very fact that the `md2_shift` could overshoot the match is the problem, but it's less clear how to fix that.

@ethanpailes Do you have any ideas here? I am probably going to disable TBM for now  because of this bug. Would it make sense to use a more traditional TBM variant where we don't use frequency analysis but instead always skip based on the last byte?

ethanpailes:
Thanks @BurntSushi for doing all the legwork on cutting down the failing test case and laying out a clear set of execution steps. It really helped me zero in on the bug.

This patch fixes an issue where skip resolution would go strait
to the default value (the md2_shift) on a match failure after
the shift_loop. Now we do the right thing, and first check in
the skip table. The problem with going strait to the md2_shift
is that you can accidentally shift to far when `window_end`
actually is in the pattern (as is the case for the failing
match).

In the issue thread I promised better modularity, but it turns
out that shift resolution was a bit too decomposed in the other
places I had mentioned. Sorry :(.

### DIFF
--- a/src/literals.rs
+++ b/src/literals.rs
@@ -627,7 +627,9 @@ impl BoyerMooreSearch {
                 if self.check_match(haystack, window_end) {
                     return Some(window_end - (self.pattern.len() - 1));
                 } else {
-                    window_end += self.md2_shift;
+                    let skip = self.skip_table[haystack[window_end] as usize];
+                    window_end +=
+                        if skip == 0 { self.md2_shift } else { skip };
                     continue;
                 }
             }
@@ -946,7 +948,6 @@ mod tests {
     }
 
     #[test]
-    #[should_panic]
     fn bm_backstop_boundary() {
         let haystack = b"\
 // aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa


ethanpailes:
Local testing: I ran the unit tests both at just this commit and with TBM turned back on. I'm currently running the quickcheck tests in a loop.

BurntSushi:
>  I'm currently running the quickcheck tests in a loop.

Did you [see my comment about the quickcheck tests](https://github.com/rust-lang/regex/issues/446#issuecomment-364279717)? Frankly, I'm surprised you got as much mileage out of them as you did. :-)

ethanpailes:
Yeah, I agree about the quickcheck tests not being that likely to catch any remaining bugs. I'm only running them because I already have a hammer, so why not thrash about to try to find a nail that this change could have turned up.

I think the principled thing to do would be to hook them into a fuzzer somehow? Those have an actual model of the state space and tweak the input to more fully exersise it right?

BurntSushi:
I don't have much experience with fuzzers, but with an `Arbitrary` instance, you can encode whatever model of state you choose. :)

ethanpailes:
I think fuzzers do some code coverage deep magic, but I also don't really have experience with them.

ethanpailes:
@BurntSushi bump

BurntSushi:
@ethanpailes Thanks again!

PR #448, opened 2018-02-10
[WIP] Add a quickcheck property to better test the positive case

ethanpailes:
I've marked this as a WIP because I want to wait until #447 get merged and then rebase on top.

I've rolled back the fix from #447 locally and am running this new test in a loop to see if it will catch the bug. Unfortunatly, it does not fall out immediately.

If you just generate two random strings, the odds are very high
that the shorter one won't be a substring of the longer one once
they reach any substantial length. This means that the existing
quickcheck cases were probably just testing the negative cases.
The exception would be the two cases that append the needle
to the haystack, but those only test behavior at the ends. This
patch adds a better quickcheck case that can test a needle anywhere
in the haystack.
    
See the comments on #446


ethanpailes:
Good news! After letting it run for a few minutes this test did in fact dig up the bug! I'm going to use it to look for more.

BurntSushi:
@ethanpailes Can you rebase this on master? Also, your last comment mentions another bug---any head way on that?

ethanpailes:
Oh no, it was the same bug. I wrote this prop test with the goal of catching bugs like the one from #446, and it I was just noting that it did in fact catch the bug.

I'll rebase on master shortly.

Issue #83, opened 2015-05-10
implement &T for Replacer where T: Replacer


RCIX:
When attempting to call replace_all on a Regex with a str as the argument for rep(as opposed to a NoExpand), the Replacer trait implementation for str is apparently being ignored, instead falling back to the FnMut implementation. [This project](https://github.com/RCIX/regex_replace_error) demonstrates the following compile error:

    src\main.rs:13:23: 13:53 error: the trait `for<'r,'r> core::ops::Fn<(&'r regex::re::Captures<'r>,)>` is not implemented for the type `str` [E0277]
    src\main.rs:13         true => regex.replace_all(&source, &replace.as_str()),
 
Where replace and source are both Strings.

Edit: just fixed the type of replace by adding as_str, as the & was not coercing the String to a str properly. See the latest commit.

BurntSushi:
If you use `replace.as_str()` instead of `&replace.as_str()`, then your code works. The `as_str` method returns a `&str`, so `&replace.as_str()` has type `&&str`, and no impl for that exists.

I wonder if we can add a `impl<'a, T: Replacer> &'a T for Replacer { ... }` impl that would fix this. The standard library uses this convention AFAIK.

BurntSushi:
I feel like it's probably not worth adding this impl unless there's a more solid reason. (It could be added in a backwards compatible manner later if we really wanted to, I think.)

PR #449, opened 2018-02-14
Add Replacer::by_ref adaptor to use a Replacer without consuming it

mbrubeck:
This is useful when you want to take a generic `Replacer` (which might not be cloneable) and use it without consuming it, so it can be used more than once.

Note: This can't simply return `&mut Self` because a generic `impl<R: Replacer> Replacer for &mut R` would conflict with libstd's generic `impl<F: FnMut> FnMut for &mut F`.

### DIFF
--- a/src/re_unicode.rs
+++ b/src/re_unicode.rs
@@ -1034,6 +1034,43 @@ pub trait Replacer {
     fn no_expansion<'r>(&'r mut self) -> Option<Cow<'r, str>> {
         None
     }
+
+    /// Return a `Replacer` that borrows and wraps this `Replacer`.
+    ///
+    /// This is useful when you want to take a generic `Replacer` (which might
+    /// not be cloneable) and use it without consuming it, so it can be used
+    /// more than once.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use regex::{Regex, Replacer};
+    ///
+    /// fn replace_all_twice<R: Replacer>(re: Regex, src: &str, mut rep: R)
+    /// -> String {
+    ///     let dst = re.replace_all(src, rep.by_ref());
+    ///     let dst = re.replace_all(&dst, rep.by_ref());
+    ///     dst.into_owned()
+    /// }
+    /// ```
+    fn by_ref<'r>(&'r mut self) -> ReplacerRef<'r, Self> {
+        ReplacerRef(self)
+    }
+}
+
+/// By-reference adaptor for a `Replacer`
+///
+/// Returned by [`Replacer::by_ref`](trait.Replacer.html#method.by_ref).
+#[derive(Debug)]
+pub struct ReplacerRef<'a, R: ?Sized + 'a>(&'a mut R);
+
+impl<'a, R: Replacer + ?Sized + 'a> Replacer for ReplacerRef<'a, R> {
+    fn replace_append(&mut self, caps: &Captures, dst: &mut String) {
+        self.0.replace_append(caps, dst)
+    }
+    fn no_expansion(&mut self) -> Option<Cow<str>> {
+        self.0.no_expansion()
+    }
 }
 
 impl<'a> Replacer for &'a str {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -527,7 +527,7 @@ pub use re_trait::Locations;
 pub use re_unicode::{
     Regex, Match, Captures,
     CaptureNames, Matches, CaptureMatches, SubCaptureMatches,
-    Replacer, NoExpand, Split, SplitN,
+    Replacer, ReplacerRef, NoExpand, Split, SplitN,
     escape,
 };
 
--- a/src/re_bytes.rs
+++ b/src/re_bytes.rs
@@ -990,6 +990,43 @@ pub trait Replacer {
     fn no_expansion<'r>(&'r mut self) -> Option<Cow<'r, [u8]>> {
         None
     }
+
+    /// Return a `Replacer` that borrows and wraps this `Replacer`.
+    ///
+    /// This is useful when you want to take a generic `Replacer` (which might
+    /// not be cloneable) and use it without consuming it, so it can be used
+    /// more than once.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use regex::bytes::{Regex, Replacer};
+    ///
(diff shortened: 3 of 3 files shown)


mbrubeck:
#83 proposed something similar but without the same motivation.

albel727:
Here's a specific problem that this solves.

https://play.rust-lang.org/?gist=3a4d8af0b9ee6b8faa984a6966eb2728&version=stable

mbrubeck:
I wrote this based on @albel727's code which was given in IRC.  @albel727, is that based on a real-world use case, or is it an exercise/toy program?

Another use case might be applying the same replacement to many different input strings, though I don't have a more concrete example of this in mind.

albel727:
It was a real use case, that I simplified into the example on the playground. 

ReplacerRef allows to reuse the same Replacer in generic context, which in this case fulfills the need to repeatedly apply the same Regex and Replacer to a string until it stops applying, because the replacement may result in appearance of something that also needs to be replaced.

I think implementing such a fixpoint operator for regex replacement is a legitimate need. Consider this example, which is closer to what I needed.

Suppose you have something like `"f(1)f(2)f(3)f(4)g(5)f(6)f(7)"`, and you need to merge adjacent `f()` contents, i.e. `"f(1)f(2)f(3)"` -> `"f(123)"`, and `"f(1)f(2)f(3)f(4)g(5)f(6)f(7)"` -> `"f(1234)g(5)f(67)"`. 

It can't be written as a regex with single replacement, but can be written as a repeated replacement on pairs of adjacent `f()`, i.e. 
```Rust
loop { // while Regex applies
    s = Regex::new(r"f\(([^)]*)\)f\(([^)]*)\)").unwrap().replace(s, "f($1$2)");
}
```

mbrubeck:
> As far as the PR itself, it looks like this doesn't work on Rust 1.12.

Fixed.

> Also, could you wrap all the doc strings to 79 (inclusive) columns? Thanks.

Done.

BurntSushi:
All right, I think I'm OK with this. The use case seems a little out there, but since there's no other way to achieve it and this is a very small API addition, then I think I'm OK with this. 

Thanks @mbrubeck!

albel727:
Well, another way to define the use-case is emulating "overlapping replacement", with the repeated application of a non-overlapping replacement, which `replace()`/`replace_all()` are. I'm convinced it's actually a pretty common thing to want.

albel727:
I'm not getting there, am I? :smile: 

Well, suppose you want to remove space between every two digits, e.g 
`"abc 1 2 3 de 4 5 fghi 6 7 8 9"` -> 
`"abc 123 de 45 fghi 6789"`. 

You can only do that with repeated application of `replace_all()`, replacing `(\d)\s(\d)` with `$1$2`. Single application would leave you with 
`"abc 12 3 de 45 fghi 67 89"`,
because pairs of digits overlap.

This is actually a simpler problem than what repeated application allows to solve in general, because it could be theoretically solved with a single `replace_all()` on a regex with lookahead. But many replacement problems can't be solved with lookahead, and `Regex` implementation doesn't have lookahead, afaik.

The want for repeated application often arises when one wants to cheat and transform some simple subset of context-free grammar language with regular expressions. E.g. merge contents of consecutive tags in some simple html, or merge consecutive `extern "C" { fn f1() } extern "C" { fn f2() }  extern "C" { fn f3() }` in bindgen-generated Rust sources. 

Alternative would be firing whole HTML/Rust parsers and transforming resulting ASTs, which is often an overkill for the task.

BurntSushi:
@albel727 No, I get it. :) I'm just confused at how you could possibly insist that something like this is common! I've been programming for a long time, and in all that time, neither myself nor any of the code I've read have ever needed something like this out of a regex library. Just because something is niche or uncommon doesn't mean I think we automatically shouldn't have it, but it definitely weighs in the cost-benefit analysis. The fact that you've presented some interesting examples, and have demonstrated that you can't really do what you want without this addition is sufficient enough for me.

But, it's impossible to truly know. After this makes it into the next release, let's come back here in two years and look at all of regex's dependents. Count me shocked if we find more than a single digit number of uses. :-)

src/re_unicode.rs
```diff
@@ -1034,6 +1034,40 @@ pub trait Replacer {
...
         None
     }
+
+    /// Return a `Replacer` that borrows and wraps this `Replacer`.
+    ///
+    /// This is useful when you want to take a generic `Replacer` (which might
+    /// not be cloneable) and use it without consuming it, so it can be used
+    /// more than once:
+    ///
+    /// ```
```
BurntSushi: Sorry to nit at these things, but could you add an `# Example` header here to be consistent with how other examples are done? (I believe this is also the style in `std` too.)

src/re_unicode.rs
```diff
@@ -1034,6 +1034,40 @@ pub trait Replacer {
...
+    /// use regex::{Regex, Replacer};
+    ///
+    /// fn replace_all_twice<R: Replacer>(re: Regex, src: &str, mut rep: R)
+    /// -> String {
+    ///     let dst = re.replace_all(src, rep.by_ref());
+    ///     let dst = re.replace_all(&dst, rep.by_ref());
+    ///     dst.into_owned()
+    /// }
+    /// ```
```
BurntSushi: Nice, thank you for writing out this example! Could you also add it to the `re_bytes` version too?

src/re_unicode.rs
```diff
@@ -1034,6 +1034,40 @@ pub trait Replacer {
...
+    fn by_ref<'r>(&'r mut self) -> ReplacerRef<'r, Self> {
+        ReplacerRef(self)
+    }
+}
+
+/// By-reference adaptor for a `Replacer`
+///
+/// Returned by [`Replacer::by_ref`](trait.Replacer.html#method.by_ref).
+pub struct ReplacerRef<'a, R: ?Sized + 'a>(&'a mut R);
```
BurntSushi: Does it make sense to add `#[derive(Debug)]` to this? (And do the same for the `re_bytes` module.)

mbrubeck:
Nits fixed.

albel727:
Alright, fairy nuff. :) Though I've needed this quite often in my years of programming, but maybe it's just me.

So now I can feel at rest, knowing that one will always be able to implement one-dimensional cellular automata with `Regex`, whenever they wanted. :rofl:  Out of curiosity, when is this expected to land?

PR #454, opened 2018-03-09
Add builder method for `nest_limit`.

meh:
I didn't know where to put the test case but I think crazy is fairly accurate (it's one of thousands of regular expressions coming from libphonenumber).

Also I know this may sound silly, but could this get a version bump too please? The `phonenumber` crate cannot parse phone numbers anymore and it was through a minor bump, so I'm kind of stuck.


tests/crazy.rs
```diff
@@ -103,3 +103,300 @@ fn dfa_handles_pathological_case() {
...
+          (?:
+            1|
+            [24]1
+          )\d{2}
+        )\d{3}
+        "#
+    )
+    .nest_limit(1024)
+    .build()
+    .unwrap();
```
BurntSushi: Hmmm just from glancing it kind of looks like this regex isn't that deeply nested. I wonder if these is a bug in the logic that computes the nest level. What nest level is reported for this regex?

meh: How do I get the nest level? The error only mentions the maximum nest level (or I can't see it).

BurntSushi: Ah I see, then it's not exposed. Requires debugging. I'd like to explore that before merging this. With that said, even if your bug wasn't a thing, exposing the nest level seems right to me.

meh: Yeah, there are some really huge regular expressions in there, so I wouldn't be suprised if some of them went beyond the nest limit in case it is indeed a bug.

BurntSushi:
@meh It turns out this was indeed a (quite embarrassing!) bug and your regex definitely should not have tripped the limit. It's fixed on master. I should be able to get a new release out before the weekend is through. If not, please ping me. :-)

BurntSushi:
@meh Errmm, I broke up your commit into two pieces but forgot to fix the author information. Sorry about that! :-(

meh:
@BurntSushi no worries, as long as it's fixed :panda_face:

meh:
@BurntSushi release ping :bell:

PR #470, opened 2018-04-28
Fix anchor bug in Match iteration

ethanpailes:
The Match literal iterator would repeatedly look for matches
in the remainder of the input after it found its first
match regardless of whether or not the regex was anchored
at the start. This patch adds logic to make sure that we
don't keep looking for matches after the first match is
returned for a start-anchored literal regex.

This makes the failing tests pass and seems like the right thing to do
to me, but there could be a better place to fix the issue.


BurntSushi:
Thank you!

Issue #468, opened 2018-04-23
Test Regex Scraped from crates.io


ethanpailes:
As part of evaluating my masters thesis, I scraped `crates.io` for regex and ran the resulting regex though my compiler to see how many could be optimized. I don't think it would be too hard to clean up my scraping script a bit and then write a test which executes each of the regex on a quickcheck generated string with all (3? 5? 6? depends how you count them) backends. I'm not sure when I'll do this, but I wanted to leave a note here so that I don't forget.

ethanpailes:
Hmm. Good point. One thing I remember reading in a note Russ Cox made about the testing approach for RE2 is that he wrote some code to construct a random matching string from a regex, so it would might be worthwhile to give that a crack. The two sources of random input would probably do a pretty good job of testing both the positive and negative cases.

BurntSushi:
Yup, that's another good avenue to try!

ethanpailes:
After a first pass at this just using quickcheck to generate random input, I've turned up the following failing test cases.

```rust
extern crate regex;

#[test]
fn word_boundary_backtracking_default_mismatch() {
    use regex::internal::ExecBuilder;

    let backtrack_re = ExecBuilder::new(r"\b")
        .bounded_backtracking()
        .build()
        .map(|exec| exec.into_regex())
        .map_err(|err| format!("{}", err))
        .unwrap();

    let default_re = ExecBuilder::new(r"\b")
        .build()
        .map(|exec| exec.into_regex())
        .map_err(|err| format!("{}", err))
        .unwrap();

    let input = "䅅\\u{a0}";

    let fi1 = backtrack_re.find_iter(input);
    let fi2 = default_re.find_iter(input);
    for (m1, m2) in fi1.zip(fi2) {
        assert_eq!(m1, m2);
    }
}

#[test]
fn uppercut_s_backtracking_bytes_default_bytes_mismatch() {
    use regex::internal::ExecBuilder;

    let backtrack_bytes_re = ExecBuilder::new("^S")
        .bounded_backtracking()
        .only_utf8(false)
        .build()
        .map(|exec| exec.into_byte_regex())
        .map_err(|err| format!("{}", err))
        .unwrap();

    let default_bytes_re = ExecBuilder::new("^S")
        .only_utf8(false)
        .build()
        .map(|exec| exec.into_byte_regex())
        .map_err(|err| format!("{}", err))
        .unwrap();

    let input = vec![83, 83];

[...]

PR #472, opened 2018-04-29
Add random testing of crates.io regex

ethanpailes:
This PR implements the first half of #468.

I've marked it as a WIP until I get to the bottom of the `\b` `find_iter` bug.

The tests this PR adds take a long time to run, so I disabled them by default, but added logic to `ci/script.sh` to run them during a travis.ci cron job. I think someone with access to the travis console will need to enable that (so probably @BurntSushi or @alexcrichton). I'm also totally happy to rip the ci changes out or just leave the tests always enabled.

Commit Msg:

This patch adds some infastructure to scrape crates.io
for regex, then run each of the regex found in this way though
a random testing gauntlet to make sure that all the different
backends behave in the same way. These random tests are
expensive, so we add a feature gate around them so that
normal CI does not get bogged down. If the CI tests are being
run in a cron job, we trigger the random testing.

To make this better we should add something which can generate
a matching string from a regex. As is this will just focus on
the negative case.


Cargo.toml
```diff
@@ -57,6 +57,9 @@ pattern = []
 # Enable to use simd acceleration.
 # Note that this is deprecated and is a no-op.
 simd-accel = []
+# When testing, run the expensive randomized testing suites in addition
+# to the fast unit tests.
+random-test = []
```
BurntSushi: I think we probably don't want a feature for this. Features are part of a crate's public API, and this is really an internal implementation detail of a specific kind of test.

We can instead use normal cfg's, which might be set by env vars. We already do this for enabling simd vector optimizations in the `build.rs`.

scripts/scrape_crates_io.py
```diff
@@ -0,0 +1,182 @@
...
+import os
+import re
+import pdb
+import shutil
+import json
+import tempfile
+import argparse
+import datetime
+import time
+from subprocess import call
```
BurntSushi: Can you sort imports alphabetically? Thanks.

BurntSushi: You probably want to run a pep8 checker on this file. I see some violations.

tests/consistent.rs
```diff
@@ -0,0 +1,213 @@
+extern crate regex;
+extern crate quickcheck;
```
BurntSushi: I generally try to keep `extern crate` statements in the crate root, even if this means repeating some `cfg`s.

tests/consistent.rs
```diff
@@ -0,0 +1,213 @@
...
+       + try!(bytes_checker::check_backends(&bytes_backends)))
+}
+
+//
+// A consistency checker parameterized by the input type (&str or &[u8]).
+//
+
+macro_rules! checker {
+    ($module_name:ident, $regex_type:path, $mk_input:expr) => {
+        mod $module_name {
```
BurntSushi: I'd recommend out-denting the module declaration, and then indenting its contents from there.

tests/test_crates_regex.rs
```diff
@@ -0,0 +1,61 @@
+extern crate regex;
+
+///////////////////////////////////////////////////////////////////////
+//                                                                   //
+//                       The Bug Trophy Case                         //
+//                                                                   //
+///////////////////////////////////////////////////////////////////////
```
BurntSushi: This style of comment isn't anywhere else in the regex crate, and I'm generally not a fan of them. They take up a lot of space, and I'm not really sure what they mean. My guess is that they are just regression tests, and we should probably call them that.

ci/script.sh
```diff
@@ -17,12 +17,18 @@ if [ "$TRAVIS_RUST_VERSION" = "1.12.0" ]; then
...
+if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then
+    CARGO_TEST_ARGS="${CARGO_TEST_ARGS} --features random-test"
+fi
 if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
-  cargo test --verbose --features unstable
-else
-  cargo test --verbose
+    CARGO_TEST_ARGS="${CARGO_TEST_ARGS} --features unstable"
 fi
+cargo test --verbose ${CARGO_TEST_ARGS}
```
BurntSushi: Why did the else block get removed here? I guess it's probably OK, but this will cause the nightly CI test to get quite a bit longer.

tests/consistent.rs
```diff
@@ -0,0 +1,213 @@
...
+impl RegexEqualityTest {
+    fn new(re1: $regex_type, re2: $regex_type) -> Self {
+        RegexEqualityTest {
+            re1: re1,
+            re2: re2,
+        }
+    }
+}
+
+impl quickcheck::Testable for RegexEqualityTest {
```
BurntSushi: This is kind of weird? What made you implement `Testable` explicitly here?

ethanpailes: I ran into some trouble with the borrow checker when I tried to use a closure over the two regex in question. I think the closures tried to take ownership of the regex, when I wanted to copy them instead.

ci/script.sh
ethanpailes: Well my plan was to just build up an arg string in the if statements so that random testing was handled seperatly from the unstable feature checking, then execute a single cargo test will all the args, but now I realize that this would cause two `--features` flags to be passed :(. This is going away as part of replacing the feature with an env var anyway.

ethanpailes:
Thanks for the review!

I tried to address the comments in the patch I just pushed.

This PR is still blocked on the word boundary bug, I'll try to get to that soon.

WRT time it takes about 35 seconds in release mode and 9ish minutes in debug mode. Wow that makes a difference! I think I should probably just ditch the ci shenannigans and enable release mode for the random testing every time.

BurntSushi:
> I think I should probably just ditch the ci shenannigans and enable release mode for the random testing every time.

Sounds good!

> This PR is still blocked on the word boundary bug, I'll try to get to that soon.

If this turns out to be hard (and it may very well be), would it be possible to comment out those tests for now in the interest of landing the PR? I think this is overall great work, and I don't want to put you in a position where you need to fix existing bugs before landing it.

ethanpailes:
> If this turns out to be hard (and it may very well be), would it be possible to comment out those tests for now in the interest of landing the PR? I think this is overall great work, and I don't want to put you in a position where you need to fix existing bugs before landing it.

Definitly! The last bug turned out the be pretty shallow when I actually looked at it, so I'm hoping this will be the same. If it leads to lazy DFA arcana I won't hesitate to bail out.

Do you happen to know if there is an easy way to enable release mode for just one test binary? If there isn't I can probably break out the `build.rs` hackery, but that seems gross to me.

BurntSushi:
Not to my knowledge unfortunately. `build.rs` hackery sounds like the best bet.

ethanpailes:
I had trouble with the word boundary bug, so I commented it out. I've removed the WIP status.

I also had trouble figuring out a build.rs hack because it looks like you are not allowed to specify arbitrary rustc flags. Instead I went back to the environment variable and CI script approach. The difference is that I no longer special case cron jobs. The random tests are run in release mode on every CI run.

ethanpailes:
I rebased this patch on top of master. Github is telling me there are changes requested, but I think the patches those comments were attached to got lost to rebasing. I do remember taking steps to address all of them though. If there is anything else I can do, please let me know!

ci/script.sh
```diff
@@ -9,12 +9,18 @@ cargo build --verbose
 cargo doc --verbose
 
 # Run tests. If we have nightly, then enable our nightly features.
+CARGO_TEST_EXTRA_FLAGS=""
 if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
-  cargo test --verbose --features unstable
-else
-  cargo test --verbose
+  CARGO_TEST_EXTRA_FLAGS="--features unstable"
```
BurntSushi: Incidentally, I think this whole thing can just be removed now. `unstable` is now a no-op. :-)

tests/consistent.rs
```diff
@@ -0,0 +1,213 @@
...
+            try!(ExecBuilder::new(re)
+                    .only_utf8(false)
+                    .build()
+                    .map(|exec| exec.into_byte_regex())
+                    .map_err(|err| format!("{}", err)))),
+    ];
+
+    Ok(try!(string_checker::check_backends(&standard_backends))
+       + try!(string_checker::check_backends(&utf8bytes_backends))
+       + try!(bytes_checker::check_backends(&bytes_backends)))
```
BurntSushi: Probably shouldn't be using `try!` any more. :-) We now require Rust 1.20!

ci/script.sh
ethanpailes: I removed the features but left the logic in place in case it is ever used again.

tests/consistent.rs
ethanpailes: I've switched from `try!` to `?`.

ethanpailes:
On my machine it takes 3:36 of wallclock time (though 17:30 of CPU time). Milage may vary on the different CI servers, but I think that the startup time of AppVeyor will probably still dominate.

BurntSushi:
@ethanpailes Hmm. I'm not getting the answer I'm looking for, so I must be asking the wrong question.

Forget about CI for a moment. Adding a couple minutes to an already 10 minute build is probably something I can live with, and we don't need to make it conditional.

What I care about a lot more is the cycle time during development. There are three different things to test here:

1. Compile and run all tests.
2. Compile all tests, run everything except for the crates.io tests.
3. Do not compile the crates.io tests at all, but compile and run everything else.

This is the kind of thing I do a lot when working on the regex crate, so it's important to keep this cycle time down. I don't know which option to pick though because I don't have the data.

When I have time, I can pull down this PR and figure out the data points if this is still unclear.

ethanpailes:
That makes sense. I'll try to provides some times that outline of various OODA loops with notes on how I personally use each that may or may not be useful.

| Action | Time | Notes |
|----------|--------|----------|
| `cargo clean && cargo test` | 1:49 | This one probably does not matter that much, but it is still worth knowing. `cargo test` alone will not actually run the crates.io tests because the RUST_REGEX_RANDOM_TEST environment variable is not set. |
| After a clean build insert `println!("a change")` into `src/exec.rs:ExecBuilder::new` and `cargo test` | 1:14 | I run this sort of thing when I'm finalizing a patch. Again, will not run the crates.io tests. |
| After a clean build insert `println!("a change")` into `src/exec.rs:ExecBuilder::new` and `cargo test --test default` | 0:15 | This sort of thing is my main driver and probably should not have changed. |
| After a clean build remove `println!("a change")` from `src/exec.rs:ExecBuilder::new` and `export RUST_REGEX_RANDOM_TEST=1 && time cargo test --release --test crates-regex` | 0:48 | This will run the crates.io tests, and is faster than compiling in debug mode. Does not run any other tests. |
| `./ci/script.sh` | 3:36 | This one compiles and runs all tests, including the crates.io tests. I usually do this when finalizing a patch. The best way to compile and run all commands is with two commands (one for the crates.io tests in release mode, and one for the rest in debug mode). |

I don't have a time for (2) from your list, because I don't know of a negative test filter you can pass to `cargo test`, so the only way I can think to do it is to enumerate all the other tests, which seems annoying.

BurntSushi:
> I don't have a time for (2) from your list, because I don't know of a negative test filter you can pass to cargo test, so the only way I can think to do it is to enumerate all the other tests, which seems annoying.

I think it's actually the second entry in your table, no? That is, it compiles all tests, but runs everything except for the crates.io tests.

I think the thing you don't have is how long `cargo test` takes without needing to compile the crates.io tests.

With that said, it's a good point about `cargo test --test default`. That does tend to be the main thing that I run, but I do just run `cargo test` a bit as well.

ethanpailes:
> I think it's actually the second entry in your table, no?

You're right. Derp.

> I do just run cargo test a bit as well

1:15 is something I'd personally be comfortable with for an outer feedback cycle (I might have been warped by 15 minute C++ outer feedback cycles), but it is definitly starting to become annoying. I think ideally we would configure `cargo test` with no extra args not to compile the crates.io tests. I ran out the steam on that last time, but I can take another crack at it if you want.

BurntSushi:
It's fine. Let's just merge this, and if I become annoyed, I'll fix it. :-)

BurntSushi:
Thanks again!

ethanpailes:
Thanks!

Issue #476, opened 2018-05-07
add no_std regex, which depends on alloc


BurntSushi:
There has been some interest in putting out a version of regex that doesn't depend on std itself, but instead depends just on alloc. This is within reach because regex already doesn't rely too much on platform specific details, and mostly just depends on dynamic memory allocation. There are however some parts of the regex API that will need to be tweaked. For example, regex uses `std::io::Error`, which isn't available in `alloc`. This is why regex 1.0 got a `use_std`/`std` feature. Namely, compiling regex _without_ that feature fails today. This will allow us to change the semantics of that compilation mode without breaking backwards compatibility.

@ZackPierce has been diligently adding support for this by starting with regex's dependencies. So far:

* https://github.com/BurntSushi/utf8-ranges/pull/8
* https://github.com/BurntSushi/aho-corasick/pull/28
* https://github.com/BurntSushi/ucd-generate/pull/1

I thought it would be good to track this issue at a higher level so we can discuss a game plan. I'd also like to share some of my thoughts/constraints on the process.

I basically think that we should do this. What I am unsure of is the timeline. For the most part, my own personal maintenance bandwidth is very limited, and to this end, I've generally avoided nightly-only support of things. (I've made some exceptions. Support for things like `Pattern` happened before I knew better, and support for SIMD happened because I am very excited about it and got involved with the SIMD stabilization effort.) Namely, I cannot and will not be beholden to nightly breakages because I simply can't keep up. To that end, I would like to know more about what the plan is for `no_std` environments. Is the `alloc` crate setup generally where we think we're going? If so, and if it remains relatively stable, I think I could get on board with this relatively soon.

It's also worth saying that some changes are simpler than others. For example, making `utf8-ranges` compatible with `no_std` is pretty reasonable, but the `aho-corasick` changes are quite a bit broader. I have concerns over peppering conditional compilation everywhere, and I think those sorts of things are very hard to maintain. I'm hopeful we can find a better way. This gets worse when the public API is impacted. The complexity and maintenance burden goes way up. This is apparently so bad that it was worth adding a new dependency (`cfg-if`) that must be paid for by everyone just to support the `no_std` users. I'm not especially excited about that, particularly if the trend continues.

For the most part, I really wasn't intending on tackling this feature until more stuff stabilized. But I wanted to get my thoughts out there so that there are no surprises.

I welcome other thoughts on the matter!

ZackPierce:
[...]

PR #477, opened 2018-05-07
Add nominal no_std + alloc support to regex-syntax

ZackPierce:
Introduces feature flags and associated conditional
compilation configuration to alloc use in no_std
builds for environments with access to a heap.

Note that until ucd-util is updated and the relevant
feature flags piped through, this mode will not
be practically usable.

The main discussion of this feature is likely to occur on the primary issue tracking the general `no_std` with heap feature for regex: https://github.com/rust-lang/regex/issues/476

### DIFF
--- a/regex-syntax/Cargo.toml
+++ b/regex-syntax/Cargo.toml
@@ -17,5 +17,4 @@ std = []
 alloc = []
 
 [dependencies]
-cfg-if = "0.1"
 ucd-util = "0.1.0"
--- a/regex-syntax/Cargo.toml
+++ b/regex-syntax/Cargo.toml
@@ -9,5 +9,13 @@ homepage = "https://github.com/rust-lang/regex"
 description = "A regular expression parser."
 workspace = ".."
 
+[features]
+default = ["std"]
+# Disable this on-by-default feature and add "alloc" to allow use in no_std builds
+std = []
+# Required for use in no_std builds, presently nightly-only
+alloc = []
+
 [dependencies]
+cfg-if = "0.1"
 ucd-util = "0.1.0"
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -112,9 +112,9 @@ done automatically in the `regex` crate.
 #[cfg(test)]
 extern crate std as std_test;
 
-#[cfg(feature = "std")]
+#[cfg(not(feature = "std"))]
 #[macro_use]
-extern crate std as core;
+extern crate core as std;
 
 #[cfg(all(feature = "alloc", not(feature = "std")))]
 #[macro_use]
@@ -130,11 +130,11 @@ mod either;
 mod error;
 pub mod hir;
 mod parser;
+mod prelude;
 mod unicode;
 mod unicode_tables;
 
-#[cfg(all(feature = "alloc", not(feature = "std")))]
-use alloc::string::String;
+use prelude::String;
 
 /// Escapes all regular expression meta characters in `text`.
 ///
@@ -186,7 +186,7 @@ pub fn is_meta_character(c: char) -> bool {
 /// `Join_Control` properties, or is in one of the `Decimal_Number`, `Mark`
 /// or `Connector_Punctuation` general categories.
 pub fn is_word_character(c: char) -> bool {
-    use core::cmp::Ordering;
+    use std::cmp::Ordering;
     use unicode_tables::perl_word::PERL_WORD;
 
     if c <= 0x7F as char && is_word_byte(c as u8) {
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -109,18 +109,18 @@ done automatically in the `regex` crate.
 #![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(alloc))]
 #![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(slice_concat_ext))]
 
-#[cfg(all(test, not(feature = "std")))]
+#[cfg(test)]
+extern crate std as std_test;
+
+#[cfg(feature = "std")]
 #[macro_use]
-extern crate std;
+extern crate std as core;
 
 #[cfg(all(feature = "alloc", not(feature = "std")))]
 #[macro_use]
 extern crate alloc;
 
 extern crate ucd_util;
-
-#[macro_use]
-extern crate cfg_if;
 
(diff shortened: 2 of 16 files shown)


regex-syntax/Cargo.toml
```diff
@@ -9,5 +9,12 @@ homepage = "https://github.com/rust-lang/regex"
 description = "A regular expression parser."
 workspace = ".."
 
+[features]
+default = ["std"]
+# Disable this on-by-default feature and add "alloc" to allow use in no_std builds
+std = []
+# Required for use in no_std builds, presently nightly-only
+alloc = []
```
dhardy: I don't think you need an `alloc` feature flag if it will always be required — i.e. unless there are plans for core-only support in the future you could just drop this feature flag and do `#![cfg_attr(not(feature = "std"), feature(alloc))]`.

regex-syntax/src/lib.rs
```diff
@@ -105,6 +105,21 @@ done automatically in the `regex` crate.
 
 #![deny(missing_docs)]
 
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(alloc))]
+#![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(slice_concat_ext))]
+
+#[cfg(test)]
+extern crate std as std_test;
```
dhardy: Why do this? If `std` is required for testing, you can just use regular includes from `std`. If you want to test where `std` is not available this will fail. It's possible to run many tests without `std` anyway, though might take some more work.

regex-syntax/src/lib.rs
```diff
@@ -105,6 +105,21 @@ done automatically in the `regex` crate.
...
+
+#[cfg(feature = "std")]
+#[macro_use]
+extern crate std as core;
```
dhardy: In retrospect, I would recommend not doing what we did in Rand but instead having `extern crate core as std;` (for non-std). The reasons being (a) it saves changing so many imports and (b) `std` may use feature flags in the future.

However, I suppose error messages may be fun in no-std mode :smile:

regex-syntax/src/lib.rs
```diff
@@ -118,6 +133,9 @@ mod parser;
 mod unicode;
 mod unicode_tables;
 
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::string::String;
```
dhardy: It's a shame `alloc` doesn't have a prelude. Can I suggest making your own (containing the usual members only `std` isn't used) then doing `use ::prelude::*;` in all modules?

ZackPierce: Good idea! I'll see how this shakes out.

Centril:
@ZackPierce so one change I'm making to the changes you made in proptest in a PR, https://github.com/Centril/proptest/blob/general-refactor/src/std_facade.rs, is to have one file for all the conditional std/alloc uses, which reexports all the needed paths. Then you just import `use std_facade::Vec`. This cut down on complexity somewhat.

ZackPierce:
Excellent, thank you for adding that!

regex-syntax/Cargo.toml
ZackPierce: I'd like to leave the door open for that possibility. We can simplify the cfg statements if those plans don't materialize.

regex-syntax/src/lib.rs
ZackPierce: The idea here is that we want to be able to run our tests against the library compiled in either mode as a way of confirming that there is no accidental behavioral differences.  

E.G. `cargo test --no-default-features --features "alloc"`

IIUIC, because we (now) do `extern crate core as std` in the main library when in no_std mode (to reduce code diffs), we need to import `std` under some other name for use in the tests so as not to conflict with the core-renamed-as-std crate being used by the library.

regex-syntax/src/lib.rs
ZackPierce: I adopted this recommendation in the latest commit to this PR.

ZackPierce:
I have updated this PR based on suggestions from @dhardy and a helpful example from @Centril . Pretty much all differences between the `no_std` + `alloc` and `std` versions of the build have been centralized and hidden in the new prelude module.

The end result is that there are about *half* as many code changes to the main, functional portions of the crate.

regex-syntax/src/lib.rs
dhardy: I don't even know that you do need `std` for most of the testing. Have a look at the Rand lib code; we allow many tests to run with `cargo test --no-default-features` without re-importing `std`. But it's possible there are other parts of `std` you do require.

regex-syntax/Cargo.toml
dhardy: Trick we do with Rand: make `std` depend on `alloc`. This has the advantage that you can use just `#[cfg(feature = "alloc")]` to feature-gate modules/items requiring an allocator (though also the disadvantage that sometimes you need to check both: `#![cfg_attr(all(feature="alloc", not(feature="std")), feature(alloc))]`).

regex-syntax/src/lib.rs
ZackPierce: Yes, I agree that we *could* make a decent chunk of the tests run without access to the standard library. 

That said, the main goal is testing the library code. AFAICT that library code can be compiled in no_std mode and exercised regardless of whether the test code itself has access to standard library.  As as we're getting the library code exercised already, I'm not clear on what value add there would be to go to the effort of triaging and modifying the test code to be able to run without the standard lib.

Centril:
Relevant: https://github.com/rust-lang/rfcs/pull/2480

BurntSushi:
I'm going to close this out for now. My plan is add support for this at some point. The main thing standing in the way right now is probably the work itself, the MSRV and the new features introduced by #613. Managing the MSRV is probably find, since this will mean enabling compilation where it previously didn't work before, so long as nothing else requires an MSRV bump. (And if it does, we could add back a version sniffing `build.rs` script, but it would be nice to avoid that complexity.)

And #613 poses an issue because I think Cargo has problems with enabling features in optional dependencies. e.g., if we had a `alloc = ["aho-corasick/alloc"]` feature, then that will unconditionally bring in the `aho-corasick` dependency when `alloc` is enabled, even if it was otherwise disabled. I think this can be worked around by introducing more features, but that's not a great user experience.

PR #479, opened 2018-05-08
Update crate version in docs example snippet.

budziq:
There is an old crate version in Cargo.toml snippet in lib.rs


BurntSushi:
Ah thanks! Looks like #478 arrived shortly before this one. :-)

PR #480, opened 2018-05-09
Fix regression in travis coverage

ethanpailes:
A recent change to the travis-ci `sudo: false` environment
broke kcov. The temporary solution that travis has recomended
is to switch to the `sudo: required` environment. The tradeoff
here is that the build will take longer to start, so
it might also be reasonable to just stop collecting coverage
info until travis finishes `sudo: false`.

See: https://github.com/travis-ci/travis-ci/issues/9061

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 dist: trusty
-sudo: false
+sudo: required
 language: rust
 rust:
 - 1.20.0


BurntSushi:
Wow, thanks so much for fixing this! I remember seeing the error and being befuddled by it, but never dove in.

The increase in build times is annoying, but I'm willing to at least try it if it means getting code coverage back. If it becomes too annoying, we can disable code coverage for now.

ethanpailes:
The `after_success.sh` script does not seem to have generated any output. I'm going to try to debug the issue by pushing more commits to this branch, but they should just be regarded as noise until I remove the `[WIP]` tag and squash.

BurntSushi:
I don't think code coverage runs in PR CI.

ethanpailes:
Oh, good point. I'll unmark it as a WIP and if there are more problems they can be fixed in a seperate PR.

PR #486, opened 2018-06-07
regex: implement size_hint for set match iterators

srijs:
Implementing `size_hint` for `SetMatchesIntoIter` and `SetMatchesIter` by just passing through the hint from the underlying iterators.

I've refrained from implementing `ExactSizeIterator` for these two types because I wasn't sure whether you'd be okay committing to this API.

### DIFF
--- a/src/re_set.rs
+++ b/src/re_set.rs
@@ -295,6 +295,10 @@ impl Iterator for SetMatchesIntoIter {
             }
         }
     }
+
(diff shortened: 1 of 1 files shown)


BurntSushi:
> I've refrained from implementing ExactSizeIterator for these two types because I wasn't sure whether you'd be okay committing to this API.

I think this would be OK! :-)

PR #488, opened 2018-06-11
Regenerate tables for Unicode 11

cuviper:
This adds `scripts/generate.py`, and uses it to regenerate all tables
with data from Unicode 11.0.0.  This also restores the character tests
that were first added in #400, with a new one for 11.


regex-syntax/src/lib.rs
```diff
@@ -213,8 +213,15 @@ mod tests {
...
-        assert!(is_word_character('β'));
+        assert!(is_word_character('a'), "ASCII");
+        assert!(is_word_character('à'), "Latin-1");
+        assert!(is_word_character('β'), "Greek");
+        assert!(is_word_character('\u{11011}'), "Brahmi (Unicode 6.0)");
+        assert!(is_word_character('\u{11611}'), "Modi (Unicode 7.0)");
+        assert!(is_word_character('\u{11711}'), "Ahom (Unicode 8.0)");
+        assert!(is_word_character('\u{17828}'), "Tangut (Unicode 9.0)");
+        assert!(is_word_character('\u{1B1B1}'), "Nushu (Unicode 10.0)");
+        assert!(is_word_character('\u{16E40}'), "Medefaidrin (Unicode 11.0)");
```
BurntSushi: Nice, thank you. :-)

BurntSushi:
This PR is in `regex-syntax 0.6.1` on crates.io. Thanks again!

PR #493, opened 2018-06-25
regex: expose lower level search APIs

### LINKED ISSUE #219: Regex::captures forces an allocation on every call
Every time one calls `captures`, a new allocation for storing the location of captures is created. This allocation has size proportional to the number of captures in the regex.

This is also true for `captures_iter`, where every iteration results in a new allocation. An iterator could reuse the allocation in theory, but ownership of the captures is transferred to the caller. Even if we could reuse the capture locations, we couldn't give the caller a mutable borrow, since that immediately puts us in the "streaming iterator" conundrum.

The most sensible API I can think of is to:
1. Permit the caller to build empty `Captures` values from a given `Regex` such that it has the right size.
2. Pass a mutable borrow to a `Captures` to a call to `captures`, which lets the caller control the allocation.

It's not quite clear how to apply this to `captures_iter` while still implementing `Iterator`. I suspect we should probably borrow from the `io::BufReader::read_line` style methods. e.g.,

``` rust
impl Regex {
    // Returns empty storage for captures for use with read_captures.
    fn new_captures(&self) -> Captures { ... }

    // On successful match, returns true and sets capture locations.
    // Otherwise returns false.
    fn read_captures(&self, caps: &mut Captures, text: &str) -> bool { ... }
    fn read_captures_iter<'r, 't>(&'r self, text: &'t str) -> ReadCapturesIter<'r, 't> { ... }
}

struct ReadCapturesIter<'r, 't> { ... }

impl<'r, 't> ReadCapturesIter<'r, 't> {
    // On successful match, returns true and sets capture locations.
    // Otherwise returns false.
    fn captures(&mut self, caps: &mut Captures) -> bool { ... }
}
```

And I think this would work well.

Main questions:
1. Any alternatives?
2. Do we _replace_ the existing API with the one from above (in 1.0)? Or do we add it? My inclination is to add it, but I really hate expanding the API with more choices.

BurntSushi:
This commit exposes two new areas of API surface:

  1. A new `captures_read` method which provides a way to access the
     offsets of submatches while amortizing the allocation of the
     space required to store those offsets. Callers should still of
     course prefer to use the higher level `captures` method, but if
     performance dictates, this lower level API may be useful.
  2. New "at" variants of
     shortest_match/is_match/find/captures/captures_read that permit
     controlling where the start of a search begins within a slice.
     This is typically useful for controlling the match semantics of
     look-around operators such as `^` and `$`, and are necessary for
     implementing non-overlapping iterators.

Fixes #219

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -30,9 +30,7 @@ cargo test --verbose --manifest-path regex-syntax/Cargo.toml
 cargo doc --verbose --manifest-path regex-syntax/Cargo.toml
 
 # Run tests on regex-capi crate.
-cargo build --verbose --manifest-path regex-capi/Cargo.toml
-(cd regex-capi/ctest && ./compile && LD_LIBRARY_PATH=../../target/debug ./test)
-(cd regex-capi/examples && ./compile && LD_LIBRARY_PATH=../../target/debug ./iter)
+ci/test-regex-capi
 
 # Make sure benchmarks compile. Don't run them though because they take a
 # very long time. Also, check that we can build the regex-debug tool.
--- a/ci/test-regex-capi
+++ b/ci/test-regex-capi
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+set -e
+
+cargo build --verbose --manifest-path regex-capi/Cargo.toml
+(cd regex-capi/ctest && ./compile && LD_LIBRARY_PATH=../../target/debug ./test)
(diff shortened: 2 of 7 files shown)


BurntSushi:
cc @raphlinus I think you mentioned looking for these APIs at one point in the long ago. These have been available as undocumented APIs for quite a while, but now I actually need them for another project and I think they've mostly served me well.

BurntSushi:
I've been using these new APIs for a bit now and everything feels OK to me. So I'm going to bring this in and push out a new release.

PR #494, opened 2018-06-26
regex 1.0.0 -> regex 1.0.1

bcko:


BurntSushi:
Thanks for this! Although, I'm not sure how much it really matters having the precise version in the README. To avoid doing this dance in the future, could you just change this to `regex = "1"`?

bcko:
Thank you for your help! @BurntSushi

PR #504, opened 2018-08-16
update some debs mostly for minimal-versions

Eh2406:
This bumps the minimal acceptable versions in Cargo.toml to versions that are compatible with `-Z minimal-versions`. This is part of the process of seeing how hard this is for crates to use in preparation for getting it stabilized for use in CI, specifically upstreaming the changes required to get criterion working with it. It is easy to use if all of your dependencies support it, but much harder if trying to impose it on them.

I am not sure I have this exactly correct. There may be more things that need to be updated for the benchmarking code to work with minimal-versions. I was not able to test with all the feachers. But I have verified that `docopt` can be made to work with this change to `regex`. So it is a step in the correct direction.

### DIFF
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -12,12 +12,12 @@ build = "build.rs"
 workspace = ".."
 
 [dependencies]
-docopt = "0.8"
+docopt = "1"
 lazy_static = "1"
 libc = "0.2"
 onig = { version = "3", optional = true }
 libpcre-sys = { version = "0.2", optional = true }
-memmap = "0.6"
+memmap = "0.6.2"
 regex = { version = "1", path = ".." }
 regex-syntax = { version = "0.6.0", path = "../regex-syntax" }
 serde = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ aho-corasick = "0.6.0"
 # For skipping along search text quickly when a leading byte is known.
 memchr = "2.0.0"
 # For managing regex caches quickly across multiple threads.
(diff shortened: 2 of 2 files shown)


Eh2406:
cc @BurntSushi follow up to our discussion at https://github.com/Amanieu/thread_local-rs/pull/13

bench/Cargo.toml
```diff
@@ -12,12 +12,12 @@ build = "build.rs"
...
 
 [dependencies]
-docopt = "0.8"
+docopt = "1"
 lazy_static = "1"
 libc = "0.2"
 onig = { version = "3", optional = true }
 libpcre-sys = { version = "0.2", optional = true }
-memmap = "0.6"
+memmap = "0.6.2"
```
BurntSushi: Oh this one is interesting. What requires `0.6.2` specifically?

Eh2406: [6.1](https://crates.io/crates/memmap/0.6.1) requires `winapi ^0.2` and `<=0.2.5` do not build on rust 1.0.

Eh2406:
Also to be clear this is enough for `cargo check --tests` to pass in the bench subfolder, the main folder, and a copy docopt pointing at this on my windows machine. Running test, and running on other machines may find more.

BurntSushi:
Thanks!

Eh2406:
To be a little more blunt, is there an ETA for a release that has this merged? Is there follow up that could hasen it?

BurntSushi:
I don't really do scheduled releases, but I just put `regex 1.0.3` on crates.io, which has this PR.

PR #507, opened 2018-08-25
deps: update to quickcheck 0.7

BurntSushi:
We also add a test to CI check for minimal versions.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -8,6 +8,15 @@ set -ex
 cargo build --verbose
 cargo doc --verbose
 
+# If we're testing on an older version of Rust, then only check that we
+# can build the crate. This is because the dev dependencies might be updated
+# more frequently, and therefore might require a newer version of Rust.
+#
+# This isn't ideal. It's a compromise.
+if [ "$TRAVIS_RUST_VERSION" = "1.20.0" ]; then
+  exit
+fi
+
 # Run tests. If we have nightly, then enable our nightly features.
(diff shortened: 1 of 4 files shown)


ci/script.sh
```diff
@@ -39,4 +48,9 @@ if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
   for x in rust rust-bytes pcre1 onig; do
     (cd bench && ./run $x --no-run --verbose)
   done
+
+  # Test minimal versions.
+  cargo +nightly generate-lockfile -Z minimal-versions
+  cargo build --verbose
+  cargo test --verbose
```
BurntSushi: cc @Eh2406

ci/script.sh
```diff
@@ -8,6 +8,15 @@ set -ex
 cargo build --verbose
 cargo doc --verbose
 
+# If we're testing on an older version of Rust, then only check that we
+# can build the crate. This is because the dev dependencies might be updated
+# more frequently, and therefore might require a newer version of Rust.
+#
+# This isn't ideal. It's a compromise.
```
Eh2406: An alternative compromise that has been suggested is to test MRV with `minimal-versions` so that it does not use the updated dependencies. This involves using a nightly to build the lockfile then using the MRV to build the result. It is not a silver bullet, but it is another option.

BurntSushi: Neat idea, but I think I'll pass unless something comes up that motivates it. :-)

PR #520, opened 2018-10-07
exclude useless files from crates.io

ignatenkobrain:


BurntSushi:
This PR is on crates.io in `regex 1.0.6`.

PR #528, opened 2018-10-19
Add index trait for regexset

hellow554:
It is useful to access the regex one provided once to the
regexset(-builder).
Instead to let the user store the regex in a seperate list to
index them later, one can now index the regexset directly.

BurntSushi mentioned that he would happily accept a PR (https://users.rust-lang.org/t/finding-set-of-substrings-in-a-long-string/21416/7) so here it is. 

This is my first contribution to the rust-lang, I'm excited :heart:

### DIFF
--- a/src/re_set.rs
+++ b/src/re_set.rs
@@ -14,6 +14,7 @@ macro_rules! define_set {
         pub mod $name {
             use std::fmt;
             use std::iter;
+            use std::ops;
             use std::slice;
             use std::vec;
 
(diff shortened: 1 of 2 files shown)


BurntSushi:
@hellow554 Thanks so much! I think your most recent comment on the forums is right though: I don't think we should define an `Index` impl on `RegexSet`. Instead, I'd propose just adding a method called `patterns` that returns `&[String]`. Otherwise though, I think your implementation looks good!

hellow554:
`&[String]` or `&[&str]`?

BurntSushi:
@hellow554 `&[&str]` isn't possible. :-) Namely, `String` and `&str` do not have the same in memory representation, so you cannot cheaply get a `&[&str]` from a `Vec<String>`.

hellow554:
After thinking about that, you're right. Any thoughts about the method name? Something like `get_set(&self)` or `get_expressions(&self)`?

BurntSushi:
I think just `patterns` is fine.

PR #529, opened 2018-10-19
Add ability get receive pattern from a RegexSet

hellow554:
It is useful to access the regex one provided once to the regexset(-builder).
Instead to let the user store the regex in a seperate list to index them later, one can now receive the patterns from regexset directly.

This is a followup to https://github.com/rust-lang/regex/pull/528#issuecomment-431342324 which not uses the index-trait but instead provides a seperate function `patterns` for receiving the stored patterns.

### DIFF
--- a/src/re_set.rs
+++ b/src/re_set.rs
@@ -219,9 +219,10 @@ impl RegexSet {
 
     /// Returns the patterns that this set will match on.
     ///
-    /// This function can be used to determine the match without storing it
-    /// in a seperate `Slice`. This is useful when you want to know the pattern
-    /// that matched.
+    /// This function can be used to determine the pattern for a match. The 
+    /// slice returned has exactly as many patterns givens to this regex set, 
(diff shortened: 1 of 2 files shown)


src/re_set.rs
```diff
@@ -216,6 +216,36 @@ impl RegexSet {
     pub fn len(&self) -> usize {
         self.0.regex_strings().len()
     }
+
+    /// Returns the patterns that this set will match on.
+    ///
+    /// This function can be used to determine the match without storing it
+    /// in a seperate `Slice`. This is useful when you want to know the pattern
+    /// that matched.
```
BurntSushi: Hmm, I feel like this phrasing is a bit off. Could you try this instead?

> This function can be used to determine the pattern for a match. The slice returned has exactly as many patterns givens to this regex set, and the order of the slice is the same as the order of the patterns provided to the set.

Thanks!

hellow554:
done

PR #534, opened 2018-10-27
Ensure `[[:blank:]]` only matches `[ \t]`. Fix #533.

### LINKED ISSUE #533: `[[:blank:]]` does not just match `[ \t]`
Both the docs of [`regex 1.0.5`](https://docs.rs/regex/1.0.5/regex/#ascii-character-classes) and [`regex-syntax 0.6.2`](https://docs.rs/regex-syntax/0.6.2/regex_syntax/ast/enum.ClassAsciiKind.html#variant.Blank) suggests that `[[:blank:]]` is equivalent to `[ \t]`, but it is actually wrong as shown in the test program below:

```rust
extern crate regex; // 1.0.5

fn main() {
    let r = regex::Regex::new("[[:blank:]]").unwrap();
    assert!(!r.is_match("\u{17}")); // unexpectedly panicked
}
```

I believe the error is due to 

https://github.com/rust-lang/regex/blob/488fe569575ea06e1d8c5b9f98f95d9644e1efa8/regex-syntax/src/hir/translate.rs#L1042-L1045

line 1043 should read

```rust
 const X: T = &[('\t', '\t'), (' ', ' ')];
```

kennytm:


regex-syntax/src/hir/translate.rs
```diff
@@ -1040,7 +1040,7 @@ fn ascii_class(kind: &ast::ClassAsciiKind) -> &'static [(char, char)] {
             X
         }
         Blank => {
-            const X: T = &[(' ', '\t')];
+            const X: T = &[('\t', '\t'), (' ', ' ')];
```
BurntSushi: Oof. Thanks for catching this! It looks like a bad transcription error. However, I don't think this is right either. The docs include more characters:

```
[[:space:]]    whitespace ([\t\n\v\f\r ])
```

And this is consistent with how `regex-syntax 0.4` and below behaved:

https://github.com/rust-lang/regex/blob/fa5cf6b86e38d010de816ad4e2a096a8c62e7640/regex-syntax/src/parser.rs#L1453-L1454

RReverser: @BurntSushi You're referring to docs for `space`, but this bug and fix are for `blank` which has fewer characters.

BurntSushi: Ug. Thanks for that. I don't think I had quite woken up yet!

PR #535, opened 2018-11-07
license: add the unicode license for unicode-tables

### LINKED ISSUE #530: Add unicode license for the unicode tables in regex-syntax
We noticed that regex-syntax uses tables derived from unicode 11, but the license files don't include a reference to the [unicode terms of use](http://www.unicode.org/copyright.html). Could this be added? Thanks so much!

dgonyeo:
Add the Unicode license to the unicode-tables directory, as the data
there comes from the Unicode Consortium.

cc @BurntSushi 

Fixes #530

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,3 +232,7 @@ This project is licensed under either of
    http://opensource.org/licenses/MIT)
 
 at your option.
+
+The data in `regex-syntax/src/unicode_tables/` is licensed under the Unicode
+License Agreement
+([LICENSE-UNICODE](http://www.unicode.org/copyright.html#License).
--- a/regex-syntax/src/unicode_tables/LICENSE-UNICODE
+++ b/regex-syntax/src/unicode_tables/LICENSE-UNICODE
@@ -1,0 +1,57 @@
+UNICODE, INC. LICENSE AGREEMENT - DATA FILES AND SOFTWARE
+
+Unicode Data Files include all data files under the directories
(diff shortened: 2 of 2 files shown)


dgonyeo:
Oh, I should also update the Cargo.toml. Let me update this before it merges.

regex-syntax/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "regex-syntax"
 version = "0.6.2"  #:version
 authors = ["The Rust Project Developers"]
-license = "MIT/Apache-2.0"
+license = "MIT/Apache-2.0/Unicode"
```
BurntSushi: I'm not sure that's entirely appropriate, since it isn't really referring to the license of the crate as a whole. With your change, it makes it look like `regex-syntax` is triply licensed, which I don't think is accurate. Specifically, I think

> this copyright and permission notice appear in associated Documentation

is satisfied by putting the Unicode license in this repo and adding a note to the README.

BurntSushi:
Thanks! I imagine other crates in the Rust ecosystem may want to do the same. cc @SimonSapin @Manishearth

regex-syntax/Cargo.toml
dgonyeo: Yeah I suppose that's true. I'll remove it.

BurntSushi:
Thanks again! This change is now on crates.io in `regex-syntax 0.6.3`.

PR #537, opened 2018-11-19
[RFC] regex-capi: expose regex::escape

marmeladema:
I needed regex::escape available in the C API so i added it, but i have only done a few hours of rust in my life so this patch might be totally wrong. I have successfully been able to rust-regex-escape some strings from Python using ctypes to load the library though.

I followed this guide to handle String as returned type: http://jakegoulding.com/rust-ffi-omnibus/string_return/

### DIFF
--- a/regex-capi/include/rure.h
+++ b/regex-capi/include/rure.h
@@ -567,6 +567,38 @@ void rure_error_free(rure_error *err);
  */
 const char *rure_error_message(rure_error *err);
 
+/*
+ * rure_escape_must returns a NULL terminated string where all meta characters
+ * have been escaped. If escaping fails for any reason, an error message is
+ * printed to stderr and the process is aborted.
+ *
+ * The pattern given should be in UTF-8. For convenience, this accepts a C
+ * string, which means the pattern cannot contain NULL byte.
+ */
+const char *rure_escape_must(const char *pattern);
+
+/*
+ * rure_escape returns a NULL terminated string where all meta characters have
+ * been escaped. The pattern must be valid UTF-8 and the length corresponds to
+ * the number of bytes in the pattern.
+ *
+ * error is set if there was a problem compiling the pattern (including if the
+ * pattern is not valid UTF-8). If error is NULL, then no error information
+ * is returned. In all cases, if an error occurs, NULL is returned.
+ *
+ * The pointer returned must not be freed directly. Instead, it should be freed
+ * by calling rure_string_free.
+ */
+const char *rure_escape(const uint8_t *pattern, size_t length,
+                        rure_error *error);
+
+/*
+ * rure_string_free frees the string given.
+ *
+ * This must be called at most once per string.
+ */
+void rure_string_free(char *s);
+
 #ifdef __cplusplus
 }
 #endif
--- a/regex-capi/src/rure.rs
+++ b/regex-capi/src/rure.rs
@@ -570,3 +570,62 @@ ffi_fn! {
         unsafe { (*re).len() }
     }
 }
+
+ffi_fn! {
+    fn rure_escape_must(pattern: *const c_char) -> *const c_char {
+        let len = unsafe { CStr::from_ptr(pattern).to_bytes().len() };
+        let pat = pattern as *const u8;
+        let mut err = Error::new(ErrorKind::None);
+        let esc = rure_escape(pat, len, &mut err);
+        if err.is_err() {
+            let _ = writeln!(&mut io::stderr(), "{}", err);
+            let _ = writeln!(
+                &mut io::stderr(), "aborting from rure_escape_must");
+            unsafe { abort() }
+        }
+        esc
+    }
+}
+
+ffi_fn! {
+    fn rure_escape(
+        pattern: *const u8,
+        length: size_t,
+        error: *mut Error
+    ) -> *const c_char {
+        let pat : &[u8] = unsafe{ slice::from_raw_parts(pattern, length) };
+        let str_pat = match str::from_utf8(pat) {
+            Ok(val)  => val,
+            Err(err) => {
+                unsafe {
+                    if !error.is_null() {
+                        *error = Error::new(ErrorKind::Str(err));
+                    }
+                    return ptr::null();
+                }
+            }
+        };
+        let esc_pat = regex::escape(str_pat);
+        let c_esc_pat = match CString::new(esc_pat) {
+            Ok(val)  => val,
+            Err(err) => {
+                unsafe {
+                    if !error.is_null() {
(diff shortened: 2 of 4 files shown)


regex-capi/include/rure.h
```diff
@@ -567,6 +567,22 @@ void rure_error_free(rure_error *err);
...
 const char *rure_error_message(rure_error *err);
 
+/*
+ * rure_escape returns a NULL terminated string where all meta characters have
+ * been escaped.
+ *
+ * The pointer returned must not be freed directly. Instead, it should be freed
+ * by calling rure_string_free.
+ */
+const char *rure_escape(const char *pattern, size_t length);
```
BurntSushi: Hmm I don't think this signature makes sense to me. It's taking a `char *` and _also_ a length. Usually a `char *` is a NUL terminated string, which means the length isn't necessary. If you want to use a length with the string, then you probably want `uint8_t *` instead. However, I think regex patterns tend to be NUL terminated string literals, so I'd just drop the length parameter here.

Also, patterns are required to be UTF-8, so this needs to be able to return an error.

regex-capi/src/rure.rs
```diff
@@ -570,3 +570,22 @@ ffi_fn! {
         unsafe { (*re).len() }
     }
 }
+
+ffi_fn! {
+    fn rure_escape(pattern: *const c_char, length: size_t) -> *const c_char {
+        let pat : &[u8] = unsafe{ slice::from_raw_parts(pattern as *const u8, length) };
+        let str_pat = str::from_utf8(pat).unwrap();
+        let esc_pat = regex::escape(str_pat);
+        let c_esc_pat = CString::new(esc_pat).unwrap();
```
BurntSushi: I don't think _either_ of these unwraps are acceptable here unfortunately. Both are errors that need to be handled because they can be triggered by user input.

BurntSushi: I think we should use `rure_error`.

BurntSushi: If you like, you can add an `rure_escape_must` that will abort the program on error, similar to `rure_compile_must`. If we do that, you might consider copying the compilation API and using `uint8_t *` for `rure_escape` and `char *` for `rure_escape_must`.

marmeladema:
Sorry to use this PR as a way to learn more about rust but i don't quite understand something.
Before using rure_error, i wanted just to return NULL instead of unwrap so i modified the actual code:
```
ffi_fn! {
    fn rure_escape(pattern: *const u8, length: size_t) -> *const c_char {
        let pat : &[u8] = unsafe{ slice::from_raw_parts(pattern, length) };
        let str_pat = str::from_utf8(pat).unwrap();
        let esc_pat = regex::escape(str_pat);
        let c_esc_pat = CString::new(esc_pat).unwrap();
        c_esc_pat.into_raw()
    }
}
```
to
```
ffi_fn! {
    fn rure_escape(pattern: *const u8, length: size_t) -> *const c_char {
        let pat : &[u8] = unsafe{ slice::from_raw_parts(pattern, length) };
        let str_pat = match str::from_utf8(pat) {
            Ok(val)  => val,
            Err(_) => return std::ptr::null(),
        };
        let esc_pat = regex::escape(str_pat);
        let c_esc_pat = match CString::new(esc_pat) {
            Ok(val)  => val,
            Err(_) => return std::ptr::null(),
        };
        c_esc_pat.into_raw()
    }
}
```
but i get an error on the last line saying that
```
   Compiling rure v0.2.0 (/home/adema/code/regex/regex-capi)                                                                                                                                                                                                   
error[E0308]: mismatched types                                                                                                                                                                                                                                 
   --> regex-capi/src/rure.rs:586:9                                                                                                                                                                                                                            
    |                                                                                                                                                                                                                                                          
586 |         c_esc_pat.into_raw()                                                                                                                                                                                                                             
    |         ^^^^^^^^^^^^^^^^^^^^ types differ in mutability                                                                                                                                                                                                  
    |                                                                                                                                                                                                                                                          
    = note: expected type `*const _`                                                                                                                                                                                                                           
               found type `*mut i8`                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                               
error: aborting due to previous error                                                                                                                                                                                                                          
```                                                                                                                                                                                                                 

Why suddenly the compiler is not able to infer the correct type ?
I had to force it with
`c_esc_pat.into_raw() as *const c_char`

BurntSushi:
Good question. I don't know off the top of my head. It might just be an inference failure.

BurntSushi:
@marmeladema I can't seem to reproduce your problem: https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=964bf197a31f7342fcf79df66163a45d

Which version of Rust are you using?

marmeladema:
rustc 1.30.0 (da5f414c2 2018-10-24)
cargo 1.30.0 (36d96825d 2018-10-24)

BurntSushi:
Dunno then. Can you reproduce it on the playground? If not, it might be the [`ffi_fn`](https://github.com/rust-lang/regex/blob/master/regex-capi/src/macros.rs) macro which does other stuff with the function body, including using it in places where coercion might not apply. But I don't know. I don't have time to dig into the details of automatic coercions right now. :-)

marmeladema:
PR updated with `rure_escape` and `rure_escape_must` and a test.
I had to introduce another variant of `ErrorKind` for `std::ffi::NulError`

marmeladema:
I have an updated version working with prototype
`const uint8_t *rure_escape(const uint8_t *pattern, size_t length,
                        size_t *escaped_len, rure_error *error);`
but before i amend my commit, i just want to understand fully the edge case here.
The only case where `regex::escape` would return a string containing a null byte is when it is already present in the input pattern ? Because i don't see any way of escaping something that would become a null byte.
If i am right, it means that:

1. `str::from_utf8` accepts bytes slice with null byte(s) inside
2. `regex::escape` will not escapes input null bytes

I can understand 1., but 2. sounds weird to me for an escaping function. I understand its not necessary to escape null byte because String handle it properly, but it make a lot of things hard to deal with, like printing etc.

marmeladema:
@BurntSushi any news about this?

BurntSushi:
I merged this in #567, after doing a bit of cleanup. I removed the `rure_escape` function and just kept `rure_escape_must`.

I'd be happy to accept another PR if you want to implement `rure_escape`, subject to my comments above. It should specifically permit NUL bytes without escaping them, but return an error on invalid UTF-8. The fact that interior NULs are difficult to deal with in C is not relevant, and it's why this function must use a `const uint8_t *` instead of a `const char *`.

Thanks for this!

PR #540, opened 2018-12-01
perf: improve dfa matching (with tweaks)

BurntSushi:
This PR is a riff on @Marwes' excellent work in #511. For the most part, this only contains tweaks in order to get it into a mergeable PR:

* We move forward with the MSRV a bit more aggressively, and bump to 1.24.1. We justify this by virtue of the fact that `lazy_static` is now also at 1.24.1, and is a transitive dependency of regex (via `thread_local`).
* A _large_ part of the diff consisted of irrelevant style changes. This made the code harder to review and was generally a huge pain to undo. This PR adds a rustfmt config to this repo that explicitly disables rustfmt in an attempt to prevent this from happening in the future.
* There were several pieces that were undocumented, and the state map's API has been improved and lifted out of a submodule.
* The optimization that avoided cloning a state by reading from the inst ptr scratch space was removed. I'm not in principle against this, but the way it was done violated abstraction too much for my taste.
* Tweaked the heap size calculation. I think it was double-counting the `Arc`s.

And I think that's it! The rest looked great! Thanks @Marwes!

Closes #511

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: trusty
 sudo: required
 language: rust
 rust:
-  - 1.20.0
+  - 1.24.1
   - stable
   - beta
   - nightly
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ recommended for general use.
 
 ### Minimum Rust version policy
 
-This crate's minimum supported `rustc` version is `1.20.0`.
+This crate's minimum supported `rustc` version is `1.24.1`.
 
 The current **tentative** policy is that the minimum Rust version required
 to use this crate can be increased in minor version updates. For example, if
--- a/bench/src/bench.rs
+++ b/bench/src/bench.rs
@@ -44,9 +44,9 @@ pub use ffi::re2::Regex;
 #[cfg(feature = "re-dphobos")]
 pub use ffi::d_phobos::Regex;
 #[cfg(feature = "re-rust")]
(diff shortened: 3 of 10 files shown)


Marwes:
> A large part of the diff consisted of irrelevant style changes. This made the code harder to review and was generally a huge pain to undo. This PR adds a rustfmt config to this repo that explicitly disables rustfmt in an attempt to prevent this from happening in the future.

Sorry about that, I try and disable rustfmt most of the time but I think in this case it got re-formatted during when fixing conflicts in a rebase :/

> The optimization that avoided cloning a state by reading from the inst ptr scratch space was removed. I'm not in principle against this, but the way it was done violated abstraction too much for my taste.

That part was ugly but it is a shame it got removed, it was actually a pretty good speedup since it avoids an allocation + copy in the common case where the cache is hit (somewhere between 5-10% if I remember).

BurntSushi:
Yeah, I figured about the optimization. It just introduced too much complexity in code that is already way too complex. I've been trying to think about ways to rewrite the DFA so that it is simpler, and in particular handles look around in a more obvious way, but I don't have any great ideas yet.

With that said, be another way of looking at this is that if these types of optimizations are helping, them the DFA is spending a lot of time either building itself or thrashing, which is probably never going to be that fast in the first place. So there might be better higher level optimizations available to you. But I'm not sure.

PR #549, opened 2019-01-15
Use https (not http) URL for meritbadge

dholbert:
Right now this readme file uses an insecure http URL to reference a meritbadge image, which ends up producing "broken https" UI on the now-only-"mostly-secure" https crates.io page https://crates.io/crates/regex .  This patch just upgrades this to an HTTPS url (which still works), to avoid that problem. (Literally a 1-character change, changing "http" to "https".)


dholbert:
Here's a screenshot of what this looks like before this change, for reference:

screenshot of current degraded-HTTPS UI

dholbert:
Thanks for the quick reply/approval!

PR #552, opened 2019-01-19
Rust 2018 README updates

rmcteggart-r7:
Making the example use 2018 behaviour the default. 

Here are the changes working in the playground (note 2018 selected):

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=931cef3284f7178eb62554da4a2c035b

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=aaf7f7a741bae6f15de710a16a7340ed

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=619ddc6c95a9a02bd94fd9e19d22c445

Other top 20 crates are updating as well so this is to fall in line. Didn't mention 2015 compatibility as this has largely been ignored elsewhere (except in log crate), with the assumption that anyone on older versions of Rust will know the extern crate syntax.

Having the default 2018 changes in means less friction for new Rustaceans, a lot of whom will only have adopted it post 2018 introduction.

FYI, I can update to include some sort of nod to 2015 if there is a need from the maintainers 👍

Lastly, I noticed docs.rs will still use older syntax. Not sure whether this should be addressed as well. Let me know on merge and I can look into it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,6 @@ If you have lots of dates in text that you'd like to iterate over, then it's
 easy to adapt the above example with an iterator:
 
 ```rust
-extern crate regex;
-
 use regex::Regex;
 
 const TO_SEARCH: &'static str = "
@@ -104,8 +102,7 @@ regular expressions are compiled exactly once.
 For example:
 
 ```rust
-#[macro_use] extern crate lazy_static;
-extern crate regex;
+use lazy_static::lazy_static;
 
 use regex::Regex;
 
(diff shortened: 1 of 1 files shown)


RReverser:
> Having the default 2018 changes in means less friction for new Rustaceans, a lot of whom will only have adopted it post 2018 introduction.

I'm not sure there's that much friction, especially given that current examples are compatible and work in both editions out of the box, while removing `extern crate` will make it work only in 2018.

rmcteggart-r7:
@RReverser can understand your viewpoint but just checking out Rust since 2018 release myself I've found it confusing stumbling across this syntax before being 100% on how it worked. Had it in some places and out others until I got far enough along on my learning to revisit the 2018 notes

PR #561, opened 2019-03-10
Fix a bug in capture with match

pipopa:
When performing "EndText" matching, it is necessary to check whether
the current position matches the input text length.However, when
capturing a submatch using the matching result of DFA, "EndText" matching
wasn't actually performed correctly because the input text is sliced.

By applying this patch we specify the match end position by the argument "end",
not using slice when performing capture with the matching result of DFA.

Fixes rust-lang#557


pipopa:
I formatted the code.

pipopa:
Thank you for reviewing and merging!

PR #578, opened 2019-04-28
Automatically check if README.md examples are working when running "cargo test"

GuillaumeGomez:
Since rustdoc nightly now provides "cfg(test)" when running on test mode, we can now use this macro on test mode only to check if README.md examples are working as expected.


GuillaumeGomez:
Indeed but in other versions, it will just be ignored since rustdoc doesn't pass `cfg(test)`. :)

BurntSushi:
Ah nice! That's right. Great.

Issue #545, opened 2018-12-17
Confusing error with {}


CAD97:
I wrote the regex `r"\\u{[^}]*}"` which works (as `\\u{[^}]*}`) on regex101 under pcre, js, python, and go flavors. When parsing with this crate, it gives:

```

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    \\u{[^}]*}
        ^
error: decimal literal empty
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

"decimal literal empty" was not helpful in figuring out the problem, which was that regex_syntax expects `{}` to be a repetition, and the other mentioned regex engines silently fell back to matching a literal `{` and `}` when it wasn't a repetition. The correct unproblematic regex escapes the braces: `r"\\u\{[^}]*\}"`.

This could be considered a bug or a suboptimal error depending on how you think this regex should be processed. I'd be perfectly happy if the error were to say something along the lines of "expected bounded repetition" here, rather than the current vague "decimal literal empty". (I would understand the error if it were `{}`, but with something other than `}` after the `{`, it's confusing.)

BurntSushi:
Yeah, the error message should definitely be improved here. The behavior does indeed match my intent. Specifically, I biased toward less implicitness in the syntax. That is, if something is a meta character and you want to use it as a literal, then it needs to be escaped. There are some exceptions to this, particularly, in character classes, e.g., `[]]` and `[-a-z]`, due to their prevalence. The thinking here is that if reading a `{` requires a human to go and interpret whether it "needs" to be escaped or not in order to determine whether it's a meta character or not, then the regex becomes harder to read.

PR #582, opened 2019-05-29
syntax: better error message for missing repetition quantifier

christianrondeau:
Hi, first contribution to the Rust ecosystem, hopefully I'll get it right.

This is to solve @CAD97 issue https://github.com/rust-lang/regex/issues/545 which raised an uncleared error message when repetition quantifiers were not followed by a valid decimal (other languages would treat the `{}` as characters and this implementation would fail, but the message was "decimal literal empty" which was confusing)

The change is fairly simple, my main objective is to be able to contribute by starting small.

A few notes:
- To reuse the Span and pattern instances, I match and return the same error values but with a different kind. This way we can keep current and future errors in parse_decimal untouched.
- I also didn't extract a method for the `match` block since the code was small, but if you believe it makes sense, I could also make a `change_error_kind(From, To, self.parse_decimal())` method or pass the desired error kind to `parse_decimal(ErrorKindWhenMissing)`, but I felt the small duplication made it more obvious to read.
- I used `RepetitionQuantifierDecimalMissing` even though it also throws when the decimal is _invalid_, because the current implementation doesn't differentiate missing and invalid decimals, and I didn't want to change the behavior. It is also valid to have a missing decimal in cases like `x{1,}` or `x{,2}`, so I'm sure there's a better name for it.

I also noticed there was no integration tests for error messages. I excluded it from this PR but if you believe it can be valuable (it helped me make sure that my change actually worked) I can include it in this PR as a separate commit (or a different PR if you prefer). The test looks like this:

```rust
fn repetition_quantifier_decimal_error() {
    assert_panic_message(r"\\u{[^}]*}", r#"
regex parse error:
    \\u{[^}]*}
        ^
error: decimal literal empty
"#);
```

I'm ready to do my homework if I missed something, so just let me know!

### DIFF
--- a/regex-syntax/src/ast/mod.rs
+++ b/regex-syntax/src/ast/mod.rs
@@ -153,6 +153,9 @@ pub enum ErrorKind {
     /// The range provided in a counted repetition operator is invalid. The
     /// range is invalid if the start is greater than the end.
     RepetitionCountInvalid,
+    /// An opening `{` was not followed by a valid decimal value.
+    /// For example, `x{}` or `x{]}` would fail.
+    RepetitionQuantifierDecimalMissing,
     /// An opening `{` was found with no corresponding closing `}`.
     RepetitionCountUnclosed,
     /// A repetition operator was applied to a missing sub-expression. This
@@ -306,6 +309,9 @@ impl fmt::Display for ErrorKind {
             RepetitionCountInvalid => {
                 write!(f, "invalid repetition count range, \
                            the start must be <= the end")
+            }
+            RepetitionQuantifierDecimalMissing => {
+                write!(f, "Repetition quantifier expects a valid decimal")
             }
             RepetitionCountUnclosed => {
                 write!(f, "unclosed counted repetition")
--- a/regex-syntax/src/ast/mod.rs
+++ b/regex-syntax/src/ast/mod.rs
@@ -95,7 +95,10 @@ pub enum ErrorKind {
     ClassRangeLiteral,
     /// An opening `[` was found with no corresponding closing `]`.
     ClassUnclosed,
-    /// An empty decimal number was given where one was expected.
+    /// Note that this error variant is no longer used. Namely, a decimal
+    /// number can only appear as a repetition quantifier. When the number
+    /// in a repetition quantifier is empty, then it gets its own specialized
+    /// error, `RepetitionCountDecimalEmpty`.
     DecimalEmpty,
     /// An invalid decimal number was given where one was expected.
     DecimalInvalid,
@@ -155,7 +158,7 @@ pub enum ErrorKind {
     RepetitionCountInvalid,
     /// An opening `{` was not followed by a valid decimal value.
     /// For example, `x{}` or `x{]}` would fail.
-    RepetitionQuantifierDecimalMissing,
+    RepetitionCountDecimalEmpty,
     /// An opening `{` was found with no corresponding closing `}`.
     RepetitionCountUnclosed,
     /// A repetition operator was applied to a missing sub-expression. This
@@ -310,8 +313,8 @@ impl fmt::Display for ErrorKind {
                 write!(f, "invalid repetition count range, \
                            the start must be <= the end")
             }
-            RepetitionQuantifierDecimalMissing => {
-                write!(f, "Repetition quantifier expects a valid decimal")
+            RepetitionCountDecimalEmpty => {
+                write!(f, "repetition quantifier expects a valid decimal")
             }
             RepetitionCountUnclosed => {
                 write!(f, "unclosed counted repetition")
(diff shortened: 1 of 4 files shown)


regex-syntax/src/ast/mod.rs
```diff
@@ -307,6 +310,9 @@ impl fmt::Display for ErrorKind {
                 write!(f, "invalid repetition count range, \
                            the start must be <= the end")
             }
+            RepetitionQuantifierDecimalMissing => {
+                write!(f, "Repetition quantifier expects a valid decimal")
```
BurntSushi: Could this match the style of surrounding style of error messages? i.e., Start with a lowercase letter.

regex-syntax/src/ast/parse.rs
```diff
@@ -1113,7 +1113,15 @@ impl<'s, P: Borrow<Parser>> ParserI<'s, P> {
                 ast::ErrorKind::RepetitionCountUnclosed,
             ));
         }
-        let count_start = self.parse_decimal()?;
+        let count_start = match self.parse_decimal() {
+            Err(ast::Error {kind: ast::ErrorKind::DecimalEmpty, span, pattern}) => return Err(ast::Error {
```
BurntSushi: Could you make this code match the style of the surrounding code? i.e., Keep all lines limited to 79 columns inclusive. Please fix this for any other long lines introduced in this PR. (It is hard to see them in Github's PR diff view.)

It might be clearest to "fix" these long lines by extracting this into a separate method, as you suggested. Then you can reuse that below.

regex-syntax/src/ast/mod.rs
```diff
@@ -153,6 +153,9 @@ pub enum ErrorKind {
     /// The range provided in a counted repetition operator is invalid. The
     /// range is invalid if the start is greater than the end.
     RepetitionCountInvalid,
+    /// An opening `{` was not followed by a valid decimal value.
+    /// For example, `x{}` or `x{]}` would fail.
+    RepetitionQuantifierDecimalMissing,
```
BurntSushi: With this change, it looks like `DecimalEmpty` is no longer used. This is a bit weird, but I see why it makes sense to do that. Could you update the docs of `DecimalEmpty` to say something like:

```rust
/// Note that this error variant is no longer used. Namely, a decimal
/// number can only appear as a repetition quantifier. When the number
/// in a repetition quantifier is empty, then it gets its own specialized
/// error, `RepetitionQuantifierDecimalMissing`.

Comment (BurntSushi) [reply regex-syntax/src/ast/mod.rs]: Also, could you name this `RepetitionCountDecimalEmpty` in order to be consistent with other error variants?

**Comment by christianrondeau** [comment]
Changes:
- Fixed upper case letter in error message (good catch)
- Extracted a function for the error specialization. Called it `specialize_err` and made it generic, so it's simple to reuse and read. It's now under 79 columns wide.
- Updated the doc of DecimalEmpty as asked (I think your text was perfect)
- Renamed `RepetitionQuantifierDecimalMissing ` to `RepetitionCountDecimalEmpty`
- Since the code is already a little bit more modified than I initially wanted, I included my integration test, I it wil be easier for you to tell me if you wish to keep it or if I should remove it.

For your first question, the code currently doesn't handle `a{}` and `a{]}` differently. So "empty" is technically correct, though I thought "missing" was more precise since it "expected" one, but didn't found one (i.e. it's not empty, it's just not there). But my thought process is probably not valid for a future reader, so I used your change. Showing a different error message for empty or invalid decimal would require looking further just for the purposes of the error message, so I thought it wasn't worth the complexity.

File: regex-syntax/src/ast/parse.rs [hunk 4]
```diff
@@ -1129,7 +1133,11 @@ impl<'s, P: Borrow<Parser>> ParserI<'s, P> {
                 ));
             }
             if self.char() != '}' {
-                let count_end = self.parse_decimal()?;
+                let count_end = specialize_err(
+                    self.parse_decimal(),
+                    ast::ErrorKind::DecimalEmpty,
+                    ast::ErrorKind::RepetitionCountDecimalEmpty,
+                )?;
```
BurntSushi: Ah yes, this is much nicer. I like `specialize_err`. :-)

regex-syntax/src/ast/parse.rs
```diff
@@ -2260,6 +2268,29 @@ impl<'p, 's, P: Borrow<Parser>> ast::Visitor for NestLimiter<'p, 's, P> {
...
 
+/// When the result is an error, transforms the ast::ErrorKind from the source
+/// Result into another one. This function is used to return clearer error
+/// messages when possible.
+fn specialize_err<T>(
+    result: Result<T>,
+    from: ast::ErrorKind,
+    to: ast::ErrorKind,
+) -> Result<T> {
+    if let Err(e) = result {
```
CAD97: This could be written with a match guard:

```rust
match result {
    Err(e) if e.kind == from => Err(ast::Error { kind: to, ..e }),
    _ => result,
}
```

CAD97: Wait nevermind this is moving

PR #585, opened 2019-06-08
Fix numbered group binding

max-sixty:
Ref https://github.com/rust-lang/regex/issues/69 should `re.replace("ab","$1_$2")` equal `"a_b"` or `"ab"`? 

In [this test](https://github.com/rust-lang/regex/blob/89467a9325c09249772920cd1e21282442fc9798/src/expand.rs#L207) `$42a` binds as a single group reference; which would suggest `$1_` did the same (I think?)

[Here](https://regexr.com/4ff3e) it looks like any non-number character following the number does not bind, and so `$42a` would only bind the `42`. I'm no regex expert but a cursory look suggests there's are various approaches to valid number references: https://www.regular-expressions.info/replacebackref.html

As an initial attempt, I changed the test case referenced above. 

Let me know if this is reasonable

### DIFF
--- a/src/expand.rs
+++ b/src/expand.rs
@@ -92,7 +92,7 @@ pub fn expand_bytes(
 /// `CaptureRef` represents a reference to a capture group inside some text.
 /// The reference is either a capture group name or a number.
 ///
-/// It is also tagged with the position in the text immediately proceeding the
+/// It is also tagged with the position in the text following the
 /// capture reference.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 struct CaptureRef<'a> {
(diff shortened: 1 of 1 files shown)


max-sixty:
Reading the docs, I actually think the current implementation is correct:

> The longest possible name is used. e.g., $1a looks up the capture group named 1a and not the capture group at index 1. To exert more precise control over the name, use braces, e.g., ${1}a.

In the motivating example, `re.replace("ab","$1_$2")` should just be `b`, because `1_` is captured and doesn't match any names.

If that's correct: should we keep the existing behavior or make the change? I have a patch that adds a couple tests in the prior case

BurntSushi:
Unfortunately, I think we're going to want to keep the existing behavior. Making this change now would be a breaking change. If the docs and implementation differed (where the docs stated the ideal behavior), then I might be willing to call this an implementation bug and fix it. But since the docs and implementation match, I don't think we can get away with fixing this.

So I think perhaps some more test on the current behavior would be great. I will mark #69 as a regex2 issue.

max-sixty:
OK no problem, I'll update with a couple of tests.

Any other issues where a beginner can contribute? This was the easiest I could find where you pointed to a function and specified the change...

max-sixty:
@BurntSushi this ended up not being much - but still worth merging? Anything else needed?

max-sixty:
Gentle ping to push a notification!

PR #589, opened 2019-07-01
Make u8x16 and u8x32 have Vector call ABI

### LINKED ISSUE #588: Unions used for type-punning should be `repr(C)`
At

https://github.com/rust-lang/regex/blob/172898a4fda4fd6a2d1be9fc7b8a0ea971c84ca6/src/vector/ssse3.rs#L80-L83

a union is declared for the purpose of type-punning. However, without `repr(C)`, the fields of that union might not be at offset 0, which makes it unsuited for type punning. To get such layout guarantees, you need to specify `repr(C)`.

Also see https://rust-lang.github.io/unsafe-code-guidelines/layout/unions.html.

gnzlbg:
Before this commit, u8x16 and u8x32 were repr(Rust) unions. This introduced
unspecified behavior because the field offsets of repr(Rust) unions are not
guaranteed to be at offset 0, so that field access was potentially UB.

This commit fixes that, and closes #588 .

The unions were also generating a lot of unnecessary memory operations. This
commit fixes that as well.

The issue is that unions have an Aggregate call ABI, which is the same as the
call ABI of arrays. That is, they are passed around by memory, and not in Vector
registers.

This is good, if most of the time one operates on them as arrays. This was,
however, not the case. Most of the operations on these unions are using SIMD
instructions. This means that the union needs to be copied into a SIMD register,
operated on, and then spilled back to the stack, on every single operation.
That's unnecessary, although apparently LLVM was able to optimize all the
unnecessary memory operations away and leave these always in registers.

This commit fixes this issue as well, by making the u8x16 and u8x32
repr(transparent) newtypes over the architecture specific vector types, giving
them the Vector ABI.

The vectors are then copied to the stack only when necessary, and as little as
possible. This is done using mem::transmute, removing the need for unions
altogether (fixing #588 by not having to worry about union layout at all).

To make it clear when the vectors are spilled into the stack, the
vector::replace(index, value) and vector::extract(index) APIs have been removed,
and instead, only a vector::bytes(self) and a vector::from_bytes(&mut self, [u8;
N]) APIs are provided instead. This prevents spilling the vectors back and forth
onto the stack every time an index needs to be modified, by using vector::bytes
to spill the vector to the stack once, making all the random-access
modifications in memory, and then using vector::from_bytes only once to move the
memory back into a SIMD register.

---

I haven't run any benchmarks, so please do benchmark that this does not introduce any performance regressions before merging this

### DIFF
--- a/src/literal/teddy_avx2/imp.rs
+++ b/src/literal/teddy_avx2/imp.rs
@@ -285,6 +285,7 @@ impl Teddy {
         res: u8x32,
         mut bitfield: u32,
     ) -> Option<Match> {
+        let patterns = res.bytes();
         while bitfield != 0 {
             // The next offset, relative to pos, where some fingerprint
             // matched.
@@ -296,7 +297,7 @@ impl Teddy {
 
             // The bitfield telling us which patterns had fingerprints that
             // match at this starting position.
-            let mut patterns = res.extract(byte_pos);
+            let mut patterns = patterns[byte_pos];
             while patterns != 0 {
                 let bucket = patterns.trailing_zeros() as usize;
                 patterns &= !(1 << bucket);
@@ -461,12 +462,20 @@ impl Mask {
         let byte_lo = (byte & 0xF) as usize;
         let byte_hi = (byte >> 4) as usize;
 
-        let lo = self.lo.extract(byte_lo) | ((1 << bucket) as u8);
-        self.lo.replace(byte_lo, lo);
-        self.lo.replace(byte_lo + 16, lo);
+        {
+            let mut lo_bytes = self.lo.bytes();
+            let lo = lo_bytes[byte_lo] | ((1 << bucket) as u8);
+            lo_bytes[byte_lo] = lo;
+            lo_bytes[byte_lo + 16] = lo;
+            self.lo.replace_bytes(lo_bytes);
+        }
 
-        let hi = self.hi.extract(byte_hi) | ((1 << bucket) as u8);
-        self.hi.replace(byte_hi, hi);
-        self.hi.replace(byte_hi + 16, hi);
+        {
+            let mut hi_bytes = self.hi.bytes();
+            let hi = hi_bytes[byte_hi] | ((1 << bucket) as u8);
+            hi_bytes[byte_hi] = hi;
+            hi_bytes[byte_hi + 16] = hi;
+            self.hi.replace_bytes(hi_bytes);
+        }
     }
 }
--- a/src/vector/avx2.rs
+++ b/src/vector/avx2.rs
@@ -58,12 +58,14 @@ impl AVX2VectorBuilder {
 #[derive(Clone, Copy)]
 #[allow(non_camel_case_types)]
 #[repr(transparent)]
-pub struct u8x32(__m256i);
+pub struct u8x32 {
+    vector: __m256i
+}
 
 impl u8x32 {
     #[inline]
     unsafe fn splat(n: u8) -> u8x32 {
-        u8x32(_mm256_set1_epi8(n as i8))
+        u8x32 { vector: _mm256_set1_epi8(n as i8) }
     }
 
     #[inline]
@@ -75,7 +77,7 @@ impl u8x32 {
     #[inline]
     unsafe fn load_unchecked_unaligned(slice: &[u8]) -> u8x32 {
         let p = slice.as_ptr() as *const u8 as *const __m256i;
-        u8x32(_mm256_loadu_si256(p))
+        u8x32 { vector: _mm256_loadu_si256(p) }
     }
 
     #[inline]
@@ -88,14 +90,14 @@ impl u8x32 {
     #[inline]
     unsafe fn load_unchecked(slice: &[u8]) -> u8x32 {
         let p = slice.as_ptr() as *const u8 as *const __m256i;
-        u8x32(_mm256_load_si256(p))
+        u8x32 { vector: _mm256_load_si256(p) }
     }
 
     #[inline]
     pub fn shuffle(self, indices: u8x32) -> u8x32 {
         // Safe because we know AVX2 is enabled.
         unsafe {
-            u8x32(_mm256_shuffle_epi8(self.0, indices.0))
+            u8x32 { vector: _mm256_shuffle_epi8(self.vector, indices.vector) }
         }
     }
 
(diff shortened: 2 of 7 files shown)


src/vector/ssse3.rs
```diff
@@ -110,83 +109,82 @@ impl u8x16 {
...
+            u8x16(_mm_andnot_si128(boolv, ones))
         }
     }
 
     #[inline]
     pub fn and(self, other: u8x16) -> u8x16 {
         // Safe because we know SSSE3 is enabled.
         unsafe {
-            u8x16 { vector: _mm_and_si128(self.vector, other.vector) }
+            u8x16(_mm_and_si128(self.0, other.0))
```
RReverser: Instead of changing all these places, maybe it would've been better to keep named field for `vector`? It wouldn't affect ability to use `repr(transparent)` in any way while would keep code a bit more readable.

src/vector/avx2.rs
```diff
@@ -164,24 +151,35 @@ impl u8x32 {
...
     }
+
+    #[inline]
+    pub fn bytes(self) -> [u8; 32] {
+        // Safe because __m256i and [u8; 32] are layout compatible
+        unsafe { mem::transmute(self) }
+    }
+
+    #[inline]
+    pub fn replace_bytes(&mut self, value: [u8; 32]) {
```
RReverser: Perhaps better to have `bytes(&self) -> &[u8; 32]` and `bytes_mut(&mut self) -> &mut [u8; 32]`? It would be more Rust-idiomatic than providing a getter+setter, and in some cases might give a faster way to change separate bytes without replacing the whole thing.

gnzlbg: > Perhaps better to have bytes(&self) -> &[u8; 32]

Where does that `&[u8; 32]` live ?

RReverser: The body of the function wouldn't change - you'd still use `transmute`, just between references and not values.

RReverser: Here's an example with both implementations and you can see how code for setting a single byte differs between two: https://rust.godbolt.org/z/XF7m95

gnzlbg: If the `__m256i` is in a SIMD register, it doesn't have a memory address, so creating a pointer to it requires spilling it to the stack to be able to give it an address. Once the modification on the stack is done, moving it back into a SIMD register requires copying the whole thing, not a single byte, particularly if dynamic indices are involved (if the indices are compile-time constant sometimes the compiler can do better, if there is a specific instruction for that).

One doesn't want this back-and-forth to happen every time one modifies a part of the vector. The proposed API forces its users to make the moves back to memory manually for the whole vector, and to copy the result manually back into the SIMD register. This revealed a couple of places where, e.g., instead of doing this back-and-forth on every iteration of a loop, one can just load the vector into memory once before the loop, operate on memory, and move the contents back into a SIMD register after the loop has completed.

src/literal/teddy_avx2/imp.rs
```diff
@@ -461,12 +462,20 @@ impl Mask {
         let byte_lo = (byte & 0xF) as usize;
         let byte_hi = (byte >> 4) as usize;
 
-        let lo = self.lo.extract(byte_lo) | ((1 << bucket) as u8);
-        self.lo.replace(byte_lo, lo);
-        self.lo.replace(byte_lo + 16, lo);
```
gnzlbg: @RReverser for example, see here. Every call to replace requires spilling the vector to memory, modifying its contents, and then writing it back to a SIMD register.

The change explicitly spills the vector to the stack once. Modifies the stack, and moves the memory back into the vector, once.

src/vector/avx2.rs
RReverser: That's a good point. I'd argue that having several `.bytes_mut()[...] = ...` patterns in a row would also be pretty obvious and most people would store the reference when they could, but I see how your proposed API forces this a bit better.

However, if that's the goal, I wonder if it's worth to go one step further and take mutations as a callback like `.with_bytes(|bytes| ...)` where one can modify contents in any way they want inside, but then on the type system level you wouldn't be able to do anything else outside, and transformation back to a register would be guaranteed to happen as part of the call?

gnzlbg: > However, if that's the goal, I wonder if it's worth to go one step further and take mutations as a callback like .with_bytes(|bytes| ...) where one can modify contents in any way they want inside, but then on the type system level you wouldn't be able to do anything else outside, and transformation back to a register would be guaranteed to happen as part of the call?

Some code only wants to read the bytes, so forcing a write would be bad for that code - I've pinged you in one example.

I also don't see really an advantage in restricting access to a scope. If you wanted to mutate two vectors, you would need to nest scopes. For an API that's only intended for internal consumption, this API would feel even more like overengineering than the one I proposed - and I consider mine borderline overengineering.

src/literal/teddy_avx2/imp.rs
```diff
@@ -285,6 +285,7 @@ impl Teddy {
         res: u8x32,
         mut bitfield: u32,
     ) -> Option<Match> {
+        let patterns = res.bytes();
```
gnzlbg: See this function, which only needs to read the bytes.

src/vector/avx2.rs
RReverser: Okay, fair enough. I'm not 100% comfortable with `replace_bytes` vs `bytes_mut` for idiomaticity reasons, but I can see how this is the least of the evils :)

gnzlbg: > Here's an example with both implementations and you can see how code for setting a single byte differs between two: https://rust.godbolt.org/z/XF7m95

As mentioned, this should properly benchmarked before being merged. Those examples do not really use the intrinsics, so the code is being generated in isolation without any objective in mind, which does not represent what this library actually does. If you actually try to use the, you'll see that they generate the exact same code when all optimizations are turned on: https://rust.godbolt.org/z/Z8sdH9

But this is not about optimizing the implementation, this is about optimizing the amount of work that LLVM has to do to produce efficient code. If you look at the LLVM-IR produced in debug mode: https://rust.godbolt.org/z/sMfxHm the version using `replace_bytes` produces 134 lines of LLVM-IR, while the version using `bytes_mut` produces 758 lines of LLVM-IR. There are a couple of factors at play here, but producing 6x more LLVM-IR for this isn't really worth it.

gnzlbg:
@BurntSushi my benchmarks results are here: https://gist.github.com/gnzlbg/dba8836f80237e0affd63d13d6274a37 - feels a bit like noise overall

BurntSushi:
Thanks so much for this! The vector/aggregate ABI issue is something I was completely unaware of, and in retrospect, makes a lot of sense. When I first ported the Teddy code to `std::arch` (which is when I introduced the errant union type punning), I noticed that small perturbations to the code would result in fairly different codegen, in particular, with lots of additional `mov` instructions. It now seems likely that the root cause of this was the tension you described between the aggregate/vector ABIs. It seems LLVM was really saving my ass here. :P 

In any case, I've run the benchmarks myself and everything looks okay. (The benchmarks are unfortunately pretty noisy.) I also did some ad hoc benchmarks by hand with `perf` and inspected the codegen. The codegen appears to be unchanged, with the exception of one fewer `mov` instruction, which makes sense given your explanation.

BurntSushi:
This PR is now in `regex 1.1.8` on crates.io.

Issue #521, opened 2018-10-08
splitn(.., 2) returning extra substring


casey:
I'm seeing an extra substring produced by `splitn` in [this code](https://play.rust-lang.org/?gist=ec14b389fde3e7080c0d8d7620a57125&version=stable&mode=debug&edition=2015).

I'd expect splitn(1), splitn(2), and splitn(3) to produce the same output, but splitn is returning an extra empty substring.

PR #606, opened 2019-08-07
Fix splitn bug

max-sixty:
I think this should close https://github.com/rust-lang/regex/issues/521


max-sixty:
I realized that `splitn` is copy pasted into two files, so changed in `re_bytes.rs` too

I don't know the repo well, but is that necessary? I made the code the same with `Self::Item` - would there be a way of reducing the actual duplication?

max-sixty:
Gentlest of reminders @BurntSushi !

max-sixty:
Could also use [fuse](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fuse) here.

max-sixty:
Can anyone review this, even if not @BurntSushi ?

src/re_bytes.rs
```diff
@@ -757,14 +757,19 @@ pub struct SplitN<'r, 't> {
 impl<'r, 't> Iterator for SplitN<'r, 't> {
     type Item = &'t [u8];
 
-    fn next(&mut self) -> Option<&'t [u8]> {
-        if self.n == 0 {
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.n <= 0 {
```
ethanpailes: How can a `usize` ever drop below 0?

src/re_unicode.rs
```diff
@@ -797,14 +797,19 @@ pub struct SplitN<'r, 't> {
 impl<'r, 't> Iterator for SplitN<'r, 't> {
     type Item = &'t str;
 
-    fn next(&mut self) -> Option<&'t str> {
-        if self.n == 0 {
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.n <= 0 {
```
ethanpailes: Same question

src/re_bytes.rs
max-sixty: Keep calling `.next()` even after it's returned `None`, and n goes past 0 to negative...

ethanpailes: I'm pretty sure it won't because this if block will always fire and cause a return before n gets decremented. Also, because n is a usize there is no pattern of bits which constitutes a negative number.

max-sixty: Yes ofc you're right, thanks. Updated

danielparks:
Unfortunately this PR breaks the case where there actually should be a trailing `""` returned.

I’m working on a fix as part of #627, which deals with `split()` losing trailing `""`s.

danielparks:
PR #628 submitted.

I did not pull in your (@max-sixty) changes to `Self::Item` since I wanted to keep it as limited as possible. Also, the other iterators I checked didn’t use `Item`. That could certainly use some magic help, but I think that needs to be another PR.

Issue #583, opened 2019-05-31
Binary Size


cramertj:
This isn't yet focused on any particular issue, but if the maintainers are okay with it, I'd like to use this as an issue to track improvements to binary size. Today, a simple hello world build with `--release` and stripped is ~240KB. Adding `regex::Regex::new("x").unwrap()` to that binary and recompiling + restripping gives ~1.25MB-- that's an increase of over 1 megabyte, which is quite large when targeting resource-constrained systems.

`cargo bloat --release` output:

```
 File  .text     Size        Crate Name
12.9%  65.9% 379.6KiB              [1519 Others]
 1.7%   8.5%  49.2KiB regex_syntax regex_syntax::ast::parse::ParserI<P>::parse_with_comments
 0.6%   2.8%  16.4KiB        regex regex::exec::ExecBuilder::build
 0.5%   2.4%  14.1KiB regex_syntax <regex_syntax::hir::translate::TranslatorI as regex_syntax::as...
 0.4%   2.0%  11.6KiB regex_syntax regex_syntax::ast::parse::ParserI<P>::parse_escape
 0.3%   1.8%  10.3KiB regex_syntax regex_syntax::unicode::class
 0.3%   1.7%   9.8KiB        regex regex::compile::Compiler::c
 0.3%   1.7%   9.8KiB regex_syntax <regex_syntax::hir::translate::TranslatorI as regex_syntax::as...
 0.3%   1.5%   8.6KiB    [Unknown] read_line_info
 0.3%   1.4%   8.0KiB          std _ZN9libunwind10CFI_ParserINS_17LocalAddressSpaceEE17parseInstr...
 0.2%   1.1%   6.5KiB    [Unknown] elf_add
 0.2%   1.1%   6.3KiB regex_syntax regex_syntax::hir::literal::suffixes
 0.2%   1.1%   6.3KiB          std __rdos_backtrace_dwarf_add
 0.2%   1.1%   6.2KiB        regex aho_corasick::nfa::Builder::build
 0.2%   1.1%   6.2KiB regex_syntax regex_syntax::hir::literal::prefixes
 0.2%   0.8%   4.8KiB        regex regex::compile::Compiler::compile
 0.2%   0.8%   4.7KiB          std rustc_demangle::demangle
 0.2%   0.8%   4.7KiB          std <rustc_demangle::Demangle as core::fmt::Display>::fmt
 0.2%   0.8%   4.5KiB regex_syntax regex_syntax::ast::visitor::visit
[...]

PR #612, opened 2019-08-28
Add optimizations into release profile

Crandel:
I think for release profile we should do some optimizations. Inspired by this [comment](https://github.com/rust-lang/regex/issues/583#issuecomment-497964678)


BurntSushi:
Thanks for the PR, but this doesn't make a lot of sense. These options are typically set on the final binary. I don't think they have much of an effect here. More to point, if they did have an effect, this is definitely not the same trade off I would choose. `lto` will _drastically_ increase compile times.

PR #614, opened 2019-09-05
Small Refactoring

jeremystucki:


src/re_bytes.rs
```diff
@@ -1070,7 +1070,7 @@ pub trait Replacer {
     /// be beneficial to avoid finding sub-captures.
     ///
     /// In general, this is called once for every call to `replacen`.
-    fn no_expansion<'r>(&'r mut self) -> Option<Cow<'r, [u8]>> {
```
BurntSushi: Hmmm I'd actually rather have these lifetimes. I don't agree with all flavors of lifetime elision. In particular, it's not necessarily obvious that the returned `Cow` has a lifetime attached to `sefl` when the lifetimes are elided.

BurntSushi: I think this comment applies to all changes in this commit except for `ClassQuery`, which I like.

bench/src/ffi/tcl.rs
```diff
@@ -134,13 +134,10 @@ impl<'r, 't> Iterator for FindMatches<'r, 't> {
...
-            None => None,
-            Some((s, e)) => {
-                self.last_match = e;
-                Some((s, e))
-            }
-        }
+        self.re.find_at(self.text, self.last_match).map(|(s, e)| {
+            self.last_match = e;
+            (s, e)
+        })
```
BurntSushi: I personally am not a fan of using `map` combined with mutations on other objects. Maybe revert this change and others like it above?

regex-syntax/src/hir/translate.rs
```diff
@@ -318,12 +318,9 @@ impl<'t, 'p> Visitor for TranslatorI<'t, 'p> {
...
                     )?;
-                    if cls.iter().next().is_none() {
-                        return Err(self.error(
-                            ast.span,
-                            ErrorKind::EmptyClassNotAllowed,
-                        ));
-                    }
+                    cls.iter().next().is_none().ok_or_else(|| {
+                        self.error(ast.span, ErrorKind::EmptyClassNotAllowed)
+                    })?;
```
BurntSushi: This also just reads weird to me. It feels much more natural as an `if` statement.

src/utf8.rs
```diff
@@ -132,11 +132,13 @@ pub fn decode_last_utf8(src: &[u8]) -> Option<(char, usize)> {
...
-        Some((_, n)) if n < src.len() - start => None,
-        Some((cp, n)) => Some((cp, n)),
-    }
+    decode_utf8(&src[start..]).and_then(|(cp, n)| {
+        if n < src.len() - start {
+            None
+        } else {
+            Some((cp, n))
+        }
+    })
```
BurntSushi: I prefer the previous version. Here, you have to process a monadic bind followed by case analysis. Before, it was just case analysis.

src/re_bytes.rs
matklad: I might be wrong, but I think `fn class(query: ClassQuery)` will trigger a warning once edition lints are enabled. I *think* it is supposed to be spelled `pub fn class(query: ClassQuery<'_>)`, ie, 2018 edition soft-deprecates lifetime elision outside of `&` and `&mut`. Similarly, the best way to spell this fn might be `fn no_expansion(&mut self) -> Option<Cow<'_, [u8]>>`. The `'_` syntax was stablized in 1.26, so I think we can use it.

BurntSushi: Not a huge fan of the `'_` syntax either to be honest... I like the fact that lifetime parameters are introduced just like type parameters.

I'll likely give in once regex switches to 2018, but I'd prefer to just leave well enough alone for now. :-)

PR #616, opened 2019-09-13
Bump quickcheck to 0.9

eclipseo:


BurntSushi:
Nope. It bumps the MSRV and I'm not ready to do that. Sorry.

ethanpailes:
Wouldn't it be ok because quickcheck is just a dev dependency? Is it just that having a different MSRV for dev vs non-dev makes testing and locking an MSRV in CI much harder?

BurntSushi:
Normally it would be fine. But rand is using a new feature in Cargo.toml that breaks non-dev builds on older versions of Rust. Just try a normal build on Rust 1.28 and you should see the error.

ethanpailes:
Gotcha

PR #620, opened 2019-10-04
Implement Ord for Class

pczarn:
Added Ord and PartialOrd for Class, IntervalSet,
ClassUnicode and ClassBytes

### DIFF
--- a/regex-syntax/src/hir/interval.rs
+++ b/regex-syntax/src/hir/interval.rs
@@ -32,7 +32,7 @@ use unicode;
 //
 // Tests on this are relegated to the public API of HIR in src/hir.rs.
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
 pub struct IntervalSet<I> {
     ranges: Vec<I>,
 }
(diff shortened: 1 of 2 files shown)


pczarn:
This probably needs a minor version bump.

RReverser:
Why though? Classes and sets don't really have a well-defined ordering semantics.

pczarn:
@RReverser should I implement Hash instead? Or should I use vecs of classes instead of maps? I wanted to use classes in map keys.

RReverser:
To be clear - I'm not a maintainer here, and my position doesn't mean much, but personally I find these traits misleading when they're implemented just for inclusion in other data structures, without actual well-defined ordering on the type.

`Hash` is IMO better, because it doesn't require any defined semantics.

BurntSushi:
@pczarn Could you say more about the high level problem you're trying to solve here?

@RReverser Note that these aren't just sets. These are ordered sets, and that is guaranteed as part of their public API. So in that sense, I think `Ord` impls do have an "obvious" semantics.

RReverser:
> These are ordered sets, and that is guaranteed as part of their public API.

Their elements are ordered, sure, but ordering of sets with each other seems less obvious.

BurntSushi:
Yeah, I wrote out a few examples and it does seem a bit weird. Using the definition in this PR for example:

```
[[1, 5]] > [[1, 4]]
[[1, 5]] > [[1, 4], [5, 5]]
```

The latter one should be equivalent in my view. Although, `Class` does have the property that any two adjacent ranges are _not_ continuous. But maybe there are more weird examples like this.

@pczarn I think implementing `Hash` for this would be uncontroversial. I'd be potentially open to `Ord`, but I think I'd need to see a more rigorous argument for it.

Issue #594, opened 2019-07-08
`\p\` interacts poorly with error reporting


CAD97:
This is best explained with examples:

```text
// this one is good
regex parse error:
    \p\
    ^^^
error: Unicode property not found

// but putting things after the "escape" behaves strangely
regex parse error:
    \p\{
       ^
error: unclosed counted repetition

regex parse error:
    \p\{\}
        ^
error: decimal literal empty

// yet this one works
regex parse error:
    \p\+
    ^^^
error: Unicode property not found
```

It seems that `\p\` is treated as "whole" for the bracket-based followers, and that error preempts the prior error about the missing Unicode property.

<details><summary>Why I noticed this</summary>

Trying out [tree-sitter](https://tree-sitter.github.io/), which uses regex-syntax for parsing regex terminals but for some reason [preprocesses curly brackets in regex](https://github.com/tree-sitter/tree-sitter/issues/380). I wrote the regex `\p{XID_Start}` which it translated into `\p\{XID_Start\}` which then gave me the very confusing error of

```text
>tree-sitter generate
Error processing rule identifier
Details:
  regex parse error:
    \p\{XID_Start\}
        ^
error: repetition quantifier expects a valid decimal
```

</details>

BurntSushi:
Nice find. Yeah, this is indeed quite odd. It looks like the problem is here:

https://github.com/rust-lang/regex/blob/df23d5abe4ce63792009bf4ef6db72a826fb7d5c/regex-syntax/src/ast/parse.rs#L2086-L2090

Which is triggered when a `\p` (or `\P`) is followed by _anything_ other than a `{`. Which means it gets triggered for `\p\{` since `\` is not a `{`. This particular code path is for handling single letter Unicode classes, e.g., `\pL` is `\p{Letter}`. The reason this gets fouled up is because the parser isn't responsible for checking the validity of the Unicode class given. e.g., The parser will happily transform `\p{fubar}` into a valid AST. Checking the validity of the class name is the responsibility of the translator (AST -> HIR).

I think it is true that a `\p` followed by a `\` is never correct, so I think the simple solution here is to just return an error in the parser for that specific case. (This might require a new error variant, which is fine.)

Dianne087:
> This is best explained with examples:
> 
> ```
> // this one is good
> regex parse error:
>     \p\
>     ^^^
> error: Unicode property not found
> 
> // but putting things after the "escape" behaves strangely
> regex parse error:
>     \p\{
>        ^
[...]

PR #622, opened 2019-10-08
add explicit error for \ as unicode class

d-dorazio:
this should close https://github.com/rust-lang/regex/issues/594 by adding a new `ErrorKind` variant that should hopefully improve diagnostics.


d-dorazio:
I'm not quite sure why only one job is failing, but I think it has nothing to do with my changes. I suspect rustfmt got updated in the meanwhile and now it wants to reformat things differently, but I've not double checked that.

PR #623, opened 2019-10-08
Add byte frequencies to regex-syntax

ffminus:
The frequencies, along with the associated `freq` method, provide a way to estimate the rarity of a given literal.

The approximation is pretty crude and assumes independence between bytes.
Possible future improvements include byte pairs analysis or input language-specific frequencies.

This change enables frequency-based heuristics for client code.
I have a [follow-up commit](https://github.com/ffminus/ripgrep/commit/7d55bb71d3d9943bf96bdff777c9b6b6ba9849a4) for `ripgrep` that yields pretty good improvements.


RReverser:
I'm not really sure that frequencies belong to the regex-syntax; I'd expect it to be just a parser without heuristics like this.

Issue #627, opened 2019-11-18
split() loses final part if it’s empty


danielparks:
I ran into a few bugs:

Regex | Text    | Limit  | Expected        | Actual       | Notes
------|---------|--------|-----------------|--------------|-------
`"-"` | `"a"`   | _none_ | `["a"]`         | `["a"]`      | 
`"-"` | `"a-"`  | _none_ | `["a", ""]`     | `["a"]`      | incorrect
`"-"` | `"a--"` | _none_ | `["a", "", ""]` | `["a", ""]`  | incorrect
`"-"` | `"a"`   | 2      | `["a"]`         | `["a", ""]`  | #521
`"-"` | `"a-"`  | 2      | `["a", ""]`     | `["a", ""]`  | 
`"-"` | `"a--"` | 2      | `["a", "-"]`    | `["a", "-"]` | 

See also #521: `splitn` incorrectly returns a trailing `""` if there is one fewer part than the limit.

danielparks:
I’ve already written tests; I should have a PR ready in the next couple of days.

PR #628, opened 2019-11-18
(#521, #627) Fixes for split() and splitn()

### LINKED ISSUE #521: splitn(.., 2) returning extra substring

danielparks:
* #627: Corrects `/-/.split("a-")` to return `["a", ""]` correctly instead of `["a"]`.
* #521: Corrects `/-/.splitn("a", 2)` to return `["a"]` correctly instead of `["a", ""]`.
* Adds a bunch of tests.
* Adds entries in CHANGELOG.md.

Note that a couple of existing tests (`split2` and `split3`) were passing incorrectly before these changes.

I’m using `/-/` as a short hand for `Regex::new("-").unwrap()`. It seems clear enough, but then I’ve written a lot of code in languages with regex literals.

This supersedes PR #606 which attempted to fix #521.

_I disclaim ownership of these changes. They are owned by whoever owns the bulk of the rest of the code. (I am not a lawyer, but I’ll head them off at the pass if I can.)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Unreleased
 * [BUG #627](https://github.com/rust-lang/regex/issues/627):
   Corrects `/-/.split("a-")` to return `["a", ""]` correctly instead of `["a"]`
   (where `/-/` is a placeholder for code generate a regex).
+* [BUG #521](https://github.com/rust-lang/regex/issues/521):
+  Corrects `/-/.splitn("a", 2)` to return `["a"]` correctly instead of
+  `["a", ""]` (where `/-/` is a placeholder for code generate a regex).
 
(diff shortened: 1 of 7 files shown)


danielparks:
The CI failure is only for Rust “stable.” It appears to be unrelated to the changes in this PR. https://travis-ci.com/rust-lang/regex/jobs/257680677

danielparks:
CI failure issue https://github.com/rust-lang/regex/issues/634

PR #633, opened 2019-12-22
Remove deprecated Error::description

AnderEnder:
`Error::description` has been documented as soft-deprecated since 1.27.0 (17 months ago). It is going to be hard-deprecated soon.

This PR:
- Removes all implementations of `description` in all error types

Related PR: https://github.com/rust-lang/rust/pull/66919


BurntSushi:
Thanks for this PR! Unfortunately, we can't just remove methods like this that are part of a crate's public API. It would require a semver breaking change release, which I don't think is worth doing just for this. Instead, in #637, I've simply squashed the deprecation warnings. I can remove them on the next semver breaking change release.

Dianne087 (commented):
Mvc thank

PR #636, opened 2020-01-06
Library type required for compiling using armv7

serjor:
I was trying to compile a program that uses regex as dependency on a Raspberry Pi, but it wasn't possible as regex-search wasn't provided as "rlib". Adding this setting to Cargo.toml makes it work, and also it works and compiles on desktop.

### DIFF
--- a/regex-syntax/Cargo.toml
+++ b/regex-syntax/Cargo.toml
@@ -30,3 +30,7 @@ unicode-gencat = []
 unicode-perl = []
 unicode-script = []
 unicode-segment = []
+
+[lib]
+crate-type = ["rlib"]
+


BurntSushi:
cc @alexcrichton Do you know why `crate-type = ["rlib"]`  is needed here? It seems fishy to me.

alexcrichton:
This should not be required since it's Cargo's default already. It's likely that the error you saw @serjor was perhaps caused by something else and accidentally looked like this fixed it?

BurntSushi:
Thanks! @serjor Feel free to re-open if there are any further developments.

serjor:
It seems that the problem only occurs when enabling "perf" feature (which I need unfortunately), so yes, the problem lies elsewhere.

Thanks for your time!!

PR #647, opened 2020-02-22
Small code improvements

rodoufu:
- Implementing is_empty;
- Removing unnecessary lifetime;
- Implementing default.


src/literal/imp.rs
```diff
@@ -645,7 +645,7 @@ impl BoyerMooreSearch {
     }
 
     fn len(&self) -> usize {
-        return self.pattern.len();
```
BurntSushi, line 648: I think this is actually the only change in this PR that I agree with. Could you remove the rest please?

rodoufu:
> Thanks! But I actually disagree with most of the changes in this PR. The `Default` impls aren't necessary, so I'm not sure why they've been added. The `is_empty` methods on `Captures` also don't really make sense.

Thanks for the response, so I'm closing the PR.

PR #649, opened 2020-03-01
Add support for additional characters in group names

bruceg:
We have an application that requires the ability to use field access
notation, which includes periods and square braces in group names. This
commit adds support and tests for these additional characters.

### DIFF
--- a/regex-syntax/src/ast/parse.rs
+++ b/regex-syntax/src/ast/parse.rs
@@ -98,10 +98,11 @@ fn is_hex(c: char) -> bool {
 /// Returns true if the given character is a valid in a capture group name.
 ///
 /// If `first` is true, then `c` is treated as the first character in the
-/// group name (which is not allowed to be a digit).
+/// group name (which must be alphabetic or underscore).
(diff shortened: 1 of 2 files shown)


BurntSushi:
So I think there are two problems with this PR.

The first one is that this only changes what's allowed in the syntax, but doesn't modify replacement expansion to account for the new characters.

The second one is that if we do modify expansion to account for the new characters, then sadly, this becomes a breaking change. Consider this case:

```
regex: \[(?P<Z>[a-z]+)
haystack: [abc
replacement: $Z[
```

The current (and correct) result is `abc[`.  But if `[` became a valid capture letter, then the expansion parser would inclue `[` as part of the name, so it would look for the group named `Z[` instead of `Z`.

The only way I can think of to do this without making it a breaking change would be to require the use of braces when using characters in a capture group outside of `[_0-9A-Za-z]`. So that the behavior above would remain the same, but folks could use, e.g., `${foo[bar].baz}` if they wanted to reference groups with `[`, `]` or `.` in their names. Doing this will require some delicate surgery on the expansion routine.

I'll look into doing this.

PR #655, opened 2020-03-24
Improve allocation of escape_into

hhirtz:
In regex-syntax,
Make escape_into allocate at most twice (all characters are escaped)
instead of more if used on an unallocated String.  This way, escape
doesn't have to allocate.

### DIFF
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -175,7 +175,7 @@ pub mod utf8;
 /// The string returned may be safely used as a literal in a regular
 /// expression.
 pub fn escape(text: &str) -> String {
-    let mut quoted = String::with_capacity(text.len());
+    let mut quoted = String::new();
     escape_into(text, &mut quoted);
     quoted
 }
@@ -185,6 +185,7 @@ pub fn escape(text: &str) -> String {
 /// This will append escape characters into the given buffer. The characters
 /// that are appended are safe to use as a literal in a regular expression.
 pub fn escape_into(text: &str, buf: &mut String) {
+    buf.reserve(text.len());
     for c in text.chars() {
         if is_meta_character(c) {
             buf.push('\\');


hhirtz:
Sorry for the second push, added `syntax:` to the commit title.

pickfire:
@hhirtz doesn't it still needs to be allocated later from `escape_into` since `escape` uses it, why should we allocate twice?

hhirtz:
hi @pickfire, I don't really understand the first part of your question, but the patch here made `escape_into` performs less allocations on average.

For example, if you called `escape_into` with `String::new`, you'd likely end up doing more allocations than necessary, because `String` allocates on `push` when there's not enough space.  There's an example here: <https://doc.rust-lang.org/std/string/struct.String.html#representation>.

Since `s = String::with_capacity(cap)` is semantically the same as `s = String::new(); s.reserve(s)`, I removed the `with_capacity` in `escape`.

Also `escape_into(text, &mut buf)` writes between `text.len()` (no character needs an escape) and `text.len() * 2` (all characters are 1-byte codepoints, and they all need to be escaped*) characters in `buf`.  Here it reserves `text.len()` bytes in `buf` because we  know it needs at least `text.len()` more bytes in `buf`.

*Though I don't know any regex meta character that uses more than 1 byte in UTF-8.

pickfire:
@hhirtz Ah, thanks for the explaination, now I understand it. By the way, if we count first for exactly how many allocations we need, will allocating / reserving once be faster than potentially allocating multiple times?

hhirtz:
Yes, because in most cases reallocation is performed by allocating a larger buffer and copying the previous one into it.  If you know how much you want to `push`, it's better to call `reserve` before.

BurntSushi:
>  will allocating / reserving once be faster than potentially allocating multiple times?

Generally, yes. This is generally referred to as "amortizing allocation" and it greatly influences the structure of a lot of performance critical code. The entire architecture of the regex crate is designed around doing just that. :-)

PR #661, opened 2020-03-31
Add triagebot configuration

Mark-Simulacrum:
This enables [assignment](https://github.com/rust-lang/triagebot/wiki/Assignment) through triagebot on this repository, in preparation for the migration from highfive to triagebot for PR assignment.

cc rust-lang/highfive#258 rust-lang/triagebot#433


BurntSushi:
Thanks! But I don't think this repo needs this? I don't use bors/highfive/any-bots for this repo. (I did use bors once a long time ago, but it was so much more trouble than it was worth, given this repo's activity.)

Or maybe I'm missing something?

Mark-Simulacrum:
I'm going by the existence of https://github.com/rust-lang/highfive/blob/master/highfive/configs/rust-lang/regex.json, but if you don't expect `r?` to work on this repository, then feel free to just close :)

Mark-Simulacrum:
(Looking at the repository settings, highfive is configured as a webhook, so at least at some point we expected it to work).

BurntSushi:
Ah I see. Yeah, those web hooks haven't been used in a long time. In fact, none of those web hooks are used any more since we're now on GitHub Actions. So I've deleted all of them. Thanks for the heads up!

Issue #659, opened 2020-03-27
Panic "BUG: reverse match implies forward match" under certain conditions


qezz:
I've been trying to find an error in my small application, hence decided to utilize [AFL](https://github.com/rust-fuzz/afl.rs) to try different inputs to it. Application is based on a library that uses regex under the hood. 

So, with help of AFL I discovered that the application crashes in certain conditions. 
In short, the following test fails:

```rust
#[test]
fn weird_thing1() {
    // Note that 'Ј' is not 'j', but cyrillic Je 
    // https://en.wikipedia.org/wiki/Je_(Cyrillic)
    let regx = "()Ј01";
    let text = "()Ј01";
    let replace_with = "w";
    
    let re = regex::Regex::new(regx).unwrap();
    assert_eq!(re.replace(text, replace_with), "()w");
}
```

* **Expected**: Probably `text` as is. **Upd**: or `"()w"`
* **Actual**: Panic: `thread 'replace::weird_thing1' panicked at 'BUG: reverse match implies forward match', src/exec.rs:889:27`

Was able to reproduce only when parens and 'Je' symbol (https://en.wikipedia.org/wiki/Je_(Cyrillic)) are present, and followed by two or more characters.

See playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=53d71f4fa917ddba17d19630b86a81ac

---

One more weird thing:

I decided to add simple tests to `tests/replace.rs` to try it out, but faced the following:

1. When only one test is added, like
   ```rust
   replace!(weird_thing1, replace, r"()Ј01", "()Ј01", t!("w"), "()w");
   ```

   test case fails with (`$ cargo test replace`)
   ```
   ...
   test replace::trim ... ok
   test replace::weird_thing1 ... FAILED
   
   failures:
   
   ---- replace::weird_thing1 stdout ----
   thread 'replace::weird_thing1' panicked at 'BUG: reverse match implies forward match', src/exec.rs:889:27
   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
   
   
   failures:
       replace::weird_thing1

   test result: FAILED. 18 passed; 1 failed; 0 ignored; 0 measured; 780 filtered out
   ```
2. But when I add another one like
   ```rust
   replace!(weird_thing1, replace, r"()Ј01", "()Ј01", t!("w"), "()w"); // the first one
   replace!(weird_thing2, replace, r"()Ј01", "()Ј01", t!("w"), "something"); // a new one
[...]

PR #677, opened 2020-05-11
better support for empty patterns

sliquister:
I propose fixing the known issue in compile.rs where empty patterns + alternations can cause compilation errors or be compiled into a wrong automaton.

I understand that the plan is to replace compile.rs with a better version, but this doesn't seem to be about to happen, and I think the fix is not an undue amount of work. If possible, I would prefer fixing `regex` than work around these issues (it's easy to avoid them in hand written regexes but when generating regexes, it's more annoying).

About the change itself: all I do is compile `Empty` into `Ok(None)` instead of `Ok(Patch{ entry: the-next-instruction, ..})` (as that is not correct in general) and follow the type system. I tried fixing this in other ways with less diff, but I think this version is simpler: the diff is bigger (mostly due to changes types and indentation, which are easy to review) but the performance is clearly ok and the type system forces me to consider where changes need to happen. The kind of diff `self.c(expr)? ` -> `self.c(expr)?.unwrap_or(self.next_inst());` is saying "at this call site, the old behavior worked".

The same automation should be generated in the case that were previously correctly and successfully compiled, except a few corner cases that involve empty subexpressions, where we may generate fewer instructions (like `(?:)?`).


sliquister:
For the test failure:

```
thread 'hir::translate::tests::analysis_is_alternation_literal' panicked at 'assertion failed: t(r"").is_alternation_literal()', regex-syntax/src/hir/translate.rs:3127:9
```

should the documentation of hir and implementation of exec.rs be changed to accept Empty nodes in is_alternation_literal?

BurntSushi:
@sliquister Good question. The docs suggest that `""` shouldn't be an alternation literal, so I'd go with what you have and adjust the tests as appropriate. Updating the docs to add `""` to the list of negative examples would also be good.

sliquister:
Thanks, the tests all pass now (and I've learned of the existence of `cargo test --all`).

sliquister:
Have you had a chance to consider this? (I'm trying to decide if I need to workaround the underlying issue)

BurntSushi:
This PR is in `regex 1.3.8` on crates.io.

As a bonus, this also fixed #659!

sliquister:
Marvelous, thank you! (especially with this change being released immediately)

Issue #650, opened 2020-03-03
Consider integrating with OSS-Fuzz?


Valloric:
[OSS-Fuzz](https://google.github.io/oss-fuzz/) is a Google-provided service that can automatically and continuously fuzz OSS libraries and report bugs for discovered issues. It supports fuzzing Rust. 

[The requirement for being accepted](https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/) in OSS-Fuzz is "project must have a significant user base and/or be critical to the global IT infrastructure" which I _personally_ believe definitely includes the Rust regex library, since it's at the core of so much stuff. [List of projects that are currently in OSS-Fuzz.](https://github.com/google/oss-fuzz/tree/master/projects)

Can't hurt to file a form and ask.

(Disclaimer: I work for Google, but not on anything fuzzing or security-related. I do not speak for my employer. This suggestion comes from "personal Val", not "business Val". :smiley:)

Valloric:
Oh, forgot to mention: OSS-Fuzz offers [monetary rewards for integration](https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html), up to $20k. If this isn't personally interesting, it can be donated to a charity of your choice and Google will double the reward.

$40k to a charity for a few days of work doesn't sound like a bad deal...

BurntSushi:
Sounds like a good idea to me. Not sure if `regex` would qualify though or when I'll get to this. It's likely we could crib ideas from RE2: https://github.com/google/re2/tree/master/re2/fuzzing

If someone else wanted to work on this, I'd be happy to give guidance and mentor it.

miclill:
@Valloric What kind of bugs do you hope to catch by fuzzing? It is my understanding that - since this library written without the use of unsafe - most bugs caught by fuzzing can not occur anyway. 

For reference: https://opensource.googleblog.com/2017/05/oss-fuzz-five-months-later-and.html

BurntSushi:
This was done in https://github.com/rust-lang/regex/pull/679

@miclill Safety bugs are not the only kind of bugs. Fuzzing is very useful for sussing out panics which should not occur.

BurntSushi:
@DavidKorczynski The fuzzer reported its first bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=26324

[...]

PR #679, opened 2020-05-14
Added fuzzing infrastructure for OSS-Fuzz. Fuzzer similar to cargo-fu…

DavidKorczynski:
…zz targets.

### DIFF
--- a/fuzz/fuzz_targets/fuzz_regex_match.rs
+++ b/fuzz/fuzz_targets/fuzz_regex_match.rs
@@ -2,13 +2,16 @@
 use libfuzzer_sys::fuzz_target;
 
 fuzz_target!(|data: &[u8]| {
-    if let Ok(data) = std::str::from_utf8(data) {
-        // split data into regular expression and actual input to search through
+    if data.len() < 2 {
+        return;
+    }
+    let split_point = data[0] as usize;
+    if let Ok(data) = std::str::from_utf8(&data[1..]) {
         use std::cmp::max;
+        // split data into regular expression and actual input to search through
         let len = data.chars().count();
-        let split_off_point = max(len / 5, 1) as usize;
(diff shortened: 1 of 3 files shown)


DavidKorczynski:
Cr https://github.com/rust-lang/regex/issues/650

fuzz/fuzz_targets/fuzz_regex_match.rs
```diff
@@ -0,0 +1,19 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    if let Ok(data) = std::str::from_utf8(data) {
+        // split data into regular expression and actual input to search through
+        use std::cmp::max;
+        let len = data.chars().count();
+        let split_off_point = max(len / 5, 1) as usize;
```
ethanpailes, line 9: What if this split point was determined by the first byte or two of the data so the fuzzer the regex and input size were not locked to one another?

DavidKorczynski: Good idea - how does the new one look?

DavidKorczynski:
Was wondering if we could land this @ethanpailes, @BurntSushi - all seems to be ready from OSS-Fuzz side of things.

fuzz/fuzz_targets/fuzz_regex_match.rs
ethanpailes: looks good to me, but it's only @BurntSushi's opinion that matters here.

DavidKorczynski:
ping - any updates here @BurntSushi ?

BurntSushi:
@DavidKorczynski Sorry for dragging my feet on this. I merged this PR in #718. What are the next steps?

Issue #602, opened 2019-07-24
Invalid capture group name doesn't fail


crides:
Example code:

```rust
use regex::Regex;

fn main() {
    let re = Regex::new(r"a(.*)f").unwrap();
    println!("{}", re.replace("asdf", "<$1a>"));
}
```

This prints "<>". However, "1a" isn't a valid capture group name (i.e. if you use it like `(?P<1a>...)` it would `panic`). When used in a replace expression like the one above, the name "1a" is accepted by the parser (which in theory it shouldn't), and doesn't cause an error to be returned or `panic`.

The ideal way to solve this is make the parser only accept `\$\d+` or `\$\w[\w\d]*` as valid capture groups.

This originally comes from [chmln/sd#44](https://github.com/chmln/sd/issues/44).

BurntSushi:
Thanks for the report.

For clarity, using `(?P<1a>...)` does not _panic_, but rather, it returns an error. This is a critical difference, because if it panicked, it would be a bug. But it does not.

> When used in a replace expression like the one above, the name "1a" is accepted by the parser (which in theory it shouldn't), and doesn't cause an error to be returned or panic.

A replacement never panics and never returns an error. The only possible behaviors we have here for an invalid capture group name such as `1a` are "substitute with empty text" or "don't substitute at all." It does indeed look like the behavior today is the former. As far as I can tell, this is actually consistent with [documented behavior](https://docs.rs/regex/1.2.0/regex/struct.Captures.html#method.expand):

> If name isn't a valid capture group (whether the name doesn't exist or isn't a valid index), then it is replaced with the empty string.

Now, there is enough wiggle room in the language here where we might be justified in changing how invalid capture group names are handled, but the docs even go on to give an example using `1a` itself:

> The longest possible name is used. e.g., `$1a` looks up the capture group named `1a` and not the capture group at index `1`. To exert more precise control over the name, use braces, e.g., `${1}a`.

It is a bit strange for the example to use an invalid capture group name, but changing the behavior to be in direct contradiction to a documented example is probably a bad idea.

I'm going to mark this issue for `regex 2` for now. We might choose to tackle this in other ways, e.g., by perhaps expanding the set of allowable capture names. I kept it super conservative to start with.

crides:
Thanks for your detailed response! My bad not to dig into the docs. But this still kind of confuses me:

> The longest possible name is used. e.g., `$1a` looks up the capture group named `1a` and not the capture group at index 1. 

[...]

PR #690, opened 2020-06-08
Less greedy named capture groups

chmln:
Hi @BurntSushi 

I've forked `regex` to change capture group matching a bit. In particular, this fixes unexpected behavior like https://github.com/chmln/sd/issues/44, and #602 partially.

This changes the matching so that `'$1foo'` is parsed as `'$1' and 'foo'` instead of the variable `$1foo`. Since named capture groups cannot begin with a number anyways, this is a common-sense change to me.

I'm submitting this pull request in case there's interest in this behavior. And of course, upstreaming the fix is obviously preferable to maintaining a fork.


BurntSushi:
There is an issue open for this: #69

It's marked as `regex2` because it's a breaking change. Arguably, we might be able to get away with a change like this, but I'd prefer to be conservative about such things.

> And of course, upstreaming the fix is obviously preferable to maintaining a fork.

You really do not need to maintain a fork. That's a huge amount of overkill. All you need to do is implement your own replacement routine. Copying out the replacement text parser from the regex crate, make your changes and implement a `replace` function with it should be pretty easy. Certainly much easier than maintaining a published fork.

PR #710, opened 2020-09-30
compile: Use std::convert::TryFrom to convert u32 to usize

arnej:
The interface and behavior doesn't change because we're using expect to
panic when try_from fails.


BurntSushi:
Thanks, but this isn't worth bumping the MSRV.

I do hope to bump the MSRV at some point, as 1.28 is quite old, but it will come with a bigger batch of changes I think. And I'll still try to pick a Rust version that is at least 6 months old.

PR #720, opened 2020-10-14
Update fuzzer cargo.toml

DavidKorczynski:
I had to apply these changes for OSS-Fuzz to continue working. Specifically I needed to add `cargo-features = ['named-profiles']` and `inherits` for any profile other than `release`. However, I have also confirmed now that the building is done with optimized + debuginfo in OSS-Fuzz (`Finished release [optimized + debuginfo] target(s) in 37.59s`).

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,3 +1,5 @@
+cargo-features = ['named-profiles']
+
 [package]
 name = "regex-fuzz"
 version = "0.0.0"
@@ -22,14 +24,12 @@ members = ["."]
 name = "fuzz_regex_match"
 path = "fuzz_targets/fuzz_regex_match.rs"
 
-[profile.debug]
-opt-level = 3
-debug = true
-
 [profile.release]
 opt-level = 3
 debug = true
 
+[profile.debug]
+inherits = "release"
+
 [profile.test]
-opt-level = 3
-debug = true
+inherits = "release"


BurntSushi:
Thanks! Also, could you share the commands you're running with `cargo fuzz` to test this?

And I wonder, if `cargo fuzz` was supposed to use release mode by default, then maybe I mis-diagnosed the report from OSS-Fuzz?

DavidKorczynski:
Here you will find the set up of rust-regex on OSS-Fuzz: https://github.com/google/oss-fuzz/tree/master/projects/rust-regex

In the `build.sh` file, you will see the specific commands used to build the fuzzers, which is just `cargo fuzz build -O` in this case:
https://github.com/google/oss-fuzz/blob/2cd2ee63d7d98c71d1682cbed39b851050b696a7/projects/rust-regex/build.sh#L18

If you would like to replicate the exact set up of OSS-Fuzz, the easiest way is to install docker, and then run the following commands:
```
git clone https://github.com/google/oss-fuzz
cd oss-fuzz
python3 infra/helper.py build_image rust-regex
python3 infra/helper.py build_fuzzers rust-regex
```
(Note you may have to run the commands with `sudo` depending on your Docker installation)
This will put the fuzzer of the project at the path `./build/out/rust-regex/fuzz_regex_match`

You can either run the fuzzer by simply running `./build/out/rust-regex/fuzz_regex_match` or by going through `helper.py` as follows:
```
python3 infra/helper.py run_fuzzer rust-regex fuzz_regex_match
```

PR #724, opened 2020-11-24
Update Computer Language Benchmarks Game URL

atouchet:
This URL appears to have moved since these were last edited.

### DIFF
--- a/examples/shootout-regex-dna-bytes.rs
+++ b/examples/shootout-regex-dna-bytes.rs
@@ -1,5 +1,5 @@
 // The Computer Language Benchmarks Game
-// http://benchmarksgame.alioth.debian.org/
+// https://benchmarksgame-team.pages.debian.net/benchmarksgame/
 //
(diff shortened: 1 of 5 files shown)


evertondanilo:
@BurntSushi Did you participate in Rust coding on CLBG? I'd like to port Rust to Wasm and run some performance tests.

BurntSushi:
@evertondanilo I don't understand what you're asking and what it has to do with this PR. Please say more words.

evertondanilo:
Hi @BurntSushi thanks for your quick reply. I'm working on a research project for my masters. The idea is to compare Wasm with JS in the browser. And find out how fast it would be? Do you think CLBG algs can be compiled to Wasm?

BurntSushi:
I don't know why you're asking me that question on the issue tracker for a regex library.

> Do you think CLBG algs can be compiled to Wasm?

Why not try it?

I think this line of questioning is way off topic for this tracker. Please take it somewhere else.

PR #728, opened 2020-12-02
implement `regex::Replacer` for `String`, `&String`, `Cow<'a, str>`, `&Cow<'a, str>`

nooberfsh:
It is inconvenient to get `&str` in some cases, for example:
```rust
extern crate regex; // 1.4.2
extern crate chrono; // 0.4.19

fn main() {
    let input = "today is time";
    let today = chrono::Local::now();
    let re = regex::Regex::new(r"time").unwrap();
    let s = re.replace(input, &*format!("{}", today)); // <-- I have to use &*format(...)
    println!("{}", s);
}
```

### DIFF
--- a/src/re_bytes.rs
+++ b/src/re_bytes.rs
@@ -1148,10 +1148,55 @@ impl<'a> Replacer for &'a [u8] {
     }
 
     fn no_expansion(&mut self) -> Option<Cow<[u8]>> {
-        match find_byte(b'$', *self) {
-            Some(_) => None,
-            None => Some(Cow::Borrowed(*self)),
-        }
+        no_expansion(self)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


BurntSushi (changes requested):
Thanks! I think this LGTM, but I'd like to see two changes before merging it:

1. Update the docs on the `Replacer` trait. Right now, it calls out the `&str` impl specifically, but now we've added more. I don't think we should list out every impl, but maybe say, "implemented for `&str` along with other variants of string types." Or something like that.
2. Please also add the analogous trait impls in `src/re_bytes.rs`. It seems weird to offer these extra impls for only the `&str` API.

nooberfsh:
@BurntSushi  updated.

BurntSushi (approved):
LGTM, thanks!

PR #729, opened 2020-12-05
PERFORMANCE: update syntax to Rust 2018

ifraixedes:
Update the code snippet of the PERFORMANCE document to use the Rust 2018
module system features.
See:
https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html

### DIFF
--- a/PERFORMANCE.md
+++ b/PERFORMANCE.md
@@ -62,9 +62,7 @@ on how your program is structured. Thankfully, the
 [`lazy_static`](https://crates.io/crates/lazy_static)
 crate provides an answer that works well:
 
-    #[macro_use] extern crate lazy_static;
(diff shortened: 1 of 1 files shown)


BurntSushi:
The `regex` crate hasn't migrated to Rust 2018 yet. Once it does, then we can change the doc materials. But until then I'd rather not be in a tweener state.

ifraixedes:
Oh, I'm sorry! I didn't know about that.
I'm a newbie in Rust and when I wanted to use this crate I found that that the syntax in the doc didn't work and I had to find out to use this one and I discovered that it was a change in Rust 2018, hence I thought that it was forgotten to be updated.

My apologies for the trouble.

PR #734, opened 2020-12-29
Add missing implementations of core iterator traits

lopopolo:
- Add missing implementations of FusedIterator.
- Add missing implementations of ExactSizeIterator.
- Add missing overrides of `Iterator::count` when wrapping a slice iterator.
- Add missing overrides of `Iterator::size_hint` when size hint is known.

Context: I'm building an implementation of Ruby `Regexp` on top of the `regex` crate, and having guarantees around whether iterators are fused will allow me to implement these traits on my wrapping `Iterator`s as well.

### DIFF
--- a/src/re_bytes.rs
+++ b/src/re_bytes.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::fmt;
+use std::iter::FusedIterator;
 use std::ops::{Index, Range};
 use std::str::FromStr;
 use std::sync::Arc;
@@ -701,6 +702,8 @@ impl<'r, 't> Iterator for Matches<'r, 't> {
     }
 }
 
+impl<'r, 't> FusedIterator for Matches<'r, 't> {}
+
 /// An iterator that yields all non-overlapping capture groups matching a
 /// particular regular expression.
 ///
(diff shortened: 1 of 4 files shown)


src/dfa.rs
```diff
@@ -301,6 +301,8 @@ impl<'a> Iterator for InstPtrs<'a> {
     }
 }
 
+impl<'a> FusedIterator for InstPtrs<'a> {}
+
```
BurntSushi, line 305: I don't think `InstPtrs` is exposed anywhere, so I don't think we need to impl `FusedIterator` for it?

src/literal/imp.rs
```diff
@@ -286,6 +287,8 @@ impl<'a> Iterator for LiteralIter<'a> {
     }
 }
 
+impl<'a> FusedIterator for LiteralIter<'a> {}
```
BurntSushi, line 290: Same here as with `InstPtrs`?

BurntSushi (changes requested):
LGTM! I think there are a couple of impls we don't need though?

lopopolo:
@BurntSushi I've removed the impls on internal types you flagged.

BurntSushi (approved):
Thanks!

PR #735, opened 2020-12-29
Missing fmt::Debug impls for public structs and enums

lopopolo:
The Rust API guidelines advise that [all public types implement common `std` traits for interoperability](https://rust-lang.github.io/api-guidelines/interoperability.html), of which `fmt::Debug` is one.

This PR adds `fmt::Debug` implementations for all public types in `regex` and `regex-syntax` crates. This PR also adds `#![warn(missing_debug_implementations)]` to `lib.rs` in each crate.

This PR also opportunistically implements `Clone` on iterators that wrap `slice::Iter`.

Context: I'm building an implementation of Ruby `Regexp` on top of the `regex` crate, and having all public types implement `fmt::Debug` will allow me to `#[derive(Debug)]` on types that wrap e.g. iterators in this crate.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::fmt;
 use std::iter;
 use std::result;
 use std::sync::Arc;
@@ -25,6 +26,9 @@ struct Patch {
 
 /// A compiler translates a regular expression AST to a sequence of
 /// instructions. The sequence of instructions represents an NFA.
+// `Compiler` is only public via the `internal` module, so avoid deriving
+// `Debug`.
+#[allow(missing_debug_implementations)]
 pub struct Compiler {
     insts: Vec<MaybeInst>,
     compiled: Program,
@@ -1051,6 +1055,7 @@ impl<'a, 'b> CompileClass<'a, 'b> {
 /// This uses similar idea to [`SparseSet`](../sparse/struct.SparseSet.html),
 /// except it uses hashes as original indices and then compares full keys for
 /// validation against `dense` array.
+#[derive(Debug)]
 struct SuffixCache {
     sparse: Box<[usize]>,
     dense: Vec<SuffixCacheEntry>,
@@ -1156,6 +1161,12 @@ impl ByteClassSet {
             i += 1;
         }
         byte_classes
+    }
+}
+
+impl fmt::Debug for ByteClassSet {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("ByteClassSet").field(&&self.0[..]).finish()
     }
 }
 
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -30,6 +30,7 @@ use utf8::next_utf8;
 /// In particular, this manages the various compiled forms of a single regular
 /// expression and the choice of which matching engine to use to execute a
 /// regular expression.
+#[derive(Debug)]
 pub struct Exec {
     /// All read only state.
     ro: Arc<ExecReadOnly>,
@@ -49,6 +50,7 @@ pub struct ExecNoSync<'c> {
 }
 
 /// `ExecNoSyncStr` is like `ExecNoSync`, but matches on &str instead of &[u8].
+#[derive(Debug)]
 pub struct ExecNoSyncStr<'c>(ExecNoSync<'c>);
 
 /// `ExecReadOnly` comprises all read only state for a regex. Namely, all such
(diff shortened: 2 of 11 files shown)


src/dfa.rs
```diff
@@ -279,6 +279,7 @@ fn push_inst_ptr(data: &mut Vec<u8>, prev: &mut InstPtr, ip: InstPtr) {
     *prev = ip;
 }
 
+#[derive(Debug)]
```
BurntSushi, line 282: I don't think this is needed?

src/re_trait.rs
```diff
@@ -172,6 +174,21 @@ where
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Matches")
+            .field("re", &self.re)
+            .field("text", &self.text)
+            .field("last_end", &self.last_end)
+            .field("last_match", &self.last_match)
+            .finish()
+    }
+}
```
BurntSushi, line 190: I suspect you're writing out this impl because `Matches` is generic and doesn't require its type parameters to impl `Debug`? The `RegularExpression` trait is an internal construct, so adding a requirement to impl `Debug` seems fine to me. (For both `Self` and for its `Text` associated type.) Then you should be able to derive `Debug` impls here.

src/exec.rs
```diff
@@ -106,6 +109,7 @@ pub struct ExecBuilder {
 
 /// Parsed represents a set of parsed regular expressions and their detected
 /// literals.
+#[derive(Debug)]
```
BurntSushi, line 112: Is this `Debug` impl necessary?

src/exec.rs
```diff
@@ -97,6 +99,7 @@ struct ExecReadOnly {
 /// Facilitates the construction of an executor by exposing various knobs
 /// to control how a regex is executed and what kinds of resources it's
 /// permitted to use.
+#[derive(Debug)]
```
BurntSushi, line 102: Is this `Debug` impl necessary?

src/compile.rs
```diff
@@ -25,6 +26,7 @@ struct Patch {
 
 /// A compiler translates a regular expression AST to a sequence of
 /// instructions. The sequence of instructions represents an NFA.
+#[derive(Debug)]
```
BurntSushi, line 29: Is this `Debug` impl necessary?

BurntSushi (approved):
LGTM, but like the other PR, my sense is that some of these impls aren't needed to pass the lint? Or does the lint apply to internal types too?

The reason why I'm being a stickler about this is that I'd rather not add impls for things that we don't need because they likely increase compile times. And the compile times for this crate are already pretty poor.

BurntSushi (changes requested):
Whoops, I meant to "request changes."

src/exec.rs
lopopolo: `ExecBuilder` is public via the `internal` module which "exists to support suspicious activity".

I'll remove the debug impl and allow the lint.

src/compile.rs
lopopolo: `Compiler` is public via the internal module which "exists to support suspicious activity".

I'll remove the debug impl and allow the lint.

BurntSushi: Haha right. Yeah, it's technically public, but is `doc(hidden)` and not a stable part of the API. I'm hoping to remove that quirk in the move to `regex-automata`.

Thanks.

src/re_trait.rs
lopopolo: I also made the same change to `CaptureMatches`.

lopopolo:
@BurntSushi I've made the requested changes. I squashed these into the commit for the `regex` crate.

For types that are public via the hidden `internal` module, I've added an `#[allow(missing_debug_implementations)]` and avoid implementing `Debug`.

lopopolo:
rebase incoming.

lopopolo:
> LGTM, but like the other PR

For reference, the other PR is #734.

BurntSushi (approved):
OK, LGTM, thanks!

lopopolo:
Whoops looks like I accidentally didn't split out the iterator traits bits from `regex-syntax` properly when splitting apart the two PRs.

Thanks for merging.

Would it be possible to get these two PRs packaged into a release soon?

BurntSushi:
Yeah I saw that, but that's okay. No biggie.

I'll try to do a release in the next couple days. Please ping me again if it doesn't happen.

lopopolo:
hi @BurntSushi following up with a friendly ping for a release.

BurntSushi:
@lopopolo OK, `regex 1.4.3` is on crates.io.

lopopolo:
Thank you!

PR #741, opened 2021-01-21
doc: Use "text" instead of "ignore" attribute for regex examples

markusolt:
Replaces the "ignore" attribute with the "text" attribute for all documentation comment code blocks that contain regular expressions. This removes the syntax highlighting and removes the warning "This example is not tested". The code blocks will instead render without any syntax highlighting or warnings.

See for example the second code block on [this page](https://docs.rs/regex/1.4.3/regex/struct.RegexSet.html#example) or the code blocks on [this page](https://docs.rs/regex-syntax/0.6.22/regex_syntax/utf8/index.html#wait-what-is-this) of the `regex-syntax` crate.


BurntSushi (approved):
Nice catch, thanks!

PR #752, opened 2021-03-14
impl: substantially reduce regex stack size

### LINKED ISSUE #750: stackoverflow in 1.4.4

If it isn't the latest version, then please upgrade and check whether the bug
is still present.

1.4.4 breaks [windows build](https://github.com/tikv/grpc-rs/runs/2099128323?check_suite_focus=true) of grpcio.

Just run `cargo test --all` on Windows.

Build fails when generating bindings using rust-bindgen, which uses regex to process sources.

When downgrade regex back to before e040c1b06397a254cccd3506ee80dbe042360afd, the build can finish successfully, so the bug is probably introduced by https://github.com/rust-lang/regex/pull/749.

It compiles successfully.

What do you expect the output to be?

### LINKED ISSUE #751: 1.44 triggers a stack overflow on Windows

1.44

Running bindgen as part of the [mozjs](https://github.com/servo/mozjs) build script triggers a stack overflow on Windows.

N.B. I recognize that from a reproducibility and isolation perspective, this is the worst possible testcase. I'm filing this so I don't lose track of it.

Build Servo in a Windows CI environment. Using regex 1.43 it completes, and 1.44 it encounters a stack overflow. More details on this in https://github.com/servo/servo/pull/28265.

I have not yet been able to reproduce this only building mozjs using github actions.

I verified that https://github.com/rust-lang/regex/commit/e040c1b06397a254cccd3506ee80dbe042360afd is the commit that triggers this change in behaviour in https://github.com/servo/servo/pull/28269.

BurntSushi:
This commit fixes a fairly large regression in the stack size of a Regex
introduced in regex 1.4.4. When I dropped thread_local and replaced it
with Pool, it turned out that Pool inlined a T into its struct and a
Regex in turn had Pool inlined into itself. It further turns out that
the T=ProgramCache is itself quite large.

We fix this by introducing an indirection in the inner regex type. That
is, we use a Box<Pool> instead of a Pool. This shrinks the size of a
Regex from 856 bytes to 16 bytes.

Interestingly, prior to regex 1.4.4, a Regex was still quite substantial
in size, coming in at around 552 bytes. So it looks like the 1.4.4
release didn't dramatically increase it, but it increased it enough that
folks started experiencing real problems: stack overflows.

Since indirection can lead to worse locality and performance loss, I did
run the benchmark suite. I couldn't see any measurable difference. This
is generally what I would expect. This is an indirection at a fairly
high level. There's lots of other indirection already, and this
indirection isn't accessed in a hot path. (The regex cache itself is of
course used in hot paths, but by the time we get there, we have already
followed this particular pointer.)

We also include a regression test that asserts a Regex (and company) are
16 bytes in size. While this isn't an API guarantee, it at least means
that increasing the size of Regex will be an intentional thing in the
future and not an accidental leakage of implementation details.

Fixes #750, Fixes #751

Ref https://github.com/servo/servo/pull/28269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+1.4.5 (2021-03-14)
+==================
+This is a small patch release that fixes a regression in the size of a `Regex`
+in the 1.4.4 release. Prior to 1.4.4, a `Regex` was 552 bytes. In the 1.4.4
+release, it was 856 bytes due to internal changes. In this release, a `Regex`
+is now 16 bytes. In general, the size of a `Regex` was never something that was
+on my radar, but this increased size in the 1.4.4 release seems to have crossed
+a threshold and resulted in stack overflows in some programs.
+
+* [BUG #750](https://github.com/rust-lang/regex/pull/750):
+  Fixes stack overflows seemingly caused by a large `Regex` size by decreasing
+  its size.
+
+
 1.4.4 (2021-03-11)
 ==================
 This is a small patch release that contains some bug fixes. Notably, it also
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -36,7 +36,14 @@ pub struct Exec {
     /// All read only state.
     ro: Arc<ExecReadOnly>,
     /// A pool of reusable values for the various matching engines.
-    pool: Pool<ProgramCache>,
+    ///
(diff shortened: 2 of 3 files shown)


BurntSushi:
This PR is on crates.io in `regex 1.4.5`.

PR #759, opened 2021-04-08
Fix breaking OSS-Fuzz coverage build

DavidKorczynski:


DavidKorczynski:
Apologies, this is a duplicate. I was going through fixing other projects and didn't realise this was one I had already fixed.

PR #760, opened 2021-04-13
Shrink size of `Inst`

marmeladema:
By using a boxed slice instead of a vector, we can shrink the size
of the `Inst` structure by 8 bytes going from 40 to 32 bytes.


BurntSushi (approved):
Thanks! I love it.

Issue #656, opened 2020-03-24
plan: moving regex engines to regex-automata


BurntSushi:
In discussions in #524, it became clearer to me that it might be good to say more about what my short/long term plans are for the `regex` crate.

TL;DR - Push all regex engines down into the [`regex-automata`](https://github.com/BurntSushi/regex-automata) crate, and bring that in as an internal-but-published dependency of `regex` as an official part of this repository. It will become a second required dependency (with the first being `regex-syntax`).

The main motivation for doing this is to clean up the regex internals, specifically with the intent to make it easier to add more optimizations. Currently, the infrastructure is a bit of a hodge podge and it can be quite difficult to know where and when a particular regex engine is being used. On top of that, since every regex engine is strictly internal, testing each engine independently is quite annoying, as it requires exporting undocumented APIs. As a result, bugs have cropped up where certain engines are inconsistent with other engines. Similarly, benchmarking each engine independent of the other is made difficult, which in turn makes it difficult to focus on optimizing certain scenarios that come up in the course of regex execution.

On top of all of this, putting the regex engines into their own crate with a proper API means that folks can use those engines directly. It also provides an outlet of sorts to provide lower level APIs that would I otherwise wouldn't want cluttering up the main `regex` crate too much. The main thing on my mind here are streaming APIs (see #425), but there may be other things of use.

I've actually been working towards this goal for a long time now. Arguably, i started around 3 years ago when I began rewriting the `regex-syntax` crate. I then moved to rewriting `memchr` and `aho-corasick` as well. In the process, I pushed complexity down as much as I could. For example, the HIR in `regex-syntax` no longer exposes any regex flags (such as case insensitivity), which makes regex compilers simpler. As another example, `aho-corasick` now supports leftmost-first matching, which matches the regex crate's match semantics precisely, and now makes it easier to defer to `aho-corasick` for an alternation of literals. Yet another example is the Teddy algorithm that uses SIMD for finding matches of multiple literals very quickly. That has been pushed down into `aho-corasick` as well, which means more people get to benefit from it and it no longer needs to be maintained inside of `regex` proper.

A little over a year ago, I started work on [`regex-automata`](https://github.com/BurntSushi/regex-automata). The main use case for `regex-automata` was to generate fully compiled DFAs that could be cheaply embedded into a program and deserialized cheaply for easy matching. (For example, they power the grapheme, word and sentence iterators in the [`bstr` crate](https://github.com/BurntSushi/bstr).)

[...]

PR #761, opened 2021-04-14
Split `Inst` enum into `BytesInst` and `UnicodeInst` enums

marmeladema:
While trying to understand better the internals of the crate, I stumbled upon this comment:
```Rust
/// Regrettably, a regex program either contains Unicode codepoint
/// instructions (Char and Ranges) or it contains byte instructions (Bytes).
/// A regex program can never contain both.
///
/// It would be worth investigating splitting this into two distinct types and
/// then figuring out how to make the matching engines polymorphic over those
/// types without sacrificing performance.
///
/// Other than the benefit of moving invariants into the type system, another
/// benefit is the decreased size. If we remove the `Char` and `Ranges`
/// instructions from the `Inst` enum, then its size shrinks from 40 bytes to
/// 24 bytes. (This is because of the removal of a `Vec` in the `Ranges`
/// variant.) Given that byte based machines are typically much bigger than
/// their Unicode analogues (because they can decode UTF-8 directly), this ends
/// up being a pretty significant savings.
```

This is what this PR proposes to implement by making `Program` generic over the instruction type: `Program<I>`.
All tests seem should pass but benchmarks have not been updated yet.

cc @BurntSushi, I hope this will be useful :+1:

### DIFF
--- a/src/backtrack.rs
+++ b/src/backtrack.rs
@@ -18,7 +18,7 @@
 
 use exec::ProgramCache;
 use input::{Input, InputAt};
-use prog::{InstPtr, Program};
+use prog::{BytesInst, InstPtr, InstTrait, Program, UnicodeInst};
 use re_trait::Slot;
 
 type Bits = u32;
@@ -41,8 +41,8 @@ pub fn should_exec(num_insts: usize, text_len: usize) -> bool {
 
 /// A backtracking matching engine.
 #[derive(Debug)]
-pub struct Bounded<'a, 'm, 'r, 's, I> {
-    prog: &'r Program,
+pub struct Bounded<'a, 'm, 'r, 's, I, P: InstTrait> {
+    prog: &'r Program<P>,
     input: I,
     matches: &'m mut [bool],
     slots: &'s mut [Slot],
@@ -59,7 +59,7 @@ pub struct Cache {
 
 impl Cache {
     /// Create new empty cache for the backtracking engine.
-    pub fn new(_prog: &Program) -> Self {
+    pub fn new<I: InstTrait>(_prog: &Program<I>) -> Self {
         Cache { jobs: vec![], visited: vec![] }
     }
 }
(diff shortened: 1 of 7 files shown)


marmeladema:
Well, tests are passing with rust `1.51.0` but not with `1.28.0` because of lifetime inference issues. This should be fixable if the PR is worthwhile.

BurntSushi:
@marmeladema I apologize for leaving that comment and suggesting that I would be okay with a PR like this. :-( In the future, for bigger changes, I do strongly advise that you file an issue first.

I really appreciate the effort towards improving memory usage here. But I'm going to close this out. The primary reason is that #656 is in progress right now, and part of that effort involves a rewrite of the NFA compiler. This comes with a new [representation for `Inst`](https://github.com/BurntSushi/regex-automata/blob/153159bdc6e100414c14fc702655355631aeb885/src/nfa/thompson/mod.rs#L290-L317). In that representation, at _currently_, there is no "Unicode" variant. Everything is just bytes. But, there are also new variants that have `Box<[...]>` in them, which inflates their size anyway.

While the existence of #656 doesn't mean we can't also accept improvements in the current code, I think this particular change is quite complex. I regret being so flippant with making everything polymorphic in my comment. In particular, while this undoubtedly improves memory usage, it's quite likely to also increase compile times and bloat binaries. So even if #656 weren't a factor here, there are still possible reasons not to do this I think.

Anyway, thanks again and I regret that you put so much effort into a PR that I'm not going to consider. Would definitely recommend filing an issue first. Or if you want to reach out on Zulip, that would be great too.

marmeladema:
@BurntSushi thank you for your detailed answer :+1: regex-automata looks very promising

It is true that I have spent quite some time on this, but the real goal was to understand better the architecture of the codebase and learn about the internals of the crate. I think it helped me a lot to do this even if not merged in the end, so all is not lost.

The actual reason of why I am digging into this in the first place is that I am maintaining a production system which uses a lot of regexes and it seems that they can take a substantial amount of memory.

I'll try to reach out to you on zulip to better explain my use cases if that's ok.

BurntSushi:
@marmeladema Sounds good!

PR #764, opened 2021-04-22
fuzz: account for Unicode class size in compiler

BurntSushi:
This improves the precision of the "expression too big" regex
compilation error. Previously, it was not considering the heap usage
from Unicode character classes.

It's possible this will make some regexes fail to compile that
previously compiled. However, this is a bug fix. If you do wind up
seeing this though, feel free to file an issue, since it would be good
to get an idea of what kinds of regexes no longer compile but did.

This was found by OSS-fuzz:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33579

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+1.4.6 (2021-04-22)
+==================
+This is a small patch release that fixes the compiler's size check on how much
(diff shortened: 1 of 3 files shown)


BurntSushi:
This PR is on crates.io in `regex 1.4.6`.

PR #767, opened 2021-04-30
update to latest memchr + upgrade to Rust 2018 + bump MSRV to Rust 1.41

BurntSushi:
The main motivation for this PR is to use the new memmem implementation in `memchr 2.4` (not quite released at time of writing, but [in a PR](https://github.com/BurntSushi/memchr/pull/82)). This let's us delete regex's own bespoke substring search implementations ("FreqyPacked" along with Boyer-Moore). The main benefit of the new implementation is that it should roughly maintain the speed of the old algorithms, but keep its speed in a lot more cases. i.e., It should have far fewer weaknesses. Plus, the algorithm is now available for anyone to use without bringing in `regex`.

While we're here, we (finally) move to Rust 2018 and bump the MSRV to Rust 1.41 (since that's what's in Debian Stable). There's no particular reason why I waited so long to do this. It was never my intent to support such an old version of Rust for so long. There was just never a strong impetus to upgrade. But with Rust 2021 around the bend, it seems appropriate to at least migrate to Rust 2018. Hopefully we'll get to Rust 2021 sooner.

(The plan is to merge this PR once I do a similar change to the `aho-corasick` crate.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         include:
         - build: pinned
           os: ubuntu-18.04
-          rust: 1.28.0
+          rust: 1.41.1
         - build: stable
           os: ubuntu-18.04
           rust: stable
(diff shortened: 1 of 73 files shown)


BurntSushi:
This PR is on crates.io in `regex 1.5.0`.

PR #772, opened 2021-05-05
Fix compilation with pattern feature

Luro02:


test
```diff
@@ -21,6 +21,7 @@ features=(
     "std perf-dfa"
     "std perf-inline"
     "std perf-literal"
+    "std pattern"
```
BurntSushi, line 24: Unfortunately, this script is only run on stable Rust. So I think you'll have to add

```
cargo test --test default --no-default-features --features 'std pattern'
```

as an explicit step in CI on nightly targets only.

BurntSushi (changes requested):
Thanks! It looks like the CI setup might need to be tweaked a bit though.

test
Luro02: done

BurntSushi:
Looks like the `unicode-perl` feature is needed to run the searcher tests.

GitHub Actions CI is FUBAR, so it doesn't run until I click a button. Please run the tests locally and make sure it works there before pushing.

BurntSushi (approved):

BurntSushi:
This PR is on crates.io in `regex 1.5.4`.

PR #773, opened 2021-05-05
fix rustc version in badge

striezel:
MSRV has been bumped to 1.41.1 in regex 1.5.0 / commit 832ba73, so this should be reflected in the readme, too.


BurntSushi (approved):
Thanks!

> so this should be reflected in the readme, too

Small clarification: it is, it's just in its own section towards the bottom. I just forgot to update the badge.

PR #776, opened 2021-05-05
Add a lifetime to Replacer

sgrif:
The current implementation requires any implementations of `Replacer` be
valid for all possible lifetimes. However, this is neither required nor
is it always desirable. For example, a user who wishes to use the
lifetime of the input text as some part of the return type when
interacting with these APIs would end up with a closure that is valid
for one specific lifetime.

A concrete example of where this was problematic was described in
https://github.com/rust-lang/regex/discussions/775

Any future code that does require a `Replacer` which is valid for all
lifetimes can do so with HRTB:

```
where
    for<'a> R: Replacer<'a>
```

or `R: Replacer<'_>` in some contexts

### DIFF
--- a/src/re_unicode.rs
+++ b/src/re_unicode.rs
@@ -503,7 +503,7 @@ impl Regex {
     /// assert_eq!(result, "$2 $last");
     /// # }
     /// ```
-    pub fn replace<'t, R: Replacer>(
+    pub fn replace<'t, R: Replacer<'t>>(
         &self,
         text: &'t str,
         rep: R,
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks so much! Yeah, the breaking changes means this can't land until `regex 2.0.0`. I filed #777 to track this. It links to this PR, so I can bring this patch in whenever that happens. But I'll close this for now. :-) Thanks again!

sgrif:
No worries. Glad I could help <3

PR #778, opened 2021-05-11
re_builder: add throwaway_on_parse_error flag

keur:
This flag is exposed to the RegexSetBuilder when the caller wants
malformed regexes thrown out. It is useful for dealing with user
provided regexes. Disabled by default because in most cases people want
this to fail loudly.

An example use case is:

```
   let set = RegexSetBuilder::new(&[
        String::from(r"foo"),
        String::from(r"bar"),
        String::from(r#"/^\=/"#), // this regex doesn't compile!
    ]).throwaway_on_parse_error(true).build().unwrap();
    assert!(set.is_match("foobar"));
```

### DIFF
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -246,8 +246,13 @@ impl ExecBuilder {
                 .allow_invalid_utf8(!self.only_utf8)
                 .nest_limit(self.options.nest_limit)
                 .build();
-            let expr =
-                parser.parse(pat).map_err(|e| Error::Syntax(e.to_string()))?;
+            let parse_res = parser.parse(pat);
+            if parse_res.is_err() && self.options.throwaway_on_parse_error {
+                // The user opted to drop regex parse errors from the set
+                // instead of erroring out.
+                continue;
+            }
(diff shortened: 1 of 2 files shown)


keur:
cc @BurntSushi is this something that would be considered, or should i close it out?

BurntSushi:
So basically, I need to sit down and very carefully think about this API. I need time to do that and I'm not sure when that will happen. The current docs aren't quite sufficient for example. If a pattern isn't compiled, then does that mean there is a pattern ID that is guaranteed to never return? (And thus, pattern IDs become sparse.) Or do the pattern IDs for invalid patterns disappear and thus everything after gets shifted back one? I think your implementation is the latter, but it seems to me like the former is more useful behavior?

There's also the problem that if an error does occur, it is dropped on the floor and it's impossible to get it. Returning errors will likely make the API quite a bit more complex since you end up having to implement "partial success."

And that's just off the top of my head without thinking too hard about it.

With that said, I do agree that an API like this is potentially worth adding.

BurntSushi:
I am going to close this out for now, because I'm not sure I have the bandwidth any time soon to consider it.

I think it would be better to open an issue here and start exploring the API design space.

PR #781, opened 2021-05-17
Use proper `#[non_exhaustive]` attribute for error enums

marmeladema:


BurntSushi:
I have a suspicion that this is technically a breaking change. I don't mind putting out breaking change releases of `regex-syntax`, but I'd rather not do that now. When I'm ready to do that, I'll see if I can swing back around and cherry pick this PR. But for now, I'm going to close it.

Issue #756, opened 2021-04-08
test marker traits more rigorously on public API types


BurntSushi:
Since the `regex` crate uses interior mutability and a bit of synchronization, it follows that marker traits like `Send`, `Sync`, `UnwindSafe` and `RefUnwindSafe` are quite relevant and can change based on internal implementation details. #749 for example, removed the `thread_local` dependency and got a bit more rigorous about these marker traits on the main `Regex`/`RegexSet` types, but stopped there.

It turns out other types are impacted too. @kangalioo noted that, for example, `Send` [was _not_ implemented on the `CaptureMatches` iterator in `regex 1.3.9`](https://docs.rs/regex/1.3.9/regex/struct.CaptureMatches.html#impl-Send), but it [_is_ now implemented in `regex 1.4.5`](https://docs.rs/regex/1.4.5/regex/struct.CaptureMatches.html#impl-Send). I'm actually not sure why it wasn't implemented before. The `regex` crate doesn't use any non-Send stuff, so I'm not sure where that was coming from. I haven't investigated yet.

Either way, this is a compatibility hazard. Going from `!Send` to `Send` is fine, but it seems just as likely that we could regress too. The only way I can think to solve this is to assert expected marker traits for all public API types. It seems kind of absurd and tedious, but I don't have any better ideas.

alexfertel:
Can I take a jab at implementing these tests? If it's ok, can you give me some pointers on where I can find similar tests?

BurntSushi:
I think `tests/test_default.rs` is the place. You'll want to look for `oibits`, which is legacy terminology for marker traits that meant "Opt-In Built-In Traits."

Ideally there would be some way to write a test (maybe using shell script or Python, not sure) that would fail in the case of a public API type that _doesn't_ have an oibit test. That way, if a new public API type is added, we're forced to add an oibit test for it. But maybe this is too much machinery.

alexfertel:
Thank you for the quick response!

So, if I understood correctly, we want to avoid having to repeat groups of assertions like:

```rust
    assert_send::<Regex>();
    assert_sync::<Regex>();
    assert_unwind_safe::<Regex>();
    assert_ref_unwind_safe::<Regex>();
    assert_send::<RegexBuilder>();
    assert_sync::<RegexBuilder>();
    assert_unwind_safe::<RegexBuilder>();
    assert_ref_unwind_safe::<RegexBuilder>();
```

[...]

PR #793, opened 2021-07-04
Add a script to manage marker traits

alexfertel:
Closes https://github.com/rust-lang/regex/issues/756.

This PR introduces a Python script named `public-types-tests.py` that manages marker traits for the public API types. Following the discussion in https://github.com/rust-lang/regex/issues/756, this script is responsible for two tasks:

* Generate a test that asserts marker traits being implemented on the public API types (This test might need to be tweaked so it passes). This is done by generating the docs for this crate, parsing them, and
writing the assertions to the file `tests/marker_traits.rs`. This shouldn't have to be done very often, probably just once. We pass the `-g` flag as an argument to the script to generate the tests.
* Check the existence of marker-trait tests for each public API type, failing if it doesn't find *at least* one of the possible four.

This accomplishes three objectives:

* Automate generating the assertion for each type.
* Failing CI when a new type is added because no test related to it exists.
* Avoid regressions in current types having marker traits auto-implemented.

Two things remain for this PR to be ready:

- [x] Add comments to the script.
- [x] Setup CI.


alexfertel:
@BurntSushi Let me know if this is somewhat near what you had in mind.

alexfertel:
@BurntSushi Sorry to bother, is something blocking this PR, or is it way too different from what you expected? 🥺

BurntSushi:
Haven't had a chance to look yet, sorry. I have a baby at home and the past few weeks have been hectic.

alexfertel:
@BurntSushi  Oh, no worries at all. Congrats!!! Sorry for bothering 😅

BurntSushi (changes requested):
OK, so I've finally had a chance to look at this. I also re-read #756, and I feel like there was a miscommunication perhaps. In particular, I specifically do not want extra automation for generating the tests. It's effectively extra code for _almost_ no value. The public API just does not change much. So adding a new test by hand is done once in a blue moon and it doesn't cost much to do.

The _only_ thing I really want here is something to tell me whether a new test needs to be added or not, and then run that in CI. I see that here, which is great. But I really think it should just be a shell script. I have _almost_ gotten rid of Python from this repo, so I'd rather not bring more of it in.

I realize this is a really tardy review, so feel free to just drop this. The baby has kept me very busy. But if you're looking to keep going here, I'd say:

1. Drop all the codegen.
2. Convert the existence check to a simple shell script.

BurntSushi:
I'm going to close this out due to inactivity, but I would still like something _simple_ that checks that we're covering the API with tests for marker traits.

PR #811, opened 2021-10-31
feat: Regex::new takes Into<String>

DeveloperC286:
Makes Regex::new more flexible and allowing the type Into<String> to be used to create a regex.

### DIFF
--- a/src/re_builder.rs
+++ b/src/re_builder.rs
@@ -55,9 +55,9 @@ macro_rules! define_builder {
                 ///
                 /// If the pattern is invalid, then an error will be returned when
                 /// `build` is called.
-                pub fn new(pattern: &str) -> RegexBuilder {
+                pub fn new<T: Into<String>>(pattern: T) -> RegexBuilder {
                     let mut builder = RegexBuilder(RegexOptions::default());
-                    builder.0.pats.push(pattern.to_owned());
+                    builder.0.pats.push(pattern.into());
                     builder
                 }
 
--- a/src/re_bytes.rs
+++ b/src/re_bytes.rs
@@ -116,7 +116,31 @@ impl Regex {
     /// to search, split or replace text in a string.
     ///
     /// If an invalid expression is given, then an error is returned.
-    pub fn new(re: &str) -> Result<Regex, Error> {
+    ///
+    /// # Example
+    ///
+    /// Create a new Regex using a &str.
+    ///
+    /// ```rust
+    /// # use regex::bytes::Regex;
+    /// # fn main() {
+    /// let regex: &str = r"\b\w{13}\b";
+    /// Regex::new(regex);
+    /// # }
+    /// ```
+    ///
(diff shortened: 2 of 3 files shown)


BurntSushi:
I'm sorry, but I'm opposed to this for a few reasons:

1. Generics add complexity. I use them when necessary or compelling, but I don't think either is the case here.
2. I suspect this will break people. I'm on mobile, so I can't work through an example, but the main thing I would be concerned about is whether the generics prevent deref to `&str` from occurring in certain instances.

BurntSushi:
I had a chance to play around with things, and this does indeed look like a bad enough breaking change to be a non-starter. For example, this compiles:

```rust
fn test1(_s: &str) {}

fn main() {
    let x: Vec<&str> = vec!["foo"];
    x.iter().map(|s| test1(s)).collect::<Vec<()>>();
}
```

but this doesn't:

```rust
fn test1(_s: impl Into<String>) {}

fn main() {
    let x: Vec<&str> = vec!["foo"];
    x.iter().map(|s| test1(s)).collect::<Vec<()>>();
}
```

and AIUI, this is exactly because of deref working in the former example due to the concrete type, where as in the latter example, the lack of concrete type makes the deref inference fail. I don't consider this a weird corner case, and although this is technically under the umbrella of "allowable breakage," I think this would be too severe in practice.

So if this change were made, it would have to wait until a `regex 2.0` release (of which there are no plans). And even then, my reason (1) in my previous comment would still apply.

DeveloperC286:
Hi, @BurntSushi yes you are right this was a breaking change. I tested my patch using my tool called `did_i_break_it` to compile all crates on crates.io that depend on regex and there were a number that failed to compile.

Regarding the unnecessary complexity that generics add, I was actually adding the generic to make it more ergonomic.In my usecase I had a `Vec<String>` I wanted to create regexes from, but as Regex::new only accept `&str` I was having to add a `&` into my code to get a reference. 

I then read regex's source code and seen you actually convert it back to a String, so thought you may aswell use `Into<String>` as it may offer a minimal performance improvement for certain cases when people give you ownership of the String, by removing another allocation(did not test, just a guess).

So if `Into<String>` is a breaking change what about `new()` taking a `AsRef<str>`? So new can take either a `&str` or `String`?

BurntSushi:
I think `AsRef` has the same problems.

> as it may offer a minimal performance improvement for certain cases

Very unlikely. The copying of a single string is a pittance compared to the work of regex compilation.

> Regarding the unnecessary complexity that generics add, I was actually adding the generic to make it more ergonomic.In my usecase I had a `Vec<String>` I wanted to create regexes from, but as Regex::new only accept `&str` I was having to add a `&` into my code to get a reference.

Yes, I understand how generics can occasionally make things more ergonomic. But it's a balancing act, as the generics will also cause less ergonomics in some cases. _And_ generics tend to increase complexity by making the API harder to read.

PR #818, opened 2021-11-19
Fix s flag documentation

nextuponstream:
`(?s:.)` mislead me while I was working with the regex crate to capture a paragraph. Replacing flag declaration with `(?s)` fixed my problem.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -221,7 +221,7 @@ Most features of the regular expressions in this crate are Unicode aware. Here
 are some examples:
 
 * `.` will match any valid UTF-8 encoded Unicode scalar value except for `\n`.
-  (To also match `\n`, enable the `s` flag, e.g., `(?s:.)`.)
+  (To also match `\n`, enable the `s` flag, e.g., `(?s)`.)
 * `\w`, `\d` and `\s` are Unicode aware. For example, `\s` will match all forms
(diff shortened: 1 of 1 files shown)


BurntSushi:
Could you say what confused you here? I don't understand why the replacement is clearer. And the replacement you've used here doesn't include `.`.

nextuponstream:
Sorry for the late answer

I got confused when I blindly copy pasted from the documentation that my code did not work as expected. 

I wanted to do multi line capture to extract a paragraph in a text. Since `.` does not capture `\n`, I followed the documentation which said `To also match \n, enable the s flag, e.g., (?s:.)`. This did not work on first try. To illustrate this, I created this example:

```rust
use regex::Regex;

fn main() {
    let s = "not captured\n\nhello\nworld\n\nnot captured";
    let re: Regex = Regex::new(r"(?s)not captured\n\n(.*)\n\nnot captured").unwrap();
    let hello_world_to_capture = re.captures(s).unwrap();

    assert_eq!(&hello_world_to_capture[1], "hello\nworld");
}
```

This code will capture "hello world". However, when I change the flag from `(?s)` to `(?s:.)`, the code panics at unwrap.

When I encountered this, I thought the documentation was out of date which is why I submitted this PR. If it is not, please feel free to close this PR.

PR #821, opened 2021-11-28
Simplify fuzzer, but also fuzz all the (non-limits) options in RegexBuilder

5225225:
I'm not sure if the OSS-fuzz integration can handle `data` not being a `&[u8]`, but `cargo-fuzz` can. (It looks like they just use `cargo fuzz`, so it should work.)

This does change the input format, which might confuse the fuzzer (but I'd *expect* it to pretty quickly how to pass strings). I could make this as a new fuzz target if you want.

One advantage taking &str over &[u8] has is that if the end of a &str is invalid UTF-8, then it will still be used, but it (the implementation of `Arbitrary` for &str) just cuts off the invalid part, which is better than not running at all.

I ran this for a bit locally and didn't find any issues. I can back out the commit to add the options and just have the one that removes the complex parsing code if you want.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2018"
 cargo-fuzz = true
 
 [dependencies]
-libfuzzer-sys = "0.4.1"
+libfuzzer-sys = { version = "0.4.1", features = ["arbitrary-derive"] }
 
 [dependencies.regex]
 path = ".."
--- a/fuzz/fuzz_targets/fuzz_regex_match.rs
+++ b/fuzz/fuzz_targets/fuzz_regex_match.rs
@@ -1,22 +1,71 @@
 #![no_main]
 use libfuzzer_sys::fuzz_target;
 
-fuzz_target!(|data: &[u8]| {
-    if data.len() < 2 {
-        return;
+use libfuzzer_sys::arbitrary;
+
+#[derive(arbitrary::Arbitrary)]
+struct FuzzCase<'a> {
+    pattern: &'a str,
+    input: &'a str,
+    case_insensitive: bool,
+    multi_line: bool,
+    dot_matches_new_line: bool,
+    swap_greed: bool,
+    ignore_whitespace: bool,
+    unicode: bool,
+    octal: bool,
+}
+
+impl std::fmt::Debug for FuzzCase<'_> {
+    fn fmt(
+        &self,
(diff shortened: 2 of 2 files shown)


BurntSushi:
@5225225 Thanks for this PR! This is a really nice simplification.

One thing that this seems to make harder though is converting failing fuzz inputs into regression tests. Today, I can do something like this:

```
#[test] 
fn regression_foo_bar() {
    let raw = include_bytes!("path/to/crash-blahblah");
    let _ = run(raw);
}

fn run(raw: &[u8]) -> Option<()> {
    // a copy of the fuzz target code
    Some(())
}
```

But with this simplification, it seems like it covers up the conversion of the fuzzer input of `&[u8]`, which I think makes writing `run` harder. Is there a way to explicitly convert `&[u8]` into the more structure types such that I could copy that into the test code above?

(I'm also open to solving the problem of fuzz regression tests in a different way.)

5225225:
I did something that I'm not sure is terrible or very clever. But it's definitely *easy*. (Below is for an artificial error, i just put in an `assert` somewhere to panic on nontrivial input).

```
Failing input:

	fuzz/artifacts/fuzz_regex_match/crash-dfcf19f0e4fba283e45253fd6eabd3c77b557cdf

Output of `std::fmt::Debug`:

	
	let r = regex::RegexBuilder::new(",@*{000120}}}}}}}")
	    .case_insensitive(false)
	    .multi_line(false)
	    .dot_matches_new_line(false)
	    .swap_greed(false)
	    .ignore_whitespace(false)
	    .unicode(false)
	    .octal(false)
	    .build();
	
	if let Ok(re) = r {
	    re.is_match("");
	}
	        

Reproduce with:

	cargo fuzz run fuzz_regex_match fuzz/artifacts/fuzz_regex_match/crash-dfcf19f0e4fba283e45253fd6eabd3c77b557cdf

Minimize test case with:

	cargo fuzz tmin fuzz_regex_match fuzz/artifacts/fuzz_regex_match/crash-dfcf19f0e4fba283e45253fd6eabd3c77b557cdf
```

Basically, the debug print of whatever you're fuzzing is printed on failure, so if you just make that debug print actually just valid rust code, you can copy paste directly from the output to a `#[test]`.

This relies on debug print on `&str` round tripping, but if it doesn't then you would just see either compilation errors, or the test wouldn't fail when it should.

PR #825, opened 2021-12-29
refactor: use shorthand syntax

pickfire:


BurntSushi:
So basically, I think this is the same kind of thing as #864.

In principle I'd accept a PR that fixed _all_ of them. But to be honest, I'm not sure it's worth it. A lot of this code is going to get dropped Real Soon Now.

PR #826, opened 2021-12-30
Add makefile to install C library; add tidbit to README

vmchale:
I am writing bindings to this library and so it would be useful to have installation instructions I can point users to!

Cheers/Thanks!

### DIFF
--- a/regex-capi/Makefile
+++ b/regex-capi/Makefile
@@ -1,0 +1,9 @@
+RS_SRC := $(shell find src -type f) Cargo.toml
+
+../target/release/librure.a ../target/release/librure.dylib &: $(RS_SRC)
+	cargo build --release
+
+install: ../target/release/librure.a ../target/release/librure.dylib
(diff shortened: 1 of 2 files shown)


BurntSushi:
So I think I'm going to pass on this unfortunately. The principle reason is that I'm not sure I want to get in the business of maintaining instructions for how to install a C library. In particular, this hard-codes an installation path and specifically recommends running `sudo make install`, which IMO is just bad juju. I personally would _never_ run that installation step by hand, so I don't think I can reasonably ask others to. Additionally, there is no uninstall option.

I think what I would rather here is to explain what Cargo does and where it drops the C artifacts. Then people can pick those up and do what they need with them.

On top of that, there is really nothing making sure that the `Makefile` here is actually correct. `RS_SRC := $(shell find src -type f) Cargo.toml` is incorrect for example, and it would be non-trivial to both get that right _and_ make sure it stays correct.

Bottom line is that this repo already has a build tool and its name is `cargo`.

PR #830, opened 2022-01-21
Fix some dead_code warnings

jyn514:
```
warning: field is never read: `printer`
  --> regex-syntax/src/ast/print.rs:66:5
   |
66 |     printer: &'p mut Printer,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field is never read: `printer`
  --> regex-syntax/src/hir/print.rs:74:5
   |
74 |     printer: &'p mut Printer,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^
```


BurntSushi:
Thanks! It looks like I already fixed these before getting to this PR.

PR #832, opened 2022-01-28
docs: clarify Captures::len includes non-matching captures

CAD97:
This caught me out: mentally I map "captured groups" to the _successful_ captures; capture groups that didn't capture anything weren't "captured." I understand why the function returns what it does in retrospect, but hopefully this can help someone else.

### DIFF
--- a/src/re_unicode.rs
+++ b/src/re_unicode.rs
@@ -989,7 +989,7 @@ impl<'t> Captures<'t> {
         expand_str(self, replacement, dst)
     }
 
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):

BurntSushi (changes requested):
Can you make the same change to `re_bytes.rs` as well?

BurntSushi:
Nevermind, I'll make the change and include it in a rollup.

PR #869, opened 2022-06-14
ci: attempt to fix cross in CI

BurntSushi:
This is a perfect example of why adding support for more targets is
costly. Every so often, the cross compiling toolchain breaks for
mysterious reasons.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,9 +75,11 @@ jobs:
     - name: Install and configure Cross
       if: matrix.target != ''
       run: |
-        # FIXME: to work around bugs in latest cross release, install master.
-        # See: https://github.com/rust-embedded/cross/issues/357
-        cargo install --git https://github.com/rust-embedded/cross
+        # We used to install 'cross' from master, but it kept failing. So now
+        # we build from a known-good version until 'cross' becomes more stable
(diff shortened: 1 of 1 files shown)


BurntSushi:
For posterity, I cannot run `cargo install --bins --git https://github.com/cross-rs/cross` to install cross. This seems to be a recent regression, but it also looks like a Cargo problem. I get the same error in CI as I do locally:

```
$ cargo install --bins --git https://github.com/cross-rs/cross
    Updating git repository `https://github.com/cross-rs/cross`
error: multiple packages with binaries found: cross, xtask. When installing a git repository, cargo will always search the entire repo for any Cargo.toml. Please specify which to install.
```

Trying `--bin cross` produces the same error.

CI log: https://github.com/rust-lang/regex/runs/6880200164?check_suite_focus=true#step:4:14

This is a _different_ error than what I was seeing yesterday. The above error is a result of new commits pushed to Cross since yesterday.  Yesterday, I was seeing this error message:

```
Warning: Falling back to `cargo` on the host.
```

when running `cross build --verbose i686-unknown-linux-gnu`.

CI log: https://github.com/rust-lang/regex/runs/6865916772?check_suite_focus=true#step:7:31

Checking the Cross issue tracker doesn't seem to show any issues.

PR #872, opened 2022-06-18
Allow use of `regex_syntax::is_meta_character` func from `regex` (and fix a minor docs typo)

alexander-beedie:
Hi, thanks for all the great regular expression work! (And a small request ;)

Currently `regex` offers the `escape` func from `regex-syntax`, which in turn calls `is_meta_character`; I find myself needing _that_ function, but not really wanting to have to maintain a reference to both crates due to the possibility that (one day!) the list of chars may be updated, and a version-mismatch might cause some minor issue before being spotted.

Given the presence of `escape`, are there any objections to also offering `is_meta_character`, as per this pull req? Would mean one less import (nice, given the size of the unicode tables), and guarantee that -no matter what changes in the future- the function will always return correctly as it is sure to be aligned with the underlying engine version.

Also: spotted/fixed a one-letter typo in the `is_meta_character` docs :)

### DIFF
--- a/regex-syntax/src/lib.rs
+++ b/regex-syntax/src/lib.rs
@@ -195,7 +195,7 @@ pub fn escape_into(text: &str, buf: &mut String) {
     }
 }
 
-/// Returns true if the give character has significance in a regex.
+/// Returns true if the given character has significance in a regex.
 ///
 /// These are the only characters that are allowed to be escaped, with one
 /// exception: an ASCII space character may be escaped when extended mode (with
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -629,7 +629,7 @@ pub use crate::re_builder::unicode::*;
 pub use crate::re_set::unicode::*;
 #[cfg(feature = "std")]
 pub use crate::re_unicode::{
-    escape, CaptureLocations, CaptureMatches, CaptureNames, Captures,
+    escape, is_meta_character, CaptureLocations, CaptureMatches, CaptureNames, Captures,
     Locations, Match, Matches, NoExpand, Regex, Replacer, ReplacerRef, Split,
     SplitN, SubCaptureMatches,
 };
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -629,9 +629,9 @@ pub use crate::re_builder::unicode::*;
 pub use crate::re_set::unicode::*;
 #[cfg(feature = "std")]
 pub use crate::re_unicode::{
-    escape, is_meta_character, CaptureLocations, CaptureMatches, CaptureNames, Captures,
-    Locations, Match, Matches, NoExpand, Regex, Replacer, ReplacerRef, Split,
-    SplitN, SubCaptureMatches,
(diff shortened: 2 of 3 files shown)


BurntSushi:
I *believe* all of your concerns are misplaced. There is no cost to adding regex-syntax to your crate if you are already using regex, since regex depends on regex-syntax. So there are no additional Unicode tables being brought in. Moreover, your compatibility concerns are moot too, since the set of meta characters can't be changed unless regex itself has a semver bump, even if is_meta_character lives in regex-syntax.

So with your concerns mitigated, I'm not sure there is a reason to elevate this function to regex proper. It seems to me like it belongs in regex-syntax. The escape function is something with general and widespread utility. The is_meta_character function is much more niche.

alexander-beedie:
Ahhh, got it - I think the wording of the `regex-syntax` crate docs had me slightly concerned that it could move out-of-step with `regex` in such a way that a mismatch in the list of metachars _might_ be a theoretical concern - if that's not the case and it is considered a bit too niche for wider adoption, happy to close! (I'm administering a minor facepalm about the import/tables, because _obviously_ they are already imported by `regex`, hah).

BurntSushi:
Aye yeah. There probably should be a few paragraphs on the compatibility story of regex-syntax. And probably the doc for is_meta_character should be improved too.

The short story is that regex-syntax can do whatever it wants, including breaking its API at a rapid cadence, but it is still ultimately responsible for maintaining compatibility at the level of the concrete syntax. Which is where the definition of "meta character" lives. Anything else is fair game. (And indeed, the crate has changed quite a bit over the years.)

alexander-beedie:
Not to worry, and thanks for the clarification - I shall breathe easy... ;)
Moved the docs/typo fix into a separate req.

BurntSushi:
This is also why regex-syntax is very explicitly not a public dependency of regex. That's why you can't go directly from an Hir to a Regex. A public dependency would mean regex-syntax isn't free to evolve its API. So the only compatibility it has to maintain is the concrete syntax. And that can't really be represented by semver on regex-syntax. It can only be represented by semver on regex proper.

BurntSushi:
Sorry for the stream of consciousness comments here. I'm on toddler duty.

I think there is one downside to using regex-syntax directly. When I do a 0.7.0 release (for example), regex will do a seamless upgrade. Until your code is updated, you'll have two different versions of regex-syntax in your dependency tree. It's harmless, but will increase compilation times.

alexander-beedie:
>  I'm on toddler duty.

A worthy task  ;)

> I think there is one downside to using regex-syntax directly. When I do a 0.7.0 release (for example), regex will do a seamless upgrade. Until your code is updated, you'll have two different versions of regex-syntax in your dependency tree. It's harmless, but will increase compilation times.

Yup, it's that sort of thing I was hoping to streamline a bit by having it available from `regex`; as `is_meta_character` is the sole function required, and it's completely standalone, the compiler _might_ just pull it out without compiling the rest, hmm... One can hope, heh.

PR #884, opened 2022-07-15
Use `#[non_exhaustive]` attribute instead of `__Nonexhaustive` enum variant hack

rhysd:
[`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) was added at [Rust 1.40](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1400-2019-12-19). Since MSRV of this project is 1.41.1, the attribute is available. This PR replaces `__Nonexhaustive` enum variant hack with the attribute.

The benefits to use the attribute are:

- The attribute is dedicated for this use case. So compiler's error message is better than an error around `__Nonexhausitve` variant.
- `#[non_exhaustive]` allows the exhaustive pattern check at `match` in the crate where the enum/struct is implemented. This can catch bugs like 31d78a6.

### DIFF
--- a/regex-syntax/src/ast/mod.rs
+++ b/regex-syntax/src/ast/mod.rs
@@ -202,13 +202,15 @@ impl error::Error for Error {
             GroupUnclosed => "unclosed group",
             GroupUnopened => "unopened group",
             NestLimitExceeded(_) => "nest limit exceeded",
+            RepetitionCountDecimalEmpty => {
+                "invalid decimal in repetition quantifier"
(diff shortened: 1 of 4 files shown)


rhysd:
I considered this change may introduce some breaking change just before creating this PR, but I could come up with nothing too. Possibly it may change a data layout of enum or how a compiler optimizes it. But they are not an API change and usually don't break user code.

PR #889, opened 2022-07-18
Update version number in Readme

atouchet:


BurntSushi:
I went a different direction to avoid having to worry about this in the future. :-) Thanks.

PR #892, opened 2022-07-20
update doc examples to use `once_cell` instead of `lazy_static`

vidhanio:
this change is useful because of the proposed addition of the `once_cell` api into std. users using the `once_cell` pattern will find it easier to swap to the `std` versions when/if they are merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "regex"
-version = "1.6.0"  #:version
+version = "1.6.0" #:version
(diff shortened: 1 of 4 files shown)


vidhanio:
also, if wanted i could update the benches to use `once_cell` instead of `lazy_static` as well.

BurntSushi:
I think I'd prefer to avoid this. I'd rather just wait until the std APIs are available to avoid churn. There's nothing egregiously wrong with `lazy_static`.

PR #897, opened 2022-07-24
docs(lib): improve section for empty matches

mrl5:
this improvement is a consequence of the Q&A discussion from
https://github.com/rust-lang/regex/discussions/896


BurntSushi (changes requested):
Thanks! Could you make sure lines are wrapped in a style consistent with the surrounding code? In this case, that's 79 columns inclusive.

BurntSushi (approved):
Thanks!

Issue #719, opened 2020-10-13
Trying to reference the Cf category generates an error


justinmchase:
If I use the unicode class `cf`:

```rs
static ref RE: Regex = Regex::new(r"^[\p{cf}]$").unwrap();
```

It generates this error:
```rs
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    ^[\p{cf}]$
      ^^^^^^
error: Unicode property not found
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)', src/runtime/modules/unicode.rs:21:28
```

But I see the format is here in the code:
https://github.com/rust-lang/regex/blob/master/regex-syntax/src/unicode_tables/property_values.rs#L72

If I change the regex to be `\p{Format}` then it actually works. This is a little confusing to me, can someone please explain why `\p{cf}`

PR #899, opened 2022-08-01
Syntax: Make \p{Sc} work

### LINKED ISSUE #835: Currency symbol abbreviation "Sc" is not a valid property

1.5.4

The unicode abbreviation for for the currency symbols as defined here https://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt
is not valid and gives a regex parse error.

Add the regex crate to the cargo.toml file

`Cargo.toml`
```toml
[dependencies]
regex = "1.5.4"
```

`main.rs`
```rust
use regex::Regex;

fn main() {
    let regex = Regex::new(r"\p{Sc}").unwrap();
    if regex.is_match("$") {
        println!("Hello, world!");
    }
}
```

```sh
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    \p{Sc}
    ^^^^^^
error: Unicode property not found
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)', src/main.rs:4:39
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected the regex to compile and return the "$" as a match.

snsmac:
I am not completely sure if this change introduces some unwanted side effects, as there seems to be support for the 'Script' property, which is not the case for the 'Case_Folding' property

Fixes #835
Related #719 b1489c8

### DIFF
--- a/regex-syntax/src/unicode.rs
+++ b/regex-syntax/src/unicode.rs
@@ -243,7 +243,9 @@ impl<'a> ClassQuery<'a> {
         // a general category. (Currently, we don't even support the
         // 'Case_Folding' property. But if we do in the future, users will be
         // required to spell it out.)
-        if norm != "cf" {
+        // 'sc' refers to the 'Currency_Symbol' general category, but is also
+        // the abbreviation for the 'Script' property.
+        if norm != "cf" && norm != "sc" {
             if let Some(canon) = canonical_prop(&norm)? {
                 return Ok(CanonicalClassQuery::Binary(canon));
             }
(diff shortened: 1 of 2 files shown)


BurntSushi:
This was closed by 9e6b09415a32738798963957afbd7f0625d93580.

PR #906, opened 2022-09-21
impl Default for RegexSet

### LINKED ISSUE #905: impl Default for RegexSet?
#### Describe your feature request

Would it be reasonable to `impl Default` on `RegexSet`, returning an empty set as if it was constructed from an empty list?

This would be helpful when using it inside a struct for which I want to `derive Default`.

sourcefrog:
Fixes #905

### DIFF
--- a/src/re_set.rs
+++ b/src/re_set.rs
@@ -289,6 +289,12 @@ impl RegexSet {
     }
 }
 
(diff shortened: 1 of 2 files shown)


sourcefrog:
Thank you!

Issue #824, opened 2021-12-22
Add convenience methods for extracting bits of text


orlp:
A lot of my use cases for regex involve extracting some information from text with a simple pattern, where full parsing would be overkill. In this scenario I generally just have `N` subsections of the text I'm looking to extract (possibly many times), where `N` is a small constant. Some examples from the recent advent of code:

```rust
let re = Regex::new(r"target area:\s*x=(-?\d+)..(-?\d+), y=(-?\d+)..(-?\d+)")?;
let re = Regex::new(r"Player 1 starting position:\s*(\d)\s*Player 2 starting position:\s*(\d)\s*")?;
let re = Regex::new(r"(on|off)\s+x=(-?\d+)..(-?\d+),y=(-?\d+)..(-?\d+),z=(-?\d+)..(-?\d+)")?;
```

In such a scenario it is quite the pain to extract the information I want using the current API. It generally involves something like this, for the last example:

```rust
use anyhow::Context;
use itertools::Itertools;
let (status, x1, x2, y1, y2, z1, z2) = re
    .captures(&line)
    .context("match failed")?
    .iter()
    .skip(1)
    .map(|c| c.unwrap().as_str())
    .collect_tuple()
    .unwrap();
```

Note that I already had to include a third-party library `itertools` just to get `collect_tuple`, without that we would either have to start repeatedly calling `next()` or `get(i)`, or invoke an unnecessary allocation, and in both cases an extra check is needed to ensure there weren't too many captures (indicating a programmer error).

I think that adding a convenience method for this incredibly common use-case would be nice. I would propose the following (with sample implementation):

```rust
/// Finds the leftmost-first match in `text` and returns a tuple containing the whole match
/// and its N participating capture groups as strings. If no match is found, `None` is returned.
///
/// # Panics
///
/// Panics if the number of participating captures is not equal to N.
fn extract<'t, const N: usize>(&self, text: &'t str) -> Option<(&'t str, [&'t str; N])> {
    let caps = re.captures(text)?;
    let mut participating = caps.iter().flatten();
    let whole_match = participating.next().unwrap().as_str();
    let captured = [0; N].map(|_| participating.next().unwrap().as_str());
    assert!(participating.next().is_none(), "too many participating capture groups");
    Some((whole_match, captured))
}
```

And similarly `extract_iter` returning `impl Iterator<Item = (&'t str, [&'t str; N])>`.

---

Then, the cumbersome example above would become the elegant:

```rust
[...]

PR #908, opened 2022-09-24
Added `participating_captures_len`

orlp:
Long overdue, but this is the first step towards adding https://github.com/rust-lang/regex/issues/824, as greenlit by @BurntSushi.

This new method of `Regex` returns the number of capture groups that will be filled during a successful match, or `None` if this can't be statically determined during `Regex` compile time.

### DIFF
--- a/regex-syntax/src/hir/mod.rs
+++ b/regex-syntax/src/hir/mod.rs
@@ -243,6 +243,7 @@ impl Hir {
         info.set_match_empty(true);
         info.set_literal(false);
         info.set_alternation_literal(false);
+        info.static_capture_count = Some(0);
         Hir { kind: HirKind::Empty, info }
     }
 
@@ -268,6 +269,7 @@ impl Hir {
         info.set_match_empty(false);
         info.set_literal(true);
         info.set_alternation_literal(true);
+        info.static_capture_count = Some(0);
         Hir { kind: HirKind::Literal(lit), info }
     }
 
@@ -285,6 +287,7 @@ impl Hir {
         info.set_match_empty(false);
         info.set_literal(false);
         info.set_alternation_literal(false);
+        info.static_capture_count = Some(0);
         Hir { kind: HirKind::Class(class), info }
     }
 
@@ -318,6 +321,7 @@ impl Hir {
         if let Anchor::EndLine = anchor {
             info.set_line_anchored_end(true);
         }
+        info.static_capture_count = Some(0);
         Hir { kind: HirKind::Anchor(anchor), info }
     }
 
@@ -345,6 +349,7 @@ impl Hir {
         if let WordBoundary::AsciiNegate = word_boundary {
             info.set_always_utf8(false);
         }
+        info.static_capture_count = Some(0);
         Hir { kind: HirKind::WordBoundary(word_boundary), info }
     }
 
@@ -372,6 +377,28 @@ impl Hir {
         info.set_match_empty(rep.is_match_empty() || rep.hir.is_match_empty());
         info.set_literal(false);
         info.set_alternation_literal(false);
+        info.static_capture_count = {
+            use RepetitionRange::{AtLeast, Bounded, Exactly};
+            match (&rep.kind, rep.hir.info.static_capture_count) {
+                // Always zero.
+                (_, Some(0)) => Some(0),
+                (RepetitionKind::Range(Exactly(0)), _) => Some(0),
+                (RepetitionKind::Range(Bounded(_, 0)), _) => Some(0),
+
+                // Impossible to know statically.
+                (_, None) => None,
+                (RepetitionKind::ZeroOrOne, _) => None,
+                (RepetitionKind::ZeroOrMore, _) => None,
+                (RepetitionKind::Range(AtLeast(0)), _) => None,
+                (RepetitionKind::Range(Bounded(0, _)), _) => None,
+
+                // Guaranteed to be static.
+                (RepetitionKind::OneOrMore, Some(n)) => Some(n),
+                (RepetitionKind::Range(Exactly(_)), Some(n)) => Some(n),
+                (RepetitionKind::Range(AtLeast(_)), Some(n)) => Some(n),
+                (RepetitionKind::Range(Bounded(_, _)), Some(n)) => Some(n),
+            }
+        };
         Hir { kind: HirKind::Repetition(rep), info }
     }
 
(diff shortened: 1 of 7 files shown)


BurntSushi (changes requested):
Thanks!

One particularly unfortunate thing about timing here is that I have a [WIP branch](https://github.com/rust-lang/regex/compare/master...ag/regex-automata) that refactors quite a bit of code that this PR also touches. In particular, `HirInfo` is dropped in favor of a new `Properties` type, and the `Repetition` HIR definition is simplified considerably. So pretty much all of the code in this PR will need to be ported over to that branch. The other unfortunate thing is that I have no timeline as to when that branch will get pulled into `master`. Maybe by the end of the year?

We could get this merged now, but the merge conflicts are going to be brutal enough that it would be easier to just port the patch by hand.

I can probably do the porting for you, but commit authorship won't be preserved.

Otherwise, there are a few changes here that I think need to be made. I'm also not entirely happy with the name `participating_captures_len`. I'll think about that.

orlp:
> We could get this merged now, but the merge conflicts are going to be brutal enough that it would be easier to just port the patch by hand.

We should probably coordinate a bit then before I make any further changes. I could re-do the patch on the WIP branch myself if it is in a compiling state and the structure is roughly the same. I don't really see the value of putting it into `master` if it's going to be replaced soon anyway. Especially since you said this:

> Could you back out the public API changes to regex please? I'm fine adding this to regex-syntax, but I'm less sure about it being in the regex API.

If `participating_captures_len` is not accessible externally from the `Regex` object itself, there is no point to this PR other than as preparatory work for the inclusion of `extract` in `regex` itself. If I convinced you about this, then that's great and I'm happy to help where possible. But if you are still on the fence about `extract` then adding `participating_captures_len` to `regex-syntax` does nothing for making `extract` possible as an extension trait.

BurntSushi:
> I could re-do the patch on the WIP branch myself if it is in a compiling state and the structure is roughly the same.

It is and the overall structure is indeed the same.

> I don't really see the value of putting it into master though if it's going to be replaced soon anyway.

Aye. Just to set expectations here, this branch is the culmination of #656, which I have been working on for years. While it might get merged by the end of this year, I can't make any promises.

> If `participating_captures_len` is not accessible externally from the `Regex` object itself, there is no point to this PR other than as preparatory work for the inclusion of `extract` in `regex` itself. If I convinced you about this, then that's great and I'm happy to help where possible. But if you are still on the fence about `extract` then adding `participating_captures_len` to `regex-syntax` does nothing for making `extract` possible as an extension trait.

Ah hmmm, yes, I see. You could still do it in another crate, but it would require a lot more ceremony. Namely, a wrapper type around `Regex` that parsed the pattern a second time to figure out the `participating_captures_len` property. That's not so good.

OK, I'm fine with adding this to the `regex` API, but it should come with at least some doc examples. One example might even be the `extract` implementation itself, since it's so simple.

Thinking about alternative names too... Ideas:

* `fixed_captures_len`
* `static_captures_len`

Other ideas?

orlp:
For the bikeshedding, other options might be `unconditional_captures_len`, `exact_captures_len`, `guaranteed_captures_len`, `proven_captures_len`.

> Aye. Just to set expectations here, this branch is the culmination of https://github.com/rust-lang/regex/issues/656, which I have been working on for years. While it might get merged by the end of this year, I can't make any promises.
> ...
> OK, I'm fine with adding this to the regex API, but it should come with at least some doc examples. One example might even be the extract implementation itself, since it's so simple.

Alright then, since you're okay with it going public on `Regex` I think I will (for now) simply make your requested changes for the `master` branch, so that hopefully in the next release of `regex` it's possible to write the checked extension trait and have it be immediately useful.

> OK, I'm fine with adding this to the regex API, but it should come with at least some doc examples. One example might even be the extract implementation itself, since it's so simple.

I think while simple it might be a bit too big for an example. I'll make sure to add some examples though.

BurntSushi:
> Alright then, since you're okay with it going public on `Regex` I think I will (for now) simply make your requested changes for the `master` branch, so that hopefully in the next release of `regex` it's possible to write the checked extension trait and have it be immediately useful.

Sorry, to be clear, I do not really want to merge this into master. As I said above, the merge conflicts against my WIP branch will be brutal.

orlp:
> > Alright then, since you're okay with it going public on `Regex` I think I will (for now) simply make your requested changes for the `master` branch, so that hopefully in the next release of `regex` it's possible to write the checked extension trait and have it be immediately useful.
> 
> Sorry, to be clear, I do not really want to merge this into master. As I said above, the merge conflicts against my WIP branch will be brutal.

Ah alright. I'll work on a pull request on the other branch then.

BurntSushi:
OK, I've ported this patch to the new branch. And I've put it in the `regex` API as well. Here is that method and its docs:

```rust
    /// Returns the total number of explicit capturing groups that appear in
    /// every possible match.
    ///
    /// If the number of capture groups can vary depending on the match, then
    /// this returns `None`. That is, a value is only returned when the number
    /// of matching groups is invariant or "static."
    ///
    /// Note that like [`Regex::captures_len`], this **does** include the
    /// implicit capturing group corresponding to the entire match. Therefore,
    /// when a non-None value is returned, it is guaranteed to be at least `1`.
    /// Stated differently, a return value of `Some(0)` is impossible.
    ///
    /// # Example
    ///
    /// This shows a few cases where a static number of capture groups is
    /// available and a few cases where it is not.
    ///
    /// ```
    /// use regex::Regex;
    ///
    /// let len = |pattern| {
    ///     Regex::new(pattern).map(|re| re.static_captures_len())
    /// };
    ///
    /// assert_eq!(Some(1), len("a")?);
    /// assert_eq!(Some(2), len("(a)")?);
    /// assert_eq!(Some(2), len("(a)|(b)")?);
    /// assert_eq!(Some(3), len("(a)(b)|(c)(d)")?);
    /// assert_eq!(None, len("(a)|b")?);
    /// assert_eq!(None, len("a|(b)")?);
    /// assert_eq!(None, len("(b)*")?);
    /// assert_eq!(Some(2), len("(b)+")?);
    ///
    /// # Ok::<(), Box<dyn std::error::Error>>(())
    /// ```
    #[inline]
    pub fn static_captures_len(&self) -> Option<usize> { ... }
```

In particular, there is one crucial difference here (other than the name): for a `Some` value, the length is **always at least `1`.** This makes it consistent with the value returned by `captures_len`.

orlp:
Seems fair enough to me 👍

BurntSushi:
@orlp Also, thank you for the great tests! I was able to copy them over pretty easily, and it looks like they have good coverage. Overall it was pretty easy to port over. The logic was all roughly the same!

Issue #330, opened 2017-01-19
Capturing group in split doesn't contain the matches


yangsibai:
When I use capturing group in my pattern, I want the result of `split` can keep the matched strings. It seems the current implementation of `split` removed them.

For example: the patten is `"(x)"`, the text is "1x2", the result should be `["1", "x", "2"]`。

BurntSushi:
Could you say more about the *actual* problem you're trying to solve? Having the captured text inside the iterator is a means to end.

In any case, I don't think something like this should be added to the crate. I find it to be very strange behavior that would be better handled by something not called "split." You might consider just writing it yourself. For example, if you use `captures_iter`, then you'll get an iterator over all matches. With a little extra state, it would be easy to turn that into an iterator over an enum like this:

```rust
enum SplitState<'t> {
    Unmatched(&'t str),
    Captured(Captures<'t>),
}
```

If you'd like help writing the code, I can do that for you, but it might take me a bit to get to.

yangsibai:
>Could you say more about the actual problem you're trying to solve? Having the captured text inside the iterator is a means to end.

This could be very useful when I want to know what string split the text.

For example:

My texts could be `"1+2"`, `"1-2"`, and the pattern is "([^\d]+)".

I want the split method get results of `["1", "+", "2"]` and `["1", "-", "2"]`, then I can know what is the operator.

Since the pattern contains  group capturing, it should keep the matches into the results. I don't agree to it's a strange behavior.

>You might consider just writing it yourself.

Thank you for telling me the `captures_iter`. I will try it later.

BurntSushi:
It might not be strange, but I don't think it's standard behavior for something called "split," and I do at least think it's niche.

In any case, I wrote the iterator I mentioned in my previous comment. It's untested, but it does handle your case. :-)

```rust
extern crate regex;

use regex::{Regex, Captures, CaptureMatches};

struct SplitCaptures<'r, 't> {
    finder: CaptureMatches<'r, 't>,
    text: &'t str,
    last: usize,
    caps: Option<Captures<'t>>,
}

impl<'r, 't> SplitCaptures<'r, 't> {
    fn new(re: &'r Regex, text: &'t str) -> SplitCaptures<'r, 't> {
        SplitCaptures {
            finder: re.captures_iter(text),
            text: text,
            last: 0,
            caps: None,
        }
    }
}

#[derive(Debug)]
enum SplitState<'t> {
    Unmatched(&'t str),
[...]

Issue #285, opened 2016-09-20
Split with matches


Keats:
I was trying to implement http://jinja.pocoo.org/docs/dev/templates/#title and I was surprised to realise that matches were not included on `split` like they are in js or python.
In JS:
```js
"foo bar".split(/(\s+)/)
["foo", " ", "bar"]
```

Is it possible to get a `split_including_matches` that matches that behaviour?
My naive implementation of the filter was:

```rust
static ref WORD_BEGINNING_SPLIT_RE: Regex = Regex::new(r"([-\s\(\{\[<]+)").unwrap();

&WORD_BEGINNING_SPLIT_RE
        .split(&s)
        .map(|item| {
            let mut c = item.chars();
            match c.next() {
                None => "".to_string(),
                Some(f) => f.to_uppercase().collect::<String>() + &c.as_str().to_lowercase()
            }
        })
        .collect::<Vec<_>>()
        .join("")
```
but since it doesn't include the matches, if I join at the end I will have lost all the whitespaces.

The current implementation takes 50 lines to do that https://github.com/Keats/tera/blob/master/src/filters/string.rs#L105-L155

BurntSushi:
I'm not sure I follow. Could you be more explicit with an example that includes your expected output?

As for Python, it looks like its behavior matches this library:

```
>>> import re
>>> re.split('\s+', 'this  is a \n\ntest')
['this', 'is', 'a', 'test']
```

Javascript also works similarly:

```
"this   is a \n\ntest".split(/\s+/)
["this", "is", "a", "test"]
```

Hmm... You're using a capturing group, I wonder if that plays a role:

```
"this   is a \n\ntest".split(/(\s+)/)
["this", "   ", "is", " ", "a", " ↵↵", "test"]
```

Ah ha! Same for Python too:

```
>>> import re
>>> re.split('(\s+)', 'this  is a \n\ntest')
['this', '  ', 'is', ' ', 'a', ' \n\n', 'test']
```

The key part of the Python docs is:

> If capturing parentheses are used in pattern, then the text of all groups in the pattern are also returned as part of the resulting list.

This seems a little strange to me, and honestly, I haven't read through your particular problem closely. I should have time for a more careful review later today and hopefully come up with a better answer.

BurntSushi:
@Keats Here's how I'd do it:

```rust
#[macro_use]
extern crate lazy_static;
extern crate regex;

use regex::Regex;

pub fn title(s: &str) -> String {
[...]

PR #917, opened 2022-10-28
Add split_inclusive

### LINKED ISSUE #681: Introduce full_split to API
This issue has been brought up before in #285 and #330 but I think it might be worth revisiting.

I think it might be useful to introduce a `full_split` method on a `Regex`. This would behave similar to the current `split` method, but would also return values that match the regex. The iterator would return an enum for every iteration, either a `Delim` (match) or a `Text` (non-match).

This could have a few helpful applications. In #330 the author suggested they were using it in some kind of calculator. Personally, I would use this for tokenizing. In the same issue, there was a suggested fix, but I think it might be helpful to include it into this crate officially.

I've based the names on OCaml's own regex api ([seen here](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Str.html#VALfull_split))

archer884:
This adds a new iterator type and a new method to both the byte and unicode version of `Regex`. Most of the code is copied from the ordinary split iterator. My doc example is probably terrible, but who doesn't like fruit?

Before opening this PR, I checked to see whether it's possible to mimic this behavior using, say, a non-capturing group with the regular split. If it is, I didn't manage to do it correctly, so maybe this is still useful to someone. No idea. In my own work, where I have wanted this, I have also wanted the captures associated with the pattern, so that might be a more useful contract than what is presented here.

Closes #681

### DIFF
--- a/src/re_unicode.rs
+++ b/src/re_unicode.rs
@@ -371,6 +371,34 @@ impl Regex {
         Split { finder: self.find_iter(text), last: 0 }
     }
 
+    /// Returns an iterator of substrings of `text` delimited by a match of the
+    /// regular expression. Each element of the iterator will include the
+    /// delimiting match if it appears at the beginning of the element.
+    ///
+    /// This method will *not* copy the text given.
+    ///
+    /// # Example
+    ///
+    /// To split a string delimited by fruit and include the fruit:
+    ///
+    /// ```rust
+    /// # use regex::Regex;
+    /// # fn main() {
+    /// let re = Regex::new(r"(apple|banana|pear)").unwrap();
+    /// let fields: Vec<&str> = re
+    ///     .split_inclusive("apples: 3 bananas: 2 pears: 4")
+    ///     .map(|s| s.trim())
+    ///     .collect();
+    /// assert_eq!(fields, vec!["", "apples: 3", "bananas: 2", "pears: 4"]);
+    /// # }
+    /// ```
+    pub fn split_inclusive<'r, 't>(
+        &'r self,
+        text: &'t str,
+    ) -> SplitInclusive<'r, 't> {
+        SplitInclusive { finder: self.find_iter(text), last: 0 }
+    }
+
     /// Returns an iterator of at most `limit` substrings of `text` delimited
     /// by a match of the regular expression. (A `limit` of `0` will return no
     /// substrings.) Namely, each element of the iterator corresponds to text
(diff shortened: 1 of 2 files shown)


src/re_unicode.rs
```diff
@@ -371,6 +371,31 @@ impl Regex {
...
+    /// # Example
+    ///
+    /// To split a string delimited by fruit and include the fruit:
+    ///
+    /// ```rust
+    /// # use regex::Regex;
+    /// # fn main() {
+    /// let re = Regex::new(r"(apple|banana|pear)").unwrap();
+    /// let fields: Vec<&str> = re.split("apples: 3 bananas: 2 pears: 4").collect();
+    /// assert_eq!(fields, vec!["apples: 3 ", "bananas: 2 ", "pears: 4"]);
```
BurntSushi, line 389: This does not look like what I'd expect it to be? From the std docs:

> An iterator over substrings of this string slice, separated by characters matched by a pattern. Differs from the iterator produced by split in that split_inclusive leaves the matched part as the terminator of the substring.

The matched part in your expected result is definitely not treated as a terminator here.

And yeah I do kind of think we could do better with an example here. I might copy std's example, but tweak it to match `\r?\n` or something like that.

src/re_unicode.rs
```diff
@@ -371,6 +371,31 @@ impl Regex {
...
+    /// This method will *not* copy the text given.
+    ///
+    /// # Example
+    ///
+    /// To split a string delimited by fruit and include the fruit:
+    ///
+    /// ```rust
+    /// # use regex::Regex;
+    /// # fn main() {
+    /// let re = Regex::new(r"(apple|banana|pear)").unwrap();
```
BurntSushi, line 387: Note that the parentheses here are superfluous.

archer884:
Apologies; I didn't realize I had to run a different command to run the doc tests. I'm fixing those now.

BurntSushi:
`cargo test` should run them automatically.

archer884:
> `cargo test` should run them automatically.

That's what I'd have thought. It's possible there are just so many tests that I didn't see the failures as they scrolled past. I was expecting to see them reported at the end and might just not have been vigilant enough. Anyway, I looked a lot more closely this last time. :p

archer884:
I've attempted to correct the behavior of the iterator to match the behavior expressed by split_inclusive in std. I've also attempted to make the docs less awful.

src/re_unicode.rs
archer884: I like the `\r?\n` idea. Gonna make that change....

BurntSushi:
Why was this closed?

archer884:
Well, there was no response for a week, and I don't like leaving things in limbo.

BurntSushi:
Fair enough. It would help communicating your expectations up front next time, and I would be able to tell you right away that I won't be able to live up to them.

archer884:
/shrugSent from my iPhoneOn Nov 9, 2022, at 5:30 AM, Andrew Gallant ***@***.***> wrote:﻿
Fair enough. It would help communicating your expectations up front next time, and I would be able to tell you right away that I won't be able to live up to them.

shner-elmo:
I'm also one of the many users struggling with importing regex patterns from Python to Rust, 
after reading #285, #330, and #681, I was really hoping for this feature to be added.

If anybody is still interested in implementing it, let me know, I would love to work on it!

BurntSushi:
@shner-elmo We should keep discussion about this on the actual issue https://github.com/rust-lang/regex/issues/681#issuecomment-1747643806. Not on closed PRs.

PR #930, opened 2022-11-28
Speed up replacen loop

xkr47:
`replacen()` (and therefore also `replace()`) was postponing checking against the `limit` until next iteration, which causes it to unnecessarily execute a "find" one extra time after the limit has been reached.

Also added two unit tests for `replacen()` since there were none.


BurntSushi (approved):
Nice find!

PR #940, opened 2023-01-03
AsRef<str> trait impl.

Siiir:


BurntSushi:
I'm not sure this is a desirable implementation. And it is, IMO, odd to want this. A `Match` isn't _just_ a `str`, it's also combined with match offsets. I also feel like adding this impl is likely to cause people to use `as_ref()` when they probably shouldn't to get the matched string, and thus open themselves up to inference breakages.

I'm going close this. But if you have a good use case for why you need this, please open a new issue where this addition can be properly discussed.

PR #942, opened 2023-01-05
capi: Add missing void

thechampagne:


BurntSushi:
It would be helpful in the future to add some explanation or motivation for a change. Instead, I found this: https://softwareengineering.stackexchange.com/questions/286490/what-is-the-difference-between-function-and-functionvoid

So basically, in C, an empty parameter list means "the function takes an unspecified number of arguments." (lol.) But an explicit `void` means "the function takes zero arguments." The latter is indeed what we want here. Thanks!

Issue #955, opened 2023-02-08
Named capture group syntax: `(?<name>exp)`


01mf02:
Would it be possible to support in addition to the existing syntax `(?P<name>exp)` for named capture groups also the syntax `(?<name>exp)`?

My use case for this is that I am currently writing a jq clone called [jaq](https://github.com/01mf02/jaq). Recently, I have added support for regular expressions to jaq using the `regex` crate, which works very well. However, because jq supports only the `(?<name>exp)` syntax (because of the [`oniguruma`](https://github.com/kkos/oniguruma) library) and jaq only the `(?P<name>exp)` syntax (because of the `regex` crate), it is currently impossible to write regexes with named capture groups that are valid in both jaq and jq.

Apart from this, the `(?<name>exp)` syntax seems [reasonably popular](https://www.regular-expressions.info/named.html), so apart from my special use case, it might make sense to add support for this syntax. :)

BurntSushi:
To get this out of the way: this would be a backwards compatible change because currently all forms of `(?<name>exp)` are invalid syntax due to `<` being interpreted as a flag. Since `<` is of course _not_ a flag, it always fails.

I think the original reason why I didn't do this was to 1) match RE2 syntax and 2) I didn't want two ways of doing the same thing. But I do think my stance has softened somewhat over time on this. Another related example is that I plan to relax the escaping rules in order to make the differences between at least the surface level syntax smaller between other regex engines. (For example, right now `\/` is forbidden.)

I'll also say though that what you're facing here is a surface level problem. There are assuredly many other differences between this regex engine and Oniguruma. How will you deal with those? If compatibility is your ultimate goal, then you probably just need to use Oniguruma itself. Or do you see this more as a "let's get _some_ compatibility, but not all of it" sort of situation? The problem there is that there may be many incompatibilities that are totally silent. (I don't have an Oniguruma environment that I can easily test with at the moment.)

I'll note that RE2 specifically only implements `(?P<name>exp)` syntax and there is this comment in the parser:

```
  // Check for named captures, first introduced in Python's regexp library.
  // As usual, there are three slightly different syntaxes:
  //
  //   (?P<name>expr)   the original, introduced by Python
  //   (?<name>expr)    the .NET alteration, adopted by Perl 5.10
  //   (?'name'expr)    another .NET alteration, adopted by Perl 5.10
  //
  // Perl 5.10 gave in and implemented the Python version too,
  // but they claim that the last two are the preferred forms.
[...]

PR #956, opened 2023-02-10
Support `(?<` syntax for named capture groups

01mf02:
In #955, we have discussed the inclusion of a new syntax, `(?<name>exp)` in addition to the existing one `(?P<name>exp)`.
This should make it easier for people coming from environments where `(?P<` is not supported (such as Java, JavaScript, .NET, Ruby, Boost) to use `regex`.

This PR implements the change and ensures that round-tripping of the AST is preserved.

### DIFF
--- a/regex-syntax/src/ast/mod.rs
+++ b/regex-syntax/src/ast/mod.rs
@@ -1203,7 +1203,9 @@ impl Group {
     /// Returns true if and only if this group is capturing.
     pub fn is_capturing(&self) -> bool {
         match self.kind {
-            GroupKind::CaptureIndex(_) | GroupKind::CaptureName(_) => true,
+            GroupKind::CaptureIndex(_)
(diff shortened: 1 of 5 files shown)


BurntSushi (approved):
Nice, I think this pretty much LGTM as it is.

BurntSushi:
I've pulled this into my [`ag/regex-automata` WIP branch](https://github.com/rust-lang/regex/compare/ag/regex-automata) (which I occasionally force push to). I expect I'll get it released within the next few months, but I don't make any promises. It is batched in with a whole bunch of other changes.

01mf02:
I'm happy to hear that. Thanks for accepting my work! :)

BurntSushi:
I would like to leave this open to reflect current status. It should close once my branch lands on master.

01mf02:
I see, sorry.

PR #978, opened 2023-04-18
rewrite the regex crate

BurntSushi:
This effectively replaces the existing `regex` crate internals with public APIs (that anyone can use) from `regex-automata`.

I'll fill our this PR a little bit more, but I don't intend to merge this for a bit until current `master` has been released. I have this up in a PR primarily to make sure CI is working as intended, and in particular, to ensure that the 1.8 release is actually sufficient preparation for the 1.9 release. (It would be bad to put out 1.8 and then realize I need another breaking change in `regex-syntax` for example in order to wire everything up.)

For those who want to browse the `regex-automata` API, you can do so here (top-level docs haven't been written yet): https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/


lopopolo:
@BurntSushi I'm not sure if this is something you're interested in, but using `git subtree` you can import `regex-automata` and all of its history into the `regex` repository: https://manpages.ubuntu.com/manpages/bionic/man1/git-subtree.1.html

BurntSushi:
@lopopolo Aye thanks for that idea. I'm going to pass for now primarily because the `regex-automata` I'm importing here is basically a total rewrite. It doesn't really have much meaningful history going back to `0.1`. It's a nearly totally different thing.

addisoncrump:
I was working on (yet another!) fuzzer for this when I realised I was way, way out of my depth. The feature space of the new regex is far larger than I expected :sweat_smile: 

The effective implementation was to make a differential fuzzer for the various optimisation strategies + the combinational strategy. My `use`s look like this:
```
use regex_automata::dfa::regex::Regex as DfaRegex;
use regex_automata::hybrid::regex::Regex as HybridRegex;
use regex_automata::meta::Regex as MetaRegex;
use regex_automata::nfa::thompson::pikevm::PikeVM as NfaRegex;
```

However, as I was writing this, I realised that DfaRegex internally refers to the PikeVM implementation. It's at this point that I am simply going to say "I'm not sure where to go from here", as there seems to be a whole ecosystem of regex engines in regex-automata offered by default, each with their own configuration settings (notably, configuring the size limit is not something I've figured out yet, so there are many OOMs... :sweat_smile:). I have posted the code for this fuzzer below:

<details>

<summary>
opt strategy fuzzer (only matching)
</summary>

```rust
#![no_main]

use libfuzzer_sys::fuzz_target;
use regex_automata::dfa::regex::Regex as DfaRegex;
use regex_automata::hybrid::regex::Regex as HybridRegex;
use regex_automata::meta::Regex as MetaRegex;
use regex_automata::nfa::thompson::pikevm::PikeVM as NfaRegex;
use regex_automata::util::iter::Searcher;
use regex_automata::{Input, Match, MatchError};
use regex_syntax::ast::Ast;

#[derive(Eq, PartialEq, arbitrary::Arbitrary)]
struct FuzzData {
    ast: Ast,
    haystack: String,
}

impl std::fmt::Debug for FuzzData {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let mut builder = f.debug_struct("FuzzData");
        builder.field("ast", &format!("{}", self.ast));
        builder.field("haystack", &self.haystack);
        builder.finish()
    }
}

fn do_fuzz(data: FuzzData) {
    let pattern = format!("{}", data.ast);
    let Ok(dfa) = DfaRegex::new(&pattern) else { return };
    let Ok(nfa) = NfaRegex::new(&pattern) else { return };
    let Ok(hybrid) = HybridRegex::new(&pattern) else { return };
    let Ok(meta) = MetaRegex::new(&pattern) else { return };

    let mut m1 = Searcher::new(Input::new(&data.haystack))
        .into_matches_iter(|i| dfa.try_search(i));
    let mut cache = nfa.create_cache();
    let mut m2 =
        Searcher::new(Input::new(&data.haystack)).into_matches_iter(|i| {
            let mut captures = nfa.create_captures();
            nfa.search(&mut cache, i, &mut captures);
            Ok(captures.get_match())
        });
    let mut cache = hybrid.create_cache();
    let mut m3 = Searcher::new(Input::new(&data.haystack))
        .into_matches_iter(|i| hybrid.try_search(&mut cache, i));
    let mut m4 = Searcher::new(Input::new(&data.haystack))
        .into_matches_iter(|i| Ok(meta.search(i)));

    let mut iters: Vec<&mut dyn Iterator<Item = Result<Match, MatchError>>> =
        vec![&mut m1, &mut m2, &mut m3, &mut m4];
    loop {
        let mut i = 0;
        let mut curr = None;
        while i < iters.len() {
            let next = match iters[i].next() {
                Some(Err(_)) => {
                    iters.remove(i);
                    continue;
                }
                Some(Ok(m)) => Some(m),
                None => None,
            };
            i += 1;
            if let Some(base) = curr {
                assert_eq!(base, next);
            } else {
                curr = Some(next)
            };
        }
        if curr.map_or(true, |e| e.is_none()) {
            break;
        }
    }
}

fuzz_target!(|data: FuzzData| { do_fuzz(data) });
```

</details>

I know you're probably really, really fed up with the fuzzers at this point, but there's an interesting opportunity for (non-exhaustive) verification here by way of differential fuzzing between the different optimisation strategies. If you have the time to look into this, I think it would be quite worthwhile. :slightly_smiling_face:

BurntSushi:
> I was working on (yet another!) fuzzer for this when I realised I was way, way out of my depth. The feature space of the new regex is far larger than I expected sweat_smile

Haha indeed.

I don't think I'm going to have time to work on this before `regex 1.9`, but I do agree it is a very interesting work. Could you start a new issue about it?

> However, as I was writing this, I realised that DfaRegex internally refers to the PikeVM implementation.

Hmmm I don't think so. It does use the `NFA` though. The `NFA` is arguably the central data type in this crate because *all* of the regex engines are built from it. The `PikeVM` and `BoundedBacktracker` use it directly, where as `dfa::regex::Regex`, `dfa::onepass::DFA` and `hybrid::regex::Regex` all build additional structures from the NFA that are used for a search.

As for size limits, yeah, all of them are disabled by default. I'm still somewhat thinking about whether that's a good idea or not, but currently strongly leaning towards keeping them disabled by default with a note about it in the docs. But there are a lot of them:

* https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/nfa/thompson/struct.Config.html#method.nfa_size_limit
* https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/dfa/dense/struct.Config.html#method.dfa_size_limit
* https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/dfa/dense/struct.Config.html#method.determinize_size_limit
* https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/dfa/onepass/struct.Config.html#method.size_limit
* https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/hybrid/dfa/struct.Config.html#method.cache_capacity
    * related: https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/hybrid/dfa/struct.Config.html#method.skip_cache_capacity_check
    * related: https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/hybrid/dfa/struct.Config.html#method.minimum_cache_clear_count
    * related: https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/hybrid/dfa/struct.Config.html#method.minimum_bytes_per_state
* https://burntsushi.net/stuff/tmp-do-not-link-me/regex/regex_automata/nfa/thompson/backtrack/struct.Config.html#method.visited_capacity

I think that's all of them? The last few are actually a "capacity" rather than a size limit, and those do have reasonableish defaults by necessity.

addisoncrump:
Another potential issue I observed when I was working on this: I noticed you changed the input shape of `fuzz_regex_match` in this version. That means that all of the existing inputs discovered for it by ClusterFuzz (OSS-Fuzz) will be effectively useless. Is that alright with you? Personally, I would recommend reverting to the previous input strategy, even though it's potentially not the most efficient or mutation-friendly.

BurntSushi:
I think it's probably fine? Do the previous inputs serve as anything other than regression tests? I ask that because whenever fuzzing finds a failure, I usually add an explicit regression test as well.

addisoncrump:
It does matter; fuzzing works by creating new inputs from mutations of the old. The existing corpus serves as a "representative set" of inputs which cover the program, at least as much as has been covered so far. By changing the input shape, you invalidate the corpus, which means that the fuzzer has to rediscover large sections of input which have been previously covered. Potentially, it will be unable to rediscover existing inputs.

BurntSushi:
Oooo interesting. Okay, that is something I did not know. The extent of my fuzzing ignorance is vast.

Okay, I will roll back `fuzz_regex_match` to its original form and then split out the code that's there now into a new fuzz target.

addisoncrump:
Fuzzing has lots of subtle complications :sweat_smile: One of these days, I want to go through and write a "what devs need to know" guide, because the mental model of developers just never seems to match reality (by no fault of their own).

As another example, `arbitrary` technically implements a structured mutator that would generally be considered poor quality in the academic fuzzing world because the mutations applied by libfuzzer can potentially corrupt the structure (because e.g. the length field of a Vec might be mutated, leading to the items after it being derived from totally unrelated data), leading a small mutation to the input to create _massive_ changes in the structured output. This effectively makes the fuzzer a random input tester for many types of mutations, which potentially prevents it/makes it arbitrarily unlikely from exploring some code regions. This is a Very Hard Problem™ to solve, though, and since `arbitrary` is "good enough" for most use cases, there's no real reason to fix this. Fuzzers like [lain](https://github.com/landaire/lain) which is a somewhat limited solution for this, but writing these mutators requires a lot of domain-/target-specific knowledge because you can't write a general solution to this problem.

BurntSushi:
Interesting. Yeah that sounds thorny. There's probably a UX problem here too. The tooling around fuzzing just seems very focused on "actions" instead of focusing on the "data" and the flow of that data through the fuzzing apparatus. But this is a very fuzzy criticism. What I mean is that a lot of the ins and outs of fuzzing are not easily discoverable through use of the tool itself. For example, I just haven't had the time to sit down and read LLVM's fuzzing docs in thorough detail. (And it's not totally clear how many insights I'd come away with if I did.)

PR #991, opened 2023-05-06
Implement Unicode class arbitrary

addisoncrump:
See late discussion in #848


addisoncrump:
I don't think this is quite complete... missing e.g. `Changes_When_Casefolded`

addisoncrump:
I attempted to optimise this further by removing the recomputation of the full unicode table sizes every time the arbitrary was used, but this seemed to actually worsen things, so this is probably Good Enough™. There is potentially a further optimisation by "skipping" members of the unicode table (property values in particular) but I don't think this is worthwhile.

addisoncrump:
Well, the force-push broke the cleanness of this a bit :joy: Would you let me know if you see any issues with the implementation I have so far? I'll port it over some point later.

BurntSushi:
So I tried this out, but I'm tempted to remove the `ast_roundtrip` fuzz target. It is producing waaaaay too many false positives and I just don't have the time to deal with them right now. At a higher level, the functionality that `ast_roundtrip` is testing is not particularly critical. It's a "nice to have" that isn't part of how the regex engines work.

I'll take this PR in for now (I cherry picked the commits), but I'm likely to remove the `ast_roundtrip` fuzzer. We can revisit it later.

Also probably a good idea not to submit more PRs to `ag/regex-automata` unfortunately. I'm still force pushing to it. I don't want to get it on to master yet.

BurntSushi:
Ah okay, the false positives were a result of a flub on my part. I put `env_logger::init()` into each of the fuzzers, and that started panicking when I ran the fuzzer across multiple threads. Since `ast_roundtrip` was the first fuzzer I ran in a multi-threaded way since adding `env_logger::init()`, I falsely assumed it had a bunch of false positives. I'm running it now and all seems good for now.

BurntSushi:
Ah yeah I see a problem. The `Arbitrary` impls are generating capture groups with invalid names:

```
Group(Group { span: Span(Position(o: 72056936891165757, l: 18446743197536222976, c: 18095463302774652928), Position(o: 47261417342, l: 9187201950444158975, c: 9187201950435737471)), kind: CaptureName { starts_with_p: true, name: CaptureName { span: Span(Position(o: 7731356377379602303, l: 15191295558507746002, c: 67961541330), Position(o: 5425479542451994624, l: 18446743297583696715, c: 5425513257945268223)), name: "KkKK>\0\0\u{b}\u{b}\u{b}\u{b}\u{b}\u{b}\u{b}?<\u{b}\u{b}\u{b}\u{b}\u{b}{\u{b}\u{b}\u{b}\u{b}\u{b}\u{b}\u{b}\u{b}\u{b}555555}}}}}}}}}eeee\u{1}\0\0\0\0\0\0,", index: 1263206624 } }, ast: Empty(Span(Position(o: 0, l: 0, c: 0), Position(o: 0, l: 0, c: 0))) })
```

The AST printer assumes that the AST values are valid, including the capture group name, so it dumps it out naively. But there's a `>` in there that closes the capture group name.

I've added another manual `Arbitrary` impl for `CaptureName` to only generate correct capture names.

PR #996, opened 2023-05-22
syntax: fix overflow for big counted repetitions

### LINKED ISSUE #995: integer overflow on a particular regex

1.8.1 (latest)

The following code causes an integer-overflow panic in debug mode:
```rust
fn main() {
    regex::Regex::new("  {2147483516}{2147483416}{5}").unwrap();
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0f0b53f5c300a71ce6098e767d13cbc2

- integer overflow
  - panic in debug mode
  - seemingly returns the right answer (`Err(CompiledTooBig(10485760))`) anyway in release mode

Return the right answer (`Err(CompiledTooBig(10485760))`) without integer overflow.

#### Acknowledgement

The fuzzing infrastructure that found this bug was provided to and integrated into `surrealdb` by Google OSS Fuzz and @silvergasp, respectively.

The consensus from reporting this as a security issue was that it shouldn't be treated as one as users can avoid panics using `catch_unwind`. The security policy and/or documentation may be updated to reflect this.

BurntSushi:
This fixes a bug where the calculation for the min/max length of a regex could overflow if the counted repetitions in the pattern are big enough. The panic only happens when debug assertions are enabled, which means there is no panic by default in release mode.

One may wonder whether other bad things happen in release mode though, since in that case, the arithmetic will wrap around instead. Since this is in new code and since the regex crate doesn't yet utilize the min/max attributes of an Hir, the wrap around in this case is completely innocuous.

Fixes #995

### DIFF
--- a/regex-syntax/src/hir/mod.rs
+++ b/regex-syntax/src/hir/mod.rs
@@ -2481,16 +2481,24 @@ impl Properties {
             props.literal = props.literal && p.is_literal();
             props.alternation_literal =
                 props.alternation_literal && p.is_alternation_literal();
-            if let Some(ref mut minimum_len) = props.minimum_len {
+            if let Some(minimum_len) = props.minimum_len {
                 match p.minimum_len() {
                     None => props.minimum_len = None,
-                    Some(len) => *minimum_len += len,
+                    Some(len) => {
+                        // We use saturating arithmetic here because the
+                        // minimum is just a lower bound. We can't go any
+                        // higher than what our number types permit.
+                        props.minimum_len =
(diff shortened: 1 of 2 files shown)


finnbear:
That was fast :crab: :100:

Thank you for the crate, this patch, and your ongoing effort to clarify the security policy! :ok_hand:

BurntSushi:
No problem! I actually had this fixed in a branch before it was reported, but I just didn't make the connection that it was impacting the regex release. (Because the bug is in new code that isn't used but is actually executed.) So I just had to split this patch out. :-)

PR #997, opened 2023-05-22
Move oss fuzz build scripts + update with debug-assertions

silvergasp:


silvergasp:
I'll follow up shortly with a PR in oss-fuzz. This one will need to be merged before the oss-fuzz one can be.

BurntSushi:
Can you remove the license notice please? I don't think it's needed here. And the Rust project generally doesn't put license notices in the headers of files. It's basically just noise.

silvergasp:
If you don't mind commenting/reviewing/approving here https://github.com/google/oss-fuzz/pull/10369 that'll speed things along once I mark it ready for review.

BurntSushi (approved):
LGTM. So if I want to run more fuzzers, I just copy them into the `$OUT` directory? Neat!

PR #1004, opened 2023-06-02
Replace `lazy_static` example with `std::sync::OnceLock` in `src/lib.rs`

lucatrv:


BurntSushi:
#978 adjusts this to recommend `once_cell`, and I think I'm going to stick with that for the time being. Rust 1.70 was literally released yesterday, so I'm not sure about suggesting APIs from it. More broadly though, I find the `Lazy` API in `once_cell` a bit nicer, and the equivalent API in `std` isn't stabilized yet.

I'll find a way to mention the `std` API in #978 though. Thanks!

PR #1005, opened 2023-06-02
Replace `lazy_static` example with `std::sync::OnceLock` in `README.md`

lucatrv:


BurntSushi:
https://github.com/rust-lang/regex/pull/1004#issuecomment-1574141002

PR #1017, opened 2023-06-23
api: add `visit_concat_in` method `Visitor` traits

plusvic:
The `Visitor` trait already has a `visit_alternation_in` that is called between child nodes of an alternation, both in `ast` and `hir`. This adds a similar method that is called between child nodes of concat.

While using the `regexp_syntax` crate in [YARA-X](https://github.com/VirusTotal/yara-x) I came across the need of this method while traversing the HIR. I think this may be useful for more people.

### DIFF
--- a/regex-syntax/src/ast/visitor.rs
+++ b/regex-syntax/src/ast/visitor.rs
@@ -49,7 +49,7 @@ pub trait Visitor {
     }
 
     /// This method is called between child nodes of a concatenation.
-    fn visit_concat_in(&mut self)-> Result<(), Self::Err> {
+    fn visit_concat_in(&mut self) -> Result<(), Self::Err> {
         Ok(())
     }
 
--- a/regex-syntax/src/ast/visitor.rs
+++ b/regex-syntax/src/ast/visitor.rs
@@ -48,6 +48,11 @@ pub trait Visitor {
         Ok(())
     }
 
+    /// This method is called between child nodes of a concatenation.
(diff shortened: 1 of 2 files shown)


BurntSushi:
> While using the `regexp_syntax` crate in [YARA-X](https://github.com/VirusTotal/yara-x) I came across the need of this method while traversing the HIR.

Can you say more about this?

plusvic:
> > While using the `regexp_syntax` crate in [YARA-X](https://github.com/VirusTotal/yara-x) I came across the need of this method while traversing the HIR.
> 
> Can you say more about this?

Sure. I was trying to use the `Visitor` trait for traversing the HIR while generating code for a regular expression. I wanted to generate code that runs backwards (something similar what the `regex` crate already does internally). However, in order to be able to generate this "backward" code, I need to traverse the child nodes of a `Concat` in reverse order. 

My first idea was adding some configuration for visiting the children of `Concat` in reverse order, but that looked like a very narrow use-case for the visitor API, and since I wanted to generate both the forward and the backward code at the same time, that wouldn't be a solution for me anyways. So, I ended up writing an `Visitor` that is able to put the code of each children of `Concat` in a temporary buffer and then produce the final code by writing the code for each children in reverse order. But in order to implement this visitor, knowing the starting and ending points of a `Concat` is not enough, I need to know when each child has been processed in order to switch from one temporary buffer to the other.

PR #1025, opened 2023-07-05
doc: dfa::Automata: sort PatternIDs before asserting

riking:
The ordering that match_pattern returns the PatternID in is clearly not stable API even if it's deterministic. Sorting the IDs before asserting them makes the test more stable.


BurntSushi:
No, the order of pattern IDs is definitely guaranteed here. Probably the docs need to be improved somewhere.

BurntSushi:
I welcome doc improvements (or ideas for doc improvements) here. The basic idea is that pattern IDs are reported with respect to match semantics and their order matters.

PR #1026, opened 2023-07-06
correct typos

jqnatividad:
Using the [typos-cli](https://crates.io/crates/typos-cli) crate (which of course, uses regex), and typo corrections applied project-wide using VSCode's ripgrep-powered Replace in Files. 😄


BurntSushi (changes requested):
Wow, that's a lot of typos. Hah.

The typo fixes look great (and are appreciated), but I'd prefer dropping the `_typos.toml` config file. I don't think fixing typos is worth having a TOML file for it.

BurntSushi (approved):
Okay, thank you!

Issue #1037, opened 2023-07-12
fuzz_regex_match was not rolled back before merge


addisoncrump:
We previously discussed how [changing the input shape may affect the usefulness of existing inputs discovered by clusterfuzz](https://github.com/rust-lang/regex/pull/978#issuecomment-1573364121) and I think this still holds true. Can we (once again) revert fuzz_regex_match? Hopefully this hasn't caused Clusterfuzz to cleanse the existing corpus too much.

BurntSushi:
Hmmm right. So I think the issue I have with this is that it effectively puts a ceiling on how much each fuzz target can change. That seems pretty sub-optimal to me. How much does re-using the existing corpus really matter? Is it just about confirming that old regressions don't re-appear? If so, I'm usually pretty good about adding separate regression tests to the test suite, so I'm not too worried about that.

addisoncrump:
Okay, to address these individually:
1. Reusing the existing corpus allows the fuzzer to have a starting point for mutating the inputs (since fuzzers work by making changes to existing inputs from the corpus). For Clusterfuzz, this tends to represent CPU-years worth of computation, and is normally a quite effective representation of the potential input space of the program. The corpus is strictly _not_ a regression test suite -- it is composed of inputs which explore the state space of the program well so that mutant inputs are meaningful.
2. I wouldn't worry about restricting fuzzers here -- you may have a corpus for *each* fuzzer in Clusterfuzz, so it is not an issue to have different fuzzers with different input shapes. What I recommend is having one fuzzer with the old pattern (so that you can reuse the exploration of the older version of regex) and adding new fuzzers with new patterns.
3. Restricting yourself to Arbitrary-style fuzzers may prevent some inputs from existing, [as we discussed previously](https://github.com/rust-lang/regex/pull/978#issuecomment-1573628890). I think in this case it's not too much of a concern, but I do worry about the effect of mixing config bytes with bytes that represent regexes.

Paging @jonathanmetzman here -- I'm not sure what the eviction policy for corpora on Clusterfuzz/OSS-Fuzz is.

BurntSushi:
Gotya.

And I just realized that probably none of the new fuzzers are running because this script needs to be updated and hasn't been yet: https://github.com/rust-lang/regex/blob/48daadc0dc5865ced38495258cfffcd3951682e4/fuzz/oss-fuzz-build.sh#L4

I put up a PR to fix this here: https://github.com/rust-lang/regex/pull/1042

[...]

PR #1042, opened 2023-07-13
fuzz: add all fuzzers to OSS-fuzz

BurntSushi:
I forgot to do this step, and as a result, OSS-fuzz hasn't been running any of the new fuzzers. Hopefully this is enough.

cc @addisoncrump 

Ref #1037

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -32,6 +32,9 @@ features = ["atty", "humantime", "termcolor"]
 [workspace]
 members = ["."]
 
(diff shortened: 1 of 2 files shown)


addisoncrump:
Small footnote here: would it be possible to add me to the list of CCs on OSS-Fuzz? I'm similarly curious about the performance of my fuzzers, and would be happy to help with test case reduction going forward.

BurntSushi:
Oh yes I would love that! I opened a PR here: https://github.com/google/oss-fuzz/pull/10702

Issue #777, opened 2021-05-05
the Replacer trait should be parameterized over a lifetime


BurntSushi:
It turns out that the `Replacer` trait is a bit more restrictive than it should be. For the closure impl of `Replacer`, it requires that the closure work for all possible lifetimes. But it's far more useful to be able to provide a closure that works on a specific lifetime. This makes it possible to do things like return a `Cow<str>` from the closure to avoid an extra alloc.

This problem was originally brought up in #775 by @tpoliaw.

That lead to @sgrif submitting #776 that changes the `Replacer` trait to be parameterized over a lifetime.

Unfortunately, it's a breaking change, so this will have to wait for `regex 2.0.0`.

N.B. This is generally an ergonomic problem I think. It is always possible to implement `Replacer` oneself and do something more optimal. @tpoliaw also provided a work-around using a closure as well, although it's a bit clunky (but perhaps not as clunky as implementing `Repalcer`).

JanBeh:
See also #1048, which I think could solve this issue.

PR #1048, opened 2023-07-20
More generic impl of Replacer for closures

JanBeh:
Implement `Replacer` also for closures that return a type which depends on lifetime of the argument. This allows performing a no-op replacement where the closure returns the whole match, for example, without needing to clone the captured string.

See [Conditional regex replacement](https://users.rust-lang.org/t/conditional-regex-replacement/97174?u=jbe) on URLO for the genesis of this approach.

### DIFF
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -2570,6 +2570,20 @@ impl<'a> Replacer for &'a Cow<'a, str> {
     }
 }
 
+/// Blanket implementation of `Replacer` for closures.
+///
+/// This implementation is basically the following, except that the return type
+/// `T` may optionally depend on lifetime `'a`.
+///
+/// ```ignore
+/// impl<F, T> Replacer for F
+/// where
+///     F: for<'a> FnMut(&a Captures<'_>) -> T,
+///     T: AsRef<str>, // `T` may also depend on `'a`, which cannot be expressed easily
+/// {
+///     /* … */
+/// }
+/// ```
 impl<F: for<'a> ReplacerClosure<'a>> Replacer for F {
     fn replace_append(&mut self, caps: &Captures<'_>, dst: &mut String) {
         dst.push_str((*self)(caps).as_ref());
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -2375,7 +2375,7 @@ impl<'c, 'h> core::iter::FusedIterator for SubCaptureMatches<'c, 'h> {}
 /// [`Replacer`].
 pub trait ReplacerClosure<'a>
 where
-    Self: FnMut(&'a Captures<'a>) -> <Self as ReplacerClosure>::Output,
+    Self: FnMut(&'a Captures<'_>) -> <Self as ReplacerClosure<'a>>::Output,
 {
     /// Return type of the closure (may depend on lifetime `'a`).
     type Output: AsRef<str>;
@@ -2383,7 +2383,7 @@ where
 
 impl<'a, F: ?Sized, O> ReplacerClosure<'a> for F
 where
-    F: FnMut(&'a Captures<'a>) -> O,
+    F: FnMut(&'a Captures<'_>) -> O,
     O: AsRef<str>,
 {
     type Output = O;
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -2371,23 +2371,27 @@ impl<'c, 'h> ExactSizeIterator for SubCaptureMatches<'c, 'h> {}
 
 impl<'c, 'h> core::iter::FusedIterator for SubCaptureMatches<'c, 'h> {}
 
-/// If a closure implements this for all `'a`, then it also implements
-/// [`Replacer`].
-pub trait ReplacerClosure<'a>
-where
-    Self: FnMut(&'a Captures<'_>) -> <Self as ReplacerClosure<'a>>::Output,
-{
-    /// Return type of the closure (may depend on lifetime `'a`).
-    type Output: AsRef<str>;
+/// Contains helper trait for blanket implementation for [`Replacer`].
+mod replacer_closure {
+    use super::*;
+    /// If a closure implements this for all `'a` and `'b`, then it also
+    /// implements [`Replacer`].
+    pub trait ReplacerClosure<'a>
+    where
+        Self: FnMut(&'a Captures<'_>) -> <Self as ReplacerClosure<'a>>::Output,
+    {
+        /// Return type of the closure (may depend on lifetime `'a`).
+        type Output: AsRef<str>;
+    }
+    impl<'a, F: ?Sized, O> ReplacerClosure<'a> for F
+    where
+        F: FnMut(&'a Captures<'_>) -> O,
+        O: AsRef<str>,
+    {
+        type Output = O;
+    }
 }
-
-impl<'a, F: ?Sized, O> ReplacerClosure<'a> for F
-where
-    F: FnMut(&'a Captures<'_>) -> O,
-    O: AsRef<str>,
-{
-    type Output = O;
-}
+use replacer_closure::*;
 
 /// A trait for types that can be used to replace matches in a haystack.
 ///
@@ -2420,6 +2424,69 @@ where
 /// let re = Regex::new(r"(?<last>[^,\s]+),\s+(?<first>\S+)").unwrap();
 /// let result = re.replace("Springsteen, Bruce", NameSwapper);
 /// assert_eq!(result, "Bruce Springsteen");
+/// ```
+///
+/// # Implementation by closures
+///
+/// Closures that take an argument of type  `&'a Captures<'b>` for any `'a` and
+/// `'b: 'a` and which return a type `T: AsRef<str>` (that may depend on `'a`)
+/// implement the `Replacer` trait through a blanket implementation.
+///
+/// A simple example looks like this:
+///
+/// ```
+/// use regex::{Captures, Regex};
+///
+/// let re = Regex::new(r"[0-9]+").unwrap();
+/// let result = re.replace_all("1234,12345", |caps: &Captures<'_>| {
+///     format!("[number with {} digits]", caps[0].len())
+/// });
+/// assert_eq!(result, "[number with 4 digits],[number with 5 digits]");
+/// ```
+///
+/// Note that the return type of the closure may depend on the lifetime of the
+/// reference that is passed as an argument to the closure. This requires the
+/// closure to be a function, unless [closure lifetime binders] are being used:
+///
+/// [closure lifetime binders]: https://rust-lang.github.io/rfcs/3216-closure-lifetime-binder.html
+/// [`Cow`]: std::borrow::Cow
+///
+/// ```
+/// use regex::{Captures, Regex, Replacer};
+/// use std::borrow::Cow;
+///
+/// let re = Regex::new(r"[0-9]+").unwrap();
+/// fn prepend_odd<'a>(caps: &'a Captures<'_>) -> Cow<'a, str> {
+///     if caps[0].len() % 2 == 1 {
+///         Cow::Owned(format!("0{}", &caps[0]))
+///     } else {
+///         Cow::Borrowed(&caps[0])
+///     }
+/// }
+/// let result = re.replace_all("1234,12345", prepend_odd);
+/// assert_eq!(result, "1234,012345");
+/// ```
+///
+/// The same example using closure lifetime binders:
+///
+/// ```
+/// #![feature(closure_lifetime_binder)]
+///
+/// use regex::{Captures, Regex, Replacer};
+/// use std::borrow::Cow;
+///
+/// let re = Regex::new(r"[0-9]+").unwrap();
+/// let result = re.replace_all(
+///     "1234,12345",
+///     for<'a, 'b> |caps: &'a Captures<'b>| -> Cow<'a, str> {
+///         if caps[0].len() % 2 == 1 {
+///             Cow::Owned(format!("0{}", &caps[0]))
+///         } else {
+///             Cow::Borrowed(&caps[0])
+///         }
+///     },
+/// );
+/// assert_eq!(result, "1234,012345");
 /// ```
 pub trait Replacer {
     /// Appends possibly empty data to `dst` to replace the current match.
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -2374,8 +2374,8 @@ impl<'c, 'h> core::iter::FusedIterator for SubCaptureMatches<'c, 'h> {}
 /// Contains helper trait for blanket implementation for [`Replacer`].
 mod replacer_closure {
     use super::*;
-    /// If a closure implements this for all `'a` and `'b`, then it also
-    /// implements [`Replacer`].
+    /// If a closure implements this for all `'a`, then it also implements
+    /// [`Replacer`].
     pub trait ReplacerClosure<'a>
     where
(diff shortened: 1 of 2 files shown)


BurntSushi:
This is interesting. Can you explain how it works? Also, adding a doc test to confirm it fixes what you think it does would be helpful. (Although I'm unlikely to merge this, see below.)

One part I'm concerned about here is `&'a Captures<'a>`. To me that means `'a` will be the shorter of the lifetime of `&Captures` and the `haystack`, and it seems like that could lead to borrow errors.

I'm not too keen on this particular solution personally indeed because of the extra public trait here. It seems like a small thing, but it makes something that is already a little difficult to understand even more difficult.

BurntSushi:
Oh, I see from the urlo post that this only helps cases that enable the `closure_lifetime_binder` feature on nightly? I think that would also be insufficient motivation as well for adding this.

JanBeh:
> One part I'm concerned about here is `&'a Captures<'a>`. To me that means `'a` will be the shorter of the lifetime of `&Captures` and the `haystack`, and it seems like that could lead to borrow errors.

I think this should not be an issue due to subtyping, but I fixed it anyway, see adf66a8c152c9eb62f14d72ca646ad76c2711c8d.

That last change requires a slightly different usage now:

```
#![feature(closure_lifetime_binder)]

use regex::{Captures, Regex};

fn main() {
    let s1 = "Hello World!";
    let s2 = Regex::new("World").unwrap().replace_all(s1, "Universe");
    assert_eq!(s2, "Hello Universe!");
    let s3 = Regex::new("World").unwrap().replace_all(
        s1,
        for<'a, 'b> |caps: &'a Captures<'b>| -> &'a str {
            &caps[0] // don't replace
        }
    );
    assert_eq!(s3, "Hello World!");
}
```

> Can you explain how it works? Also, adding a doc test to confirm it fixes what you think it does would be helpful.

I will try to do both later.

> I'm not too keen on this particular solution personally indeed because of the extra public trait here. It seems like a small thing, but it makes something that is already a little difficult to understand even more difficult.

Targeting a potential new `regex 2.0.0`, the question is whether an additional lifetime parameter (as proposed in #776 and #777) would be better. I feel like the extra trait might be the better trade-off (besides not breaking backward compatibility).

However, I agree that the helper trait is a downside and makes the API appear more complex.

> I see from the urlo post that this only helps cases that enable the `closure_lifetime_binder` feature on nightly? I think that would also be insufficient motivation as well for adding this.

In either case, it would probably be wise to wait until/if [`closure_lifetime_binder`](https://rust-lang.github.io/rfcs/3216-closure-lifetime-binder.html) stabilizes.

JanBeh:
> I'm not too keen on this particular solution personally indeed because of the extra public trait here.

That issue is now addressed by hiding the trait in a private module, see a40d8f5dadc91499bb9f525d5fc495fe8759c557. The added documentation of `Replacer` explains for which closures the `Replacer` trait is implemented. I also added corresponding documentation tests.

> […] I would be happy to re-litigate this in more detail if and when a regex 2.0 ever happens.

I think this should rather depend on the `closure_lifetime_binder` feature instead of a new `regex` API. As I said, this proposal shouldn't break the `regex` API.

Note, however, that even using stable Rust, there is a use case. The relaxed bound can be used for *functions* (instead of closures) that return `Cow`, for example (see added documentation test):

```
let re = Regex::new(r"[0-9]+").unwrap();
fn prepend_odd<'a>(caps: &'a Captures<'_>) -> Cow<'a, str> {
    if caps[0].len() % 2 == 1 {
        Cow::Owned(format!("0{}", &caps[0]))
    } else {
        Cow::Borrowed(&caps[0])
    }
}
let result = re.replace_all("1234,12345", prepend_odd);
assert_eq!(result, "1234,012345");
```

JanBeh:
It's also possible to use the relaxed bound for closures on stable Rust as pointed out in [this URLO post](https://users.rust-lang.org/t/conditional-regex-replacement/97174/25?u=jbe). I updated the documentation test, see 43222f2e2e749b5c8c0fa2e5804a13606b8b0f06.

JanBeh:
> Can you explain how it works?

Basically the idea is to work around the problem that when specifying a bound for a closure, e.g. `F: FnMut(&Captures<'_>)` you *have* to specify a return type (as otherwise the closure is expected to return `()`). So, for example, you write `F: FnMut(&Captures<'_>) -> T`. Type `T` has to be declared outside of the bound, which means it can't depend on the anonymous lifetime of the closure's argument.

The idea of this PR is to use a helper trait which allows to add a bound that doesn't specifies the return type `T` of the closure. The return type is accessible through an associated type ([this one](https://github.com/rust-lang/regex/blob/2c8c8ddaa139320e4894ddb094391a13f6123236/src/regex/string.rs#L2383-L2384)). That helper trait is parameterized with a lifetime `'a` that's the lifetime of the `&'a Captures<'_>` reference. Thus, the return type may also depend on that lifetime. In order for a closure to implement `Replacer` it must implement that helper trait for all `'a`.

In the most recent draft, I have hidden the helper trait as an implementation detail in a private module. With 2c8c8ddaa139320e4894ddb094391a13f6123236, a documentation comment explains what the (hidden) complex implementation actually means. Hiding the helper trait should give some leverage when future language features are added to Rust, which might render the helper trait superfluous.

JanBeh:
>> One part I'm concerned about here is `&'a Captures<'a>`. To me that means `'a` will be the shorter of the lifetime of `&Captures` and the `haystack`, and it seems like that could lead to borrow errors.
> 
> I think this should not be an issue due to subtyping, but I fixed it anyway, see [adf66a8](https://github.com/rust-lang/regex/commit/adf66a8c152c9eb62f14d72ca646ad76c2711c8d).
> 
> That last change requires a slightly different usage now: […]

I was undoing that change again in 664a0f2e5a6f843d960b9f7489e225362c34c350, because it is sufficient for a closure to accept `&'a Captures<'a>` for all `'a`. The caller (i.e. the implementation of `Replacer` for closures) can always coerce any `&'a Captures<'b>` to `&'a Captures<'a>` as `Captures<'b>` is covariant over `'b`.

This also allows a closure to return a type that depends either on `'a` or `'b` if it is more generic and has two distinct lifetimes in its argument. Test cases have been added to `tests/misc.rs` for that matter.

JanBeh:
With the recent changes, I would like to know if this has a chance to land (before `regex 2.0.0`).

I tried to address all concerns. Specifically:

* The helper trait is a [private implementation detail](https://github.com/rust-lang/regex/blob/664a0f2e5a6f843d960b9f7489e225362c34c350/src/regex/string.rs#L2374-L2394) now, which is not exposed to the API.
* [Documentation for the blanket implementation](https://github.com/rust-lang/regex/blob/664a0f2e5a6f843d960b9f7489e225362c34c350/src/regex/string.rs#L2575-L2593) describes for which closures `Replacer` is implemented (even though the implementation itself is a hidden detail).
* The changes should be backward compatible; all tests are passing.
* As pointed out above, it's possible to use the generalized implementations of `Replacer` also on stable Rust. The documentation comment and test [shows how this works](https://github.com/rust-lang/regex/blob/664a0f2e5a6f843d960b9f7489e225362c34c350/src/regex/string.rs#L2449-L2476). Using the unstable [`closure_lifetime_binder`](https://rust-lang.github.io/rfcs/3216-closure-lifetime-binder.html) feature is optional and could improve ergonomics in the future (but isn't a blocker).

If there is further interest in this, I could reproduce the changes also for the `bytes` module of the crate (which is basically a copy that's just using `[u8]` instead of `str`, right?).

zjp-CN:
> as Captures<'b> is covariant over 'b

What if `Captures<'b>` becomes incovariant over `'b`? You'll have another breaking change thne.

Beside, it'd be better to keep the lifetime contract right instead of weirdly working.
When you first see `&Captures<'_>`, `'_` and lifetime on `&` are already distinct. To make them the same is overdone IMO.

JanBeh:
> What if `Captures<'b>` becomes incovariant over `'b`?

That would already be a breaking change, so I don' think that's a problem.

> Beside, it'd be better to keep the lifetime contract right instead of weirdly working.

I'm not sure what's "right" here. Consider the already existing [`impl<'a> Replacer for &'a Cow<'a, str>`](https://docs.rs/regex/1.9.1/regex/trait.Replacer.html#impl-Replacer-for-%26'a+Cow%3C'a,+str%3E), which also uses a single lifetime `'a` instead of two lifetimes `'a, 'b: 'a`.

Not sure what's the correct approach. I tried to use two different lifetimes, but then the helper trait `ReplacerClosure` should also be parameterized over both lifetimes (such that the output type may depend on either lifetime).

I tried this:

```diff
-    pub trait ReplacerClosure<'a>
+    pub trait ReplacerClosure<'a, 'b>
     where
-        Self: FnMut(&'a Captures<'a>) -> <Self as ReplacerClosure<'a>>::Output,
+        'b: 'a,
+        Self: FnMut(&'a Captures<'b>) -> <Self as ReplacerClosure<'a, 'b>>::Output,
     {
-        /// Return type of the closure (may depend on lifetime `'a`).
+        /// Return type of the closure (may depend on lifetime `'a` or `'b`).
         type Output: AsRef<str>;
     }
-    impl<'a, F: ?Sized, O> ReplacerClosure<'a> for F
+    impl<'a, 'b, F: ?Sized, O> ReplacerClosure<'a, 'b> for F
     where
-        F: FnMut(&'a Captures<'a>) -> O,
+        'b: 'a,
+        F: FnMut(&'a Captures<'b>) -> O,
         O: AsRef<str>,
     {
```

But then I also have to do that:

```diff
-impl<F: for<'a> ReplacerClosure<'a>> Replacer for F {
+impl<F: for<'a, 'b: 'a> ReplacerClosure<'a, 'b>> Replacer for F {
```

Which isn't valid syntax:

```
error: lifetime bounds cannot be used in this context
    --> src/regex/string.rs:2591:21
     |
2591 | impl<F: for<'a, 'b: 'a> ReplacerClosure<'a, 'b>> Replacer for F {
     |                     ^^
```

If I instead use:

```rust
impl<F: for<'a, 'b> ReplacerClosure<'a, 'b>> Replacer for F {
```

then the compiler won't be able to tell that `F` only needs to implement `ReplacerClosure<'a, 'b>` for `'b: 'a`; and then several tests fail, e.g.:

```
error: implementation of `regex::regex::string::replacer_closure::ReplacerClosure` is not general enough
  --> tests/replace.rs:7:24
   |
7  |               assert_eq!(re.$which($search, $replace), $result);
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ implementation of `regex::regex::string::replacer_closure::ReplacerClosure` is not general enough
...
84 | / replace!(
85 | |     closure_returning_reference,
86 | |     replace,
87 | |     r"([0-9]+)",
...  |
90 | |     "age: 2"
91 | | );
   | |_- in this macro invocation
   |
   = note: `[closure@tests/replace.rs:89:5: 89:37]` must implement `regex::regex::string::replacer_closure::ReplacerClosure<'0, '1>`, for any two lifetimes `'0` and `'1`...
   = note: ...but it actually implements `regex::regex::string::replacer_closure::ReplacerClosure<'_, '2>`, for some specific lifetime `'2`
   = note: this error originates in the macro `replace` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I believed using a single lifetime instead of two lifetimes is okay here (which avoids these problems). But I'm not sure.

BurntSushi:
Thanks for the explanation, that was helpful. To say where I'm at currently:

* `ReplacerClosure` is still part of the public API because it shows up in a bound in a public API. I don't see a way around this. Making it so users of the crate can't import has the advantage of not confusing the public API, but it also has the disadvantage of removing options from the user in case it would be useful to use it in bounds of their own.
* It's not totally clear to me that this isn't a breaking change. I would generally expect the kind of breakage to happen here to be inference related, if at all. Inference breakages are something I could tolerate, but only if they are incredibly rare. Testing this sort of change in `std` is easier because of crater, but I don't have a tool like that to do that sort of testing before unleashing this change on to everyone.
* Looking at this holistically, I don't think the juice is worth the squeeze. The docs in this PR talk about unstable features and show examples with weird `coerce` function hacks. That's not the kind of thing I want to see in the `regex` crate personally. I'd much rather tell people to go write their own replacement routine then to deal with these type system shenanigans.

JanBeh:
@BurntSushi thanks for your feedback.

> * `ReplacerClosure` is still part of the public API because it shows up in a bound in a public API.

`rustdoc` displays it as a non-linked trait in regard to the blanket implementation of `Replacer` for `F`. However, as it can't be named from outside the crate, it doesn't really affect the users of the API in any way except in regard to which closures implement `Replacer`. The trait that's really exposed to the user of the API is `Replacer` (before and after the changes), and not `ReplacerClosure`.

Nonetheless, having an unnameable trait showing up in the API can be considered somewhat ugly (similar to the sealed trait pattern that's commonly used).

> * It's not totally clear to me that this isn't a breaking change. I would generally expect the kind of breakage to happen here to be inference related, if at all. Inference breakages are something I could tolerate, but only if they are incredibly rare. Testing this sort of change in std is easier because of crater, but I don't have a tool like that to do that sort of testing before unleashing this change on to everyone.

I understand, and I don't fully overlook this either.

> * Looking at this holistically, I don't think the juice is worth the squeeze. The docs in this PR talk about unstable features and show examples with weird coerce function hacks. That's not the kind of thing I want to see in the regex crate personally. I'd much rather tell people to go write their own replacement routine then to deal with these type system shenanigans.

Would you reconsider this if/when [`closure_lifetime_binder`](https://github.com/rust-lang/rust/issues/97362) stabilizes?

I would propose to [merge this](https://github.com/rust-lang/regex/commit/6c8b5497f63c25c3bec8107f734813ba26d2f237) to make the current approach more "clean" and to highlight the open issues, and then close this PR for future reference. It could be reopened or recreated when there is some progress on that matter.

JanBeh:
Thanks to [this post](https://users.rust-lang.org/t/why-cant-i-use-lifetime-bounds-in-hrtbs/97277/6?u=jbe) I have been able to refactor the code to use two different lifetimes `'a` and `'b` in `&'a Captures<'b>` (d89b31a18637e0cecfef1f1d4a73164cb7a9d16d).

> * The docs in this PR talk about unstable features and show examples with weird `coerce` function hacks.

I have improved that (1aab0b8fdad2ef182c30fb600ff1e19d5c70bc71). The example code isn't using any hacks anymore (but simply a straight-forward function).

BurntSushi:
> Would you reconsider this if/when [`closure_lifetime_binder`](https://github.com/rust-lang/rust/issues/97362) stabilizes?
> 
> I would propose to [add these changes to the draft](https://github.com/rust-lang/regex/commit/6c8b5497f63c25c3bec8107f734813ba26d2f237) to make the current approach more "clean" and to highlight the open issues, and then close this PR for future reference. It could be reopened or recreated when there is some progress on that matter.

SGTM. I'm open to re-considering, but probably some effort will need to be put into ensuring this isn't a breaking change. (I don't necessarily mean "technically breaking" but rather "practically breaking.")

The existence of the helper trait is a bummer but not a blocker for me. The breaking changes and the requirement to use `coerce` function hacks are bigger issues to me. What I'd want here is for the obvious thing to "just work." Otherwise it's probably too much ceremony to be worth doing IMO.

JanBeh:
Closing this then. Could be reconsidered when

* https://github.com/rust-lang/rust/issues/97362

or a similar mechanism is implemented, which makes using such closures more idiomatic.

Thanks for all the feedback and assessment.

PR #1049, opened 2023-07-21
Update the old UTF-8 automaton algorithm in comment

Licheam:
The old codes in comment are outdated, so is the regex-cli command. I did some modification fixing it. It runs as expected in comment now when replacing this code for UTF-8 automaton.


BurntSushi (approved):
Nice, thank you. Yeah `regex-cli` went through a few different iterations before its initial release. :-)

PR #1051, opened 2023-07-21
syntax: optimize some functions in IntervalSet.

Licheam:
Replace the canonicalize function the in-place way with constant memory.
Replace the negate function the in-place way with constant memory.
Simplify the difference function without changing the algorithm.
Implement symmetric_difference function like others.
Implement push function like others.
Force the symmetric_difference function in Interval return ranges in order.

### DIFF
--- a/regex-syntax/src/hir/interval.rs
+++ b/regex-syntax/src/hir/interval.rs
@@ -236,7 +236,7 @@ impl<I: Interval> IntervalSet<I> {
         }
 
         self.ranges.drain(..drain_end);
-        self.folded = self.folded && other.folded;
+        self.folded = self.ranges.is_empty() || (self.folded && other.folded);
     }
 
     /// Compute the symmetric difference of the two sets, in place.
--- a/regex-syntax/src/hir/interval.rs
+++ b/regex-syntax/src/hir/interval.rs
@@ -321,28 +321,29 @@ impl<I: Interval> IntervalSet<I> {
             return;
         }
 
-        // There should be a way to do this in-place with constant memory,
-        // but I couldn't figure out a simple way to do it. So just append
-        // the negation to the end of this range, and then drain it before
-        // we're done.
-        let drain_end = self.ranges.len();
-
         // We do checked arithmetic below because of the canonical ordering
         // invariant.
         if self.ranges[0].lower() > I::Bound::min_value() {
-            let upper = self.ranges[0].lower().decrement();
-            self.ranges.push(I::create(I::Bound::min_value(), upper));
+            let mut pre_upper = self.ranges[0].upper();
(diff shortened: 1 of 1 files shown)


regex-syntax/src/hir/interval.rs
```diff
@@ -347,24 +347,19 @@ impl<I: Interval> IntervalSet<I> {
...
-            // merge this range in the previously added range.
-            if self.ranges.len() > drain_end {
-                let (last, rest) = self.ranges.split_last_mut().unwrap();
-                if let Some(union) = last.union(&rest[oldi]) {
-                    *last = union;
-                    continue;
-                }
+        // We consistently try to merge range with previous range
+        // and merge them if possible. Otherwise, we make it the
+        // range as the last one.
```
BurntSushi, line 352: Could you clean this comment up please? It's a little hard to parse. A high level comment explaining why this works would also be helpful.

BurntSushi (changes requested):
w00t! This is awesome. Nice work. :-)

regex-syntax/src/hir/interval.rs
Licheam: Just rewrote the comments, hope it is more clear.

BurntSushi:
I see you're working on more stuff here. If possible, please avoid merge commits. Instead, please rebase and force push.

Licheam:
> I see you're working on more stuff here. If possible, please avoid merge commits. Instead, please rebase and force push.

Sorry about the inconvenience I brought here, and thank you for your professional advices. I am still learning to contribute codes in public repo. The work on this PR is down from my perspective and sorry again for my inexperienced in git.

BurntSushi:
No need to apologize! Everyone deals with git and commit history a little differently. If you look at the commit log for this repo, it's a straight linear line with (usually) decent messages. Basically, if you can just avoid merges then it should be easier for me to massage your PR.

Licheam:
I see. Thx for the explanation. I'll pay attention to this point in future potential PRs.

BurntSushi (approved):
Thank you!

Some of this looks... quite gnarly. But I feel pretty confident with the test suite. I also tried some random mutations to provoke test failures, and it did. So that's good too.

This is really great work. I remember banging my head against the wall trying to write these routines without using extra memory and just gave up.

BurntSushi:
FYI, I ended up needing to revert this change because of bugs caught by OSS-fuzz. See https://github.com/rust-lang/regex/pull/1102 for more info.

Licheam:
> FYI, I ended up needing to revert this change because of bugs caught by OSS-fuzz. See #1102 for more info.

Based on the test cases, it turns out that the implementation of `push` is not very thoughtful. I refractor the whole `push` function making it more direct and readable while fixing the bugs found. Thx for the test cases.

PR #1053, opened 2023-07-23
automata/doc: update outdated regex-cli test command

Licheam:


regex-automata/src/dfa/dense.rs
```diff
@@ -1230,8 +1230,8 @@ impl Builder {
             //
             // Test case:
             //
-            //   regex-cli find hybrid regex -w @conn.json.1000x.log \
-            //     '^#' '\b10\.55\.182\.100\b'
+            //   regex-cli debug dense dfa --unicode-word-boundary \
+            //   @conn.json.1000x.log '^#' '\b10\.55\.182\.100\b'
```
BurntSushi, line 1234: This isn't quite right. You've changed a `find` command to a `debug` command. `find` executes a search. `debug` just prints a human readable description of an object from the regex API.

BurntSushi (changes requested):

regex-automata/src/dfa/dense.rs
Licheam: Oh, I thought it intents to build a dfa coz it's under the AOT dfa build function. But why is hydrid method adopts here in the comment of building full dfa.
Is it `regex-cli find match hybrid --unicode-word-boundary -p '^#' -p '\b10\.55\.182\.100\b' -y @conn.json.1000x.log` or `regex-cli find match dense --unicode-word-boundary -p '^#' -p '\b10\.55\.182\.100\b' -y @conn.json.1000x.log`.

BurntSushi: `find match dense` makes sense to me. That's not the point I was arguing. I was objecting to using the `debug` command.

Licheam: Understood, that's just my confusion arises from encountering such testing command in the first place.

PR #1061, opened 2023-08-04
doc's: change for docs.rs . Update lib.rs

Spiker01:
This code example didn't work. I changed it. Now this code works.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,12 +100,12 @@ Third, edit `src/main.rs`. Delete what's there and replace it with this:
 use regex::Regex;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


BurntSushi:
> This code example didn't work.

Why not elaborate? The code example [does work](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=11589996f7c9e08ccea1b9e45d1cd295). So you're going to need to share more details.

In general, code examples like this in the docs are built and tested. So if it isn't working, then CI would be red.

Spiker01:
I am sorry. It may have been an outdated Rustc 1.6x.x on my computer. I updated rust and its dependencies using rustup. At the moment everything is working. Thank you for attention.

PR #1068, opened 2023-08-23
Spelling corrections in docs

gold37a:


BurntSushi (approved):
Thanks! (Not sure why #1067 is also opened, as this seems to supersede it.)

PR #1074, opened 2023-08-27
Add a feature to regex-syntax that enables look-around expressions

AhoyISki:
This PR adds a feature (`look-ahead-and-behind`) that enables the previously prohibited look-ahead and look-behind expressions (`(?=a)`, `(?<=a)`) from producing an `Ast`. This feature is disabled by default, since `regex` does not support these types of expressions, but it can be useful to others that are importing just the `regex-syntax` crate.

By default, nothing actually changes (besides some additional docs that I've added), and the code is essentially the same, returning the same error when these expressions are used. The error message, however, has been changed, instead of saying that these expressions are unsupported, it mentions that you can enable them with this feature.

The PR also returns an error when a look-ahead expression is followed by something, or when a look-behind is preceded by something. The kinds of these errors are, respectively, `ast::ErrorKind::LookAheadBeforeEnd` and `ast::ErrorKind::LookBehindAfterStart`, both inaccessible with the feature disabled.

These errors are smart in the way that they determine if something is concatenated or not, for example,`(start 1)((((?<=start 2))))` is not valid, but `(start 1)|((((?<=start 2))))` is.

This error reporting is, technically speaking, not necessary, as some regex engines don't really care about the position of look-around expressions, but for my use case, this is important. But if you feel like it doesn't belong, I could just move this to another feature or something like that.

Initially this PR was going to just be a standalone crate, and as with all my crates, I ran clippy over the code, and fixed a bunch of clippy warnings, which are also in this crate. There are also some minor functionality changes in some parts of the functions that I modified, which are used even with this feature disabled, but they are purely syntactic and not semantic.

I know that this PR doesn't really make much sense in the context of the `regex` crate, but it seems incorrect to make just a few changes to a crate, and just publish it to `crates.io`, filling the website with two almost identical crates.

### DIFF
--- a/regex-syntax/Cargo.toml
+++ b/regex-syntax/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "regex-syntax"
 version = "0.7.5"  #:version
-authors = ["The Rust Project Developers", "Andrew Gallant <jamslam@gmail.com>"]
+authors = ["Andrew Gallant <jamslam@gmail.com>", "Mateus Yoshiaki <ahoyiski@gmail.com>"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/regex/tree/master/regex-syntax"
 documentation = "https://docs.rs/regex-syntax"
@@ -33,6 +33,8 @@ unicode-gencat = []
 unicode-perl = []
 unicode-script = []
 unicode-segment = []
+
+look-ahead-and-behind = []
 
 [dependencies]
 arbitrary = { version = "1.3.0", features = ["derive"], optional = true }
--- a/regex-syntax/README.md
+++ b/regex-syntax/README.md
@@ -64,6 +64,11 @@ or all of these data tables. If a regular expression attempts to use Unicode
 data that is not available, then an error will occur when translating the `Ast`
 to the `Hir`.
 
+There are also look-aheads and look-behinds (`(?=a)` and `(?<=a)`), which this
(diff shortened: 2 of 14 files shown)


BurntSushi:
> Initially this PR was going to just be a standalone crate

This is what you should do. I realize you said it feels incorrect to you, but I'd instead nudge you to consider a different perspective. For example, my perspective is not "mostly `regex-syntax`, but with one small logical change to support look-around," but rather, "here's a pile of code for something I worked on, now you can own it and maintain it going forward." I... don't want to do that. Look-around isn't something I'm interested in supporting in the `regex` crate or any of its dependencies. It's important to have some kind of scope for a project, and I'm drawing a line in the sand at this point. Sorry.

(And if I did accept this, I'd demand all of the Clippy changes and other things unrelated to the logical change here be removed. You might consider doing that anyway if you want to maintain a fork as I'm guessing it will make maintaining a patch easier.)

> The error message, however, has been changed, instead of saying that these expressions are unsupported, it mentions that you can enable them with this feature.

This would be inappropriate for `regex-syntax` because these errors are surfaced to users of the `regex` crate. They wouldn't be able to enable any feature to get look-around support in `regex` proper. It would be a supremely confusing error message.

I realize it seems like you need to do this regardless, but for the future, it would be good to discuss large PRs before submitting them. My guess is you probably did at least some additional work to prepare this as a PR versus publishing it as your own crate. (Maybe not.)

PR #1084, opened 2023-09-06
capi: remove unnecessary heap allocation

thechampagne:


BurntSushi:
This doesn't remove any heap allocs. `CString::new` accepts an `impl Into<Vec<u8>>`. So if you give it an owned `Vec<u8>`, then it will be used as-is without any new allocs. If you give it a `&[u8]`, as in this PR here, then it will get copied into a `Vec<u8>` by virtue of its `Into<Vec<u8>>` impl. So either way, you get an alloc.

Thank you though!

Issue #1041, opened 2023-07-12
regex v1.9.1 / regex-lite v0.1.0: test failures on 32-bit architectures


decathorpe:
This is a followup from issue #1039 - it appears that similar problems are also present in regex.

Failed doctests are:

```
    src/builders.rs - builders::bytes::RegexBuilder::size_limit (line 1848)
    src/builders.rs - builders::bytes::RegexSetBuilder::size_limit (line 2420)
    src/builders.rs - builders::string::RegexBuilder::size_limit (line 671)
    src/builders.rs - builders::string::RegexSetBuilder::size_limit (line 1238)
    src/regex/bytes.rs - regex::bytes::CaptureLocations (line 2027)
    src/regex/string.rs - regex::string::CaptureLocations (line 2030)
```

The first four due to the same size limit being too high for 32-bit architectures, and the last two are again usize literals that are too big for their data type.

Full output for the failures:

```
failures:
---- src/builders.rs - builders::bytes::RegexBuilder::size_limit (line 1848) stdout ----
Test executable failed (exit status: 101).
stderr:
thread 'main' panicked at 'assertion failed: RegexBuilder::new(r\"\\w\").size_limit(45_000).build().is_err()', src/builders.rs:9:1
stack backtrace:
   0: 0x566abcfe - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h5e3f417bcf02dcd3
   1: 0x566c7030 - core::fmt::write::hffb0d28c8178d868
   2: 0x566a9d41 - std::io::Write::write_fmt::hd9c1e94d8083a97a
   3: 0x566abacf - std::sys_common::backtrace::print::hb33e9d6b17b3830e
   4: 0x566ad065 - std::panicking::default_hook::{{closure}}::hf736b577d9673442
   5: 0x566ace0e - std::panicking::default_hook::h9d3c62164da10493
   6: 0x566ad63f - std::panicking::rust_panic_with_hook::he6cb4c574f9b74cc
   7: 0x566ad4d2 - std::panicking::begin_panic_handler::{{closure}}::h3fc1035efef798c2
   8: 0x566ac185 - std::sys_common::backtrace::__rust_end_short_backtrace::h581f3c2cbdbe7ea7
   9: 0x566ad242 - rust_begin_unwind
  10: 0x565edb65 - core::panicking::panic_fmt::ha43a609b26c80be9
[...]

PR #1101, opened 2023-10-10
regex-lite: fix compilation of doctests on 32-bit architectures

decathorpe:
This is a followup for #1041. Returning early on non-64-bit architectures is not enough in this case, since the doctest actually fails to compile due to two numeric literals being too large for `usize` on 32-bit architectures.

With this change, "lib" and "doc" tests for the the regex-lite crate pass on all CPU architectures that I have access to.


BurntSushi:
This PR is on crates.io in `regex-lite 0.1.3`.

decathorpe:
Yeah, I agree that this is not ideal :( Thanks for merging + the new release!

Issue #1103, opened 2023-10-12
`Regex::is_match` unexpectedly returning false with regex-syntax v0.8.0


edmorley:
Hi 😄 

First, thank you for a fantastic crate!

#### What version of regex are you using?

```
$ cargo tree
testcasev0.1.0 (/Users/emorley/src/testcase)
└── regex v1.10.0
    ├── aho-corasick v1.1.2
    │   └── memchr v2.6.4
    ├── memchr v2.6.4
    ├── regex-automata v0.4.1
    │   ├── aho-corasick v1.1.2 (*)
    │   ├── memchr v2.6.4
    │   └── regex-syntax v0.8.0
    └── regex-syntax v0.8.0
```

#### Describe the bug at a high level.

Using `regex-syntax` v0.8.0, a `Regex::is_match` that previously matched in `regex-syntax` v0.7.4 no longer matches. 

Whilst the issue no longer reproduces as part of the `regex-syntax` v0.8.1 release (I would guess due to #1102), I thought you might still wish to see the testcase in case it covers something that the newly added fuzz tests do not, ready for the future relanding of #1051.

#### What are the steps to reproduce the behavior?

1. `cargo new testcase && cd $_`
2. `cargo add regex@1.10.0`
3. `cargo update -p regex-syntax --precise 0.8.0`
4. Add this to the end of `main.rs`:
    ```rust
    #[cfg(test)]
    mod tests {
        #[test]
        fn testcase() {
            assert!(regex::Regex::new(r"^[[:alnum:]./-]+$")
                .unwrap()
                .is_match("a-b"));
        }
    }
    ```
5. `cargo test`
6. `cargo update -p regex-syntax --precise 0.8.1`
7. `cargo test`

#### What is the actual behavior?

- `cargo test` at step 5 (ie: when using regex-syntax v0.8.0) fails with the output below
- `cargo test` at step 5 (ie: when using regex-syntax v0.8.1) passes

```
---- tests::testcase stdout ----
thread 'tests::testcase' panicked at src/main.rs:9:9:
assertion failed: regex::Regex::new(r\"^[[:alnum:]./-]+$\").unwrap().is_match(\"a-b\")
```

Note: If I replace the `[:alnum:]` with eg `a-z` then the issue stops reproducing.

#### What is the expected behavior?

`cargo test` passes regardless of regex-syntax version.

BurntSushi:
[...]

PR #1104, opened 2023-10-12
syntax: fix interval set optimizations

Licheam:
Refractor the `push` function in #1051.
This fix the bug found in https://github.com/rust-lang/regex/issues/1103 and https://github.com/rust-lang/regex/issues/1102 based on commit https://github.com/rust-lang/regex/commit/6d2b09ed6fbc136cca007ce0c57ec9cbae16f3b4.


BurntSushi:
I think you'll want to rebase on master.

Given the bugs the last optimizations caused, it's clear that the tests covering these operations are insufficient. Since this code is incredibly subtle, I have to insist that we improve our testing strategy before we start overhauling how these ops are implemented.

Licheam:
Indeed, maybe I gotta design some more delicate test cases for those functions.

PR #1109, opened 2023-10-16
Inline format-args

nyurik:
This makes the code a bit easier to read and smaller.

Some of it was done with this command, and later fixed by hand:

```
 cargo clippy --workspace --allow-dirty --fix --benches --tests --bins -- -A clippy::all -W clippy::uninlined_format_args
 ```

### DIFF
--- a/regex-capi/Cargo.toml
+++ b/regex-capi/Cargo.toml
@@ -11,7 +11,7 @@ description = """
 A C API for Rust's regular expression library.
 """
 workspace = ".."
-edition = "2018"
+edition = "2021"
 
 [lib]
 name = "rure"
--- a/regex-capi/src/macros.rs
+++ b/regex-capi/src/macros.rs
@@ -18,10 +18,7 @@ macro_rules! ffi_fn {
                     } else {
                         "UNABLE TO SHOW RESULT OF PANIC.".to_owned()
                     };
-                    let _ = writeln!(
-                        &mut io::stderr(),
-                        "panic unwind caught, aborting: {:?}",
-                        msg);
+                    let _ = writeln!(&mut io::stderr(), "panic unwind caught, aborting: {msg:?}");
                     unsafe { abort() }
                 }
             }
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

File: regex-capi/Cargo.toml [hunk 1]
```diff
@@ -11,7 +11,7 @@ description = """
 A C API for Rust's regular expression library.
 """
 workspace = ".."
-edition = "2018"
+edition = "2021"
```
BurntSushi, line 14: Probably set `rust-version = "1.65"` for consistency with other crates?

regex-capi/src/macros.rs
```diff
@@ -18,10 +18,7 @@ macro_rules! ffi_fn {
                     } else {
                         "UNABLE TO SHOW RESULT OF PANIC.".to_owned()
                     };
-                    let _ = writeln!(
-                        &mut io::stderr(),
-                        "panic unwind caught, aborting: {:?}",
-                        msg);
+                    let _ = writeln!(&mut io::stderr(), "panic unwind caught, aborting: {msg:?}");
```
BurntSushi, line 21: I follow a 79 column (inclusive) rule. I'm kind of surprised `rustfmt` doesn't take care of this here... Oh, because it's in a macro. So it will need to be wrapped manually.

BurntSushi (changes requested):
Thank you! Just a couple of nits and this should be good to go.

nyurik:
fixed

nyurik:
@BurntSushi hi, friendly ping - let me know if anything else is needed for this and other regex PRs. Thx!

regex-capi/Cargo.toml
nyurik: done

nyurik (commented):
i think this one is done and addressed all feedback.  Let me know if anything else needs fixing

BurntSushi:
Please split the regex-capi changes out from everything else.

PR #1112, opened 2023-10-16
Rename is_escapable_character()

nyurik:
See #1108 - it includes the rest of the rename which wasn't `pub`


BurntSushi:
I don't think this is worth doing. I'll just plan to rename this routine in the next breaking release.

Issue #1113, opened 2023-10-21
regex-automata fails to build for mips targets


brentalanmiller:
#### What version of regex are you using?

It looks like this bug was introduced in regex-automata@0.3.7 and effects all versions through 0.4.3

#### Describe the bug at a high level.

When compiling for `mips-unknown-linux-musl` or `mipsel-unknown-linux-musl` targets, the build will fail with:
```
   Compiling regex-automata v0.3.7
error: Undefined temporary symbol $BB505_1

error: Undefined temporary symbol $BB506_1

error: Undefined temporary symbol $BB507_1

error: could not compile `regex-automata` (lib) due to 3 previous errors
```

#### What are the steps to reproduce the behavior?
```
cargo install cross
cargo new regex-mips-issue
cd regex-mips-issue/
cargo add regex-automata@=0.3.7
cross build --target mips-unknown-linux-musl
```

#### What is the actual behavior?

```
$ cross build --target mips-unknown-linux-musl
   Compiling memchr v2.6.4
   Compiling regex-syntax v0.7.5
   Compiling aho-corasick v1.1.2
   Compiling regex-automata v0.3.7
error: Undefined temporary symbol $BB505_1

error: Undefined temporary symbol $BB506_1

error: Undefined temporary symbol $BB507_1

error: could not compile `regex-automata` (lib) due to 3 previous errors
```

(Note that the same error occurs when using `cargo build --target mips-unknown-linux-musl`, provided you have the target installed, so I don't think its `cross` specific.)

#### What is the expected behavior?
Switch back to the previous version of regex-automata:
```
cargo add regex-automata@=0.3.6
```
And everything builds fine:
```
$ cross build --target mips-unknown-linux-musl
   Compiling regex-automata v0.3.6
   Compiling regex-mips-issue v0.1.0 (/project)
    Finished dev [unoptimized + debuginfo] target(s) in 6.46s
```

PR #1127, opened 2023-11-14
chore: attempt to re-enable mipsel in CI

jonathanhood:
This set of changes attempts to address feedback in #1113 that mipsel targets couldn't be supported since they aren't working in CI. I'm also encountering the issue called out there. While I can't claim to know how to fix the original issue - I know a thing about cross compiling so here we go.

Raising this as a draft while I struggle with the CI runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           target: powerpc64-unknown-linux-gnu
         - build: stable-mipsel
           os: ubuntu-latest
-          rust: stable
+          rust: 1.72.0
           target: mipsel-unknown-linux-gnu
         - build: stable-s390x
           os: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           target: powerpc64-unknown-linux-gnu
         - build: stable-mipsel
           os: ubuntu-latest
-          rust: 1.72.0
+          rust: 1.71.0
           target: mipsel-unknown-linux-gnu
         - build: stable-s390x
           os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jonathanhood:
It appears this is a consequence of mips support being dropped in status: https://github.com/rust-lang/compiler-team/issues/648

As a result, the latest rust version which passes CI using cross/xargo/etc is `0.71.0`. Since the relevant support was dropped in [`0.72.0`](https://releases.rs/docs/1.72.0/), that makes sense. 

I guess the best fix at this level, then, to keep this target in CI is pin the rust version there to the latest supported. I suspect a lot of people stuck in mips land (including myself) will be forced down this road in various ways.

I don't know if this solution is amenable to the maintainers here. It's a frustrating situation, but I also understand the need to focus the community. I'll leave this PR up, changing it from draft to "ready", as a way to get feedback on what the next steps here are.

BurntSushi:
Yes, I knew that mips support was dropped in status by Rust itself. :-)

I'm not open to pinning the Rust version like that. It might work temporarily, but it's not a long term solution because eventually this crate will require a newer Rust version and thus the pin will just break. The only way I can stomach mips in CI is if it reliably builds with the latest stable Rust release. Otherwise, folks relying on mips will just need to submit patches on an as-needed basis.

If it helps, there generally shouldn't need to be any mips specific code in this crate. So I generally don't expect to need any mips specific patches. I would generally expect most of the work required here to be supporting mips in Rust itself.

jonathanhood:
Unfortunately it appears the core issue in rust actually has to do with relevant support from llvm - which I'm _not_ equipped to solve unfortunately. Thanks for at least looking into this. I'll close this PR while I consider what the next possible steps are.

PR #1129, opened 2023-11-22
Clippy fixes

chris-ha458:
Currently, the default `cargo clippy` shows a lot of potential fixes.
I've prepared a potential initial batch that could be applied.
There are more, and if this PR merges I shall prepare them further.

Most are one liner changes employing clippy suggestions with adjustments as necessary.

### DIFF
--- a/regex-automata/src/nfa/thompson/range_trie.rs
+++ b/regex-automata/src/nfa/thompson/range_trie.rs
@@ -235,7 +235,7 @@ impl RangeTrie {
     /// Clear this range trie such that it is empty. Clearing a range trie
     /// and reusing it can beneficial because this may reuse allocations.
     pub fn clear(&mut self) {
-        self.free.extend(self.states.drain(..));
+        self.free.append(&mut self.states);
         self.add_empty(); // final
         self.add_empty(); // root
     }
@@ -296,7 +296,7 @@ impl RangeTrie {
         assert!(!ranges.is_empty());
         assert!(ranges.len() <= 4);
 
-        let mut stack = mem::replace(&mut self.insert_stack, vec![]);
+        let mut stack = core::mem::replace(&mut self.insert_stack, vec![]);
         stack.clear();
 
         stack.push(NextInsert::new(ROOT, ranges));
@@ -866,7 +866,7 @@ impl Split {
 
 impl fmt::Debug for RangeTrie {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        writeln!(f, "")?;
+        writeln!(f)?;
         for (i, state) in self.states.iter().enumerate() {
             let status = if i == FINAL.as_usize() { '*' } else { ' ' };
             writeln!(f, "{}{:06}: {:?}", status, i, state)?;
--- a/regex-automata/src/util/determinize/state.rs
+++ b/regex-automata/src/util/determinize/state.rs
@@ -115,7 +115,7 @@ pub(crate) struct State(Arc<[u8]>);
 /// without having to convert it into a State first.
 impl core::borrow::Borrow<[u8]> for State {
     fn borrow(&self) -> &[u8] {
-        &*self.0
+        &self.0
     }
 }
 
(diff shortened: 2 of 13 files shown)


regex-automata/src/nfa/thompson/range_trie.rs
```diff
@@ -296,7 +296,7 @@ impl RangeTrie {
         assert!(!ranges.is_empty());
         assert!(ranges.len() <= 4);
 
-        let mut stack = mem::replace(&mut self.insert_stack, vec![]);
+        let mut stack = std::mem::take(&mut self.insert_stack);
```
BurntSushi, line 299: This needs to be `core::mem::take` because `regex-automata` has a conditional dependency on `std` (and even `alloc`).

regex-automata/src/util/determinize/state.rs
```diff
@@ -460,12 +460,10 @@ impl<'a> Repr<'a> {
...
-        if !self.is_match() {
-            return 0;
-        } else if !self.has_pattern_ids() {
-            1
-        } else {
-            self.encoded_pattern_len()
+        match (self.is_match(), self.has_pattern_ids()) {
+            (false, _) => 0,
+            (true, false) => 1,
+            (true, true) => self.encoded_pattern_len(),
```
BurntSushi, line 466: I find the pre-existing code easier to read. Although it looks like you could at least change `return 0;` to `0`.

regex-automata/src/util/escape.rs
```diff
@@ -31,7 +31,7 @@ impl core::fmt::Debug for DebugByte {
         let mut len = 0;
         for (i, mut b) in core::ascii::escape_default(self.0).enumerate() {
             // capitalize \xab to \xAB
-            if i >= 2 && b'a' <= b && b <= b'f' {
+            if i >= 2 && (b'a'..=b'f').contains(&b) {
```
BurntSushi, line 34: I like the pre-existing code better.

BurntSushi (changes requested):
Nice, thanks!

I don't usually run Clippy because there's a lot I disagree with, and just don't have the time to sit down and configure the lint list to reflect my taste. (And also don't like littering code with `allow(..)` annotations on a case-by-case basis.)

With that said, I actually agree with most of these. There are a few that I would like to see reverted, but most look good. Part of the reason there are a lot is because `regex-automata` went through several rather large revisions while I worked on it, and not all code got thoroughly refactored. (That explains things like superfluous `ref` bindings and unnecessary `clone` calls.) And then of course, `regex-automata`'s MSRV was much lower than what it is now compared to when I started.

regex-automata/src/nfa/thompson/range_trie.rs
chris-ha458: reverted

BurntSushi:
The new support in Cargo is great, but it doesn't really change my calculus here. And indeed, at least for a lot of my crates, we'd need to be mindful of MSRV. (It's possible it won't matter, for example, if Cargo will just ignore that section otherwise.)

I'll try to be a bit clearer here... I personally do not want to have several dozen PRs scattered throughout all of my projects for setting up lint configurations right now. In part because I don't actually know whether I'll go through with it, so I don't want you to spend all that effort on it. Namely, the key thing that can't be addressed by any amount of work or configuration is that Clippy will invariably lead to adding `allow(..)` annotations in the source code. I can abide a couple here and there, but it's just in general not something I really want to put up with.

If my projects had a lot more contributors or multiple active maintainers then I would very likely feel differently about this.

chris-ha458:
I understand what you are saying.

I have reverted the issues you mentioned in this PR, are there any further changes you'd like to see?

regex-automata/src/util/search.rs
```diff
@@ -1365,16 +1365,17 @@ impl<'a> Iterator for PatternSetIter<'a> {
...
-                // permit storing invalid pattern IDs. Thus, 'yes' is only true
-                // precisely when 'index' corresponds to a valid 'PatternID'.
-                return Some(PatternID::new_unchecked(index));
+                Some(PatternID::new_unchecked(index))
+            } else {
+                None
             }
-        }
-        None
+        })
```
BurntSushi, line 1378: I find the previous code easier to read.

PR #1133, opened 2023-11-30
docs: fix inconsistent usage of a/an in `rure.h`

ThomasFrans:
A small fix for a typo in a documentation comment in the `rure.h` header file.


BurntSushi:
I don't think this was a typo.

PR #1140, opened 2023-12-20
Fixing documentation for Index<&str>

kloune:
The documentation for Index<&'n str> currently references the get method instead of the name method.


BurntSushi (changes requested):
Thanks! Could you please also fix the [same bug in the `bytes` submodule](https://github.com/rust-lang/regex/blob/4f5992fa442bb469be7042454e6a1b74181dd9a5/src/regex/bytes.rs#L1992)?

kloune:
Done for the bytes.rs submodule as well.

BurntSushi (approved):
Thanks!

PR #1150, opened 2024-01-12
cargo: Ensure 'perf' doesn't enable 'std' implicitly

Palkovsky:
Enabling `perf-literal` in the `regex` implicitly enables `std` (through `perf-literal-multisubstring` in the `regex-automata`). Effectively, it is impossible to use enable `perf` for no-std targets.

Quoting @BurntSushi:
> I think one could make the argument that features like perf-literal-multisubstring are generally orthogonal to whether a dependency on std should be introduced. Indeed, that feature does enable optimizations other than SIMD. So it seems perfectly cromulent to me that one might want to enable it without dependency on std.

> So I think here is what I propose:
>  - Remove the std feature from perf-literal-multisubstring.
>  - Add a new section to the crate features docs prominently calling out that some of the SIMD optimizations require a dependency on std. And that most features, including perf-* features, will not also implicitly add a dependency on std. Instead, it is expected that users configuring the features for the crate will do that.

Fixes https://github.com/rust-lang/regex/discussions/1147

### DIFF
--- a/regex-automata/Cargo.toml
+++ b/regex-automata/Cargo.toml
@@ -44,7 +44,7 @@ perf = ["perf-inline", "perf-literal"]
 perf-inline = []
 perf-literal = ["perf-literal-substring", "perf-literal-multisubstring"]
 perf-literal-substring = ["aho-corasick?/perf-literal", "dep:memchr"]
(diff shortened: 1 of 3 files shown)


BurntSushi (approved):
It's possible this breaks people, but I consider this a bug fix.

PR #1154, opened 2024-01-19
Make `Input::new` guard against incorrect `AsRef` implementations

SkiFire13:
Currently `Input::new` calls `haystack.as_ref()` twice, once to get the actual `haystack` slice and the second time to get its length. It makes the assumption that the second call will return the same slice, but malicious implementations of `AsRef` can return different slices and thus different lengths. This is important because there's `unsafe` code relying on the `Input`'s span being inbounds with respect to the `haystack`, but if the second call to `.as_ref()` returns a bigger slice this won't be true.

For example, this snippet causes MIRI to report UB on an unchecked slice access in `find_fwd_imp` (though it will also panic sometime later when run normally, but at that point the UB already happened):

```rs
use regex_automata::{Input, meta::{Builder, Config}};
use std::cell::Cell;

struct Bad(Cell<bool>);

impl AsRef<[u8]> for Bad {
    fn as_ref(&self) -> &[u8] {
        if self.0.replace(false) {
            &[]
        } else {
            &[0; 1000]
        }
    }
}

let bad = Bad(Cell::new(true));
let input = Input::new(&bad);
let regex = Builder::new()
    .configure(Config::new().auto_prefilter(false)) // Not setting this causes some checked access to occur before the unchecked ones, avoiding the UB
    .build("a+")
    .unwrap();
regex.find(input);
```

The proposed fix is to just call `.as_ref()` once and use the length of the returned slice as the span's end value. A regression test has also been added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,14 @@
 1.10.3 (TBD)
 ============
 This is a new patch release that fixes the feature configuration of optional
-dependencies.
+dependencies, and fixes an unsound use of bounds check elision.
 
 Bug fixes:
 
 * [BUG #1147](https://github.com/rust-lang/regex/issues/1147):
 Set `default-features=false` for the `memchr` and `aho-corasick` dependencies.
+* [BUG #1154](https://github.com/rust-lang/regex/pull/1154):
+Fix unsound bounds check elision.
 
 
 1.10.2 (2023-10-16)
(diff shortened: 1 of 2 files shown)


BurntSushi (changes requested):
Nice fix! Out of curiosity, how did you find this?

I think this overall looks good, but I'd like to find a word other than "malicious." The issue here really isn't "malicious" per se, because the threat model here doesn't really involve some bad actor doing something sneaky (if a bad actor can insert a malicious `AsRef` impl, then they can do a whole bunch of other stuff without need for such things). Perhaps "guarding against incorrect `AsRef` impls" is a better way to phrase it.

SkiFire13:
> Nice fix! Out of curiosity, how did you find this?

I happened to gave a quick look at `Input::new`'s source code and the two calls to `.as_ref()` reminded me of rust-lang/rust#80335 so I quickly checked if there was `unsafe` code relying on the span's end and there was.

BurntSushi (approved):

BurntSushi:
This PR is on crates.io in `regex 1.10.3`.

PR #1156, opened 2024-01-25
make additional prefileter metadata public

pascalkuthe:
These are the changes that I mentioned a while ago in #425

I now have a fully working streaming regex implementation at https://github.com/pascalkuthe/regex-cursor. These two APIs are the last two blockers before I can make helix switch to that implementation (we don't allow git dependencies/patches).

* makes `Prefilter::is_fast` public. Necessary to implement a meta engine downstream (also useful for any other downstream user that wants to write a custom meta engine)
* add a function that returns the maximum length of a prefilter. This is necessary to determine how much buffering is needed for streaming prefilters.

### DIFF
--- a/regex-automata/src/util/prefilter/mod.rs
+++ b/regex-automata/src/util/prefilter/mod.rs
@@ -146,6 +146,8 @@ pub struct Prefilter {
     pre: Arc<dyn PrefilterI>,
     #[cfg(feature = "alloc")]
     is_fast: bool,
+    #[cfg(feature = "alloc")]
+    max_needle_len: usize,
 }
 
 impl Prefilter {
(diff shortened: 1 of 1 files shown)


regex-automata/src/util/prefilter/mod.rs
```diff
@@ -411,6 +420,19 @@ impl Prefilter {
         }
     }
 
+    /// Return the length of the longest needle
+    /// in this Prefilter
+    pub fn max_needle_len(&self) -> usize {
```
BurntSushi, line 425: This should probably be marked `#[inline]`.

BurntSushi (changes requested):
One little nit and I think this is good to go! Thank you!

pascalkuthe:
done, thanks for the quick feedback!

BurntSushi (approved):

BurntSushi:
This PR is on crates.io in `regex-automata 0.4.5`.

PR #1172, opened 2024-03-10
Implement Display trait for string::Match

volsa:
I regularly have to convert matches of a named capture group into a string, which with the current master branch is achieved by `capture.as_str().to_string()`. This PR implements the `Display` trait to simplify the given snippet into `capture.to_string()`.

### DIFF
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -1549,6 +1549,12 @@ impl<'h> Match<'h> {
     }
 }
 
+impl<'h> std::fmt::Display for Match<'h> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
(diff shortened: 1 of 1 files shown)


BurntSushi:
`Display` is for user facing output, and there are many ways in which a `Match` could be formatted for users. Because of that, this trait impl isn't appropriate.

volsa:
Thanks for the feedback! I'm just curious and would like to understand why a user would want to see anything other than the `haystack` field? Put differently there are only two other fields (`start`, `end`) which I reckon are only interesting for debug cases and already covered by the `Debug` trait hence wouldn't the `Display` trait be valid for just displaying `haystack`?

BurntSushi:
Offsets might be important to users. Grep for example has options to print offsets. Offsets aren't just for debugging.

PR #1175, opened 2024-03-13
PR: add codes key/legend for debug output in regex-automata

Crypto-Spartan:
While I was looking at debug outputs of DFAs I found myself wondering what some of the symbols meant. `D` & `Q` were fairly intuitive, but it took me some additional digging to figure out what some of the other symbols meant. I took some inspiration from networking command output and added it to the debug output for the 3 DFA types (onepass, dense, & sparse) as well as the Thompson NFA.

Output of a network command: 
image

My proposed change:
image

You can see the `Codes` key/legend added at the top of the debug output.

It's a rather small PR, but I felt that the extra detail may help someone looking at the debug output in the future.


Crypto-Spartan:
also fixed some of the `write!` & `writeln!` clippy lints

PR #1180, opened 2024-04-05
Expose Regex memory usage info (fixes #943)

### LINKED ISSUE #943: API to get size of compiled regex
As an extension of the `RegexBuilder::size_limit` function,  it might be useful to know the actual size taken by a compiled regular expression. A possible use for this is for when you want to process a number of untrusted regular expressions, and be assured that collectively they don't exceed a limit.

I propose that the size of the compiled program (the same one used when checking against the `size_limit` of the builder) be stored inside a regular expression struct, and be accessible through a new function `Regex::approximate_size`.

mispecto:


Ambyjkl:
@BurntSushi sorry for the ping but I wanted to bring your attention to this PR, it would be great if this could get merged

Ambyjkl:
@spektom can you also expose the function on `RegexSet`?

BurntSushi:
I have concerns about this that I explained in #943. Those need to be resolved first.

LuisDuarte1:
I have almost the same usecase as @Ambyjkl - I basically want a "LRU cache" of regex rules as I'm in a semi-constrained memory environment. In my opinion, calling it `memory_usage_estimate` is more than good enough providing we document it properly.

typetetris:
Having a function like `memory_usage_estimate` for the exact same reason of having a size based cache would be very helpful to me, too.

At the moment I use regex-automata directly for that purpose and need to copy some stuff from the `regex` crate.

ashdnazg:
I also have the exact same caching motivation.

@BurntSushi Any way to help push this beyond suggesting names? I think we'd all be happy with any of `memory|heap|allocated` with any of `usage|bytes` etc. (avoid using `size` to make it clear it is not about the size limit), add `approximate|estimated` before or `estimate` after, and that should be fine.

IMO best is `approximate_allocated_bytes` since even though it's quite the mouthful, it sounds like it has nothing to do with the size limit.

BurntSushi:
I think `memory_usage` is probably good enough of a name. But the documentation should specifically call out that this isn't related to `RegexBuilder::size_limit`. That is, one cannot count on `memory_usage() < RegexBuilder::size_limit()`.

ashdnazg:
I opened a new PR https://github.com/rust-lang/regex/pull/1377 since I can't update this one.

PR #1185, opened 2024-04-22
syntax/utf8: avoid a spurious vector reallocation

lucab:
This reworks `Utf8Sequences` logic in order to avoid allocating a 0-sized vector and immediately reallocating it for the initial element. Directly create the populated vector instead.

### DIFF
--- a/regex-syntax/src/utf8.rs
+++ b/regex-syntax/src/utf8.rs
@@ -302,9 +302,9 @@ impl Utf8Sequences {
     /// Create a new iterator over UTF-8 byte ranges for the scalar value range
     /// given.
     pub fn new(start: char, end: char) -> Self {
-        let mut it = Utf8Sequences { range_stack: vec![] };
-        it.push(u32::from(start), u32::from(end));
-        it
+        let range =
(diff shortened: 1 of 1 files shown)


BurntSushi:
Can you say more about what motivated this change? Is this just something you happened to notice while reviewing the code, or did you notice something in a benchmark?

lucab:
Hi @BurntSushi and thanks for the prompt reply.
I was looking at the memory usage patterns of [rolldown](https://github.com/rolldown/rolldown) through [heaptrack](https://github.com/KDE/heaptrack), and this spot showed up as a potentially-spurious temporary allocation. The consumer side is [here](https://github.com/rolldown/rolldown/blob/ce36a195ed4e9ce7c446557cefff4750a2268e01/crates/rolldown/src/utils/extract_hash_pattern.rs#L12).
I do not have a specific benchmark for this.

BurntSushi (approved):
Sounds good. I was mostly just curious. While I agree there is a potentially spurious alloc here (assuming the optimizer doesn't elide it), I doubt it matters relative to all the other work being done. But still, this seems like an improvement. Thank you!

PR #1189, opened 2024-05-04
automata: remove some unsafe code

joshlf:
Note that the changes in `wire.rs` require zerocopy's "derive" feature, but the changes in `accel.rs` do not. If you'd prefer not to take a dependency on a proc macro derive, we can still salvage some of this PR.

### DIFF
--- a/regex-automata/Cargo.toml
+++ b/regex-automata/Cargo.toml
@@ -87,6 +87,7 @@ aho-corasick = { version = "1.0.0", optional = true, default-features = false }
 log = { version = "0.4.14", optional = true }
 memchr = { version = "2.6.0", optional = true, default-features = false }
 regex-syntax = { path = "../regex-syntax", version = "0.8.2", optional = true, default-features = false }
+zerocopy = { version = "0.7.32", default-features = false, features = ["derive"] }
 
 [dev-dependencies]
 anyhow = "1.0.69"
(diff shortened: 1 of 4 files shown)


BurntSushi:
I was aware of zerocopy (and similar crates) when I wrote this code, and knew that it could probably eliminate some `unsafe` blocks. But as a rule of thumb, I don't add dependencies to `regex`. I have spent a lot of time and effort to keep its dependency tree small and reasonable (the entire tree is owned by me for example). So not only does taking a proc macro dependency not meet my bar, zerocopy itself doesn't either. Or at least, it would need to provide a much bigger benefit.

The way to get rid of these unsafe blocks is to push on the "safe transmute" effort I think.

joshlf:
Yeah no worries, I understand the desire to avoid dependencies. We do the same. Happy to just close this if you'd prefer.

As for safe transmute, IIUC (@jswrenn is more directly involved and can provide more context), the idea is that safe transmute will basically just enable a trait bound, and you'll still need library code to provide useful abstractions given that bound. Roughly, safe transmute will replace zerocopy-derive, but not zerocopy (and presumably the same for bytemuck/bytemuck-derive, etc), and code like this PR would still require zerocopy in order to avoid writing `unsafe`.

BurntSushi:
Aye. Understood. Thanks for the attempt! I do overall support trying to remove `unsafe` blocks, but probably not with the addition of other dependencies.

PR #1195, opened 2024-05-17
automata: Fix broken universal start states with sparse DFA

ishitatsuyuki:
The state IDs were not remapped, which will usually result in an index out of range error.

### DIFF
--- a/regex-automata/src/dfa/sparse.rs
+++ b/regex-automata/src/dfa/sparse.rs
@@ -1846,6 +1846,14 @@ impl StartTable<Vec<u8>> {
             let new_start_id = remap[dfa.to_index(old_start_id)];
             sl.set_start(anchored, sty, new_start_id);
         }
+        for ustart in [
+            &mut sl.universal_start_unanchored,
+            &mut sl.universal_start_anchored,
+        ] {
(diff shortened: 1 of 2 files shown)


BurntSushi:
Is it possible to add a regression test?

ishitatsuyuki:
Sure, let me give it a try tomorrow (it's late here today).

ishitatsuyuki:
Hi, I added a test based on examples in `Automaton::is_special_state`. I'm asserting that the initial state is `.is_start_state()`, but this seems to be broken for some reason. Could you help figuring out why? I'm afraid I'm not familiar enough with how the start state ranges are being assigned.

ishitatsuyuki:
Alternatively, I could remove the assert for now and file an issue for it to be fixed later.

ishitatsuyuki:
It looks like start states were simply being removed with `set_no_special_start_states`, so this was the intended behavior. I have removed the assert now. Let me know if the tests look good to you.

ishitatsuyuki:
Hi, mind giving this a review?

ishitatsuyuki:
Friendly ping, let me know if there's something else I need to address

PR #1198, opened 2024-06-02
[docs] Fix words duplication in docs `extract`

denzenin:
Fix words duplication in docs `extract` in three files.

### DIFF
--- a/regex-lite/src/string.rs
+++ b/regex-lite/src/string.rs
@@ -1717,7 +1717,7 @@ impl<'h> Captures<'h> {
     ///
     /// This returns a tuple where the first element corresponds to the full
     /// substring of the haystack that matched the regex. The second element is
-    /// an array of substrings, with each corresponding to the to the substring
+    /// an array of substrings, with each corresponding to the substring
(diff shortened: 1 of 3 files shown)


BurntSushi (changes requested):
Please reflow the text.

BurntSushi (approved):

denzenin:
@BurntSushi Andrew, hi. Thanks for checking my PR so quickly. I am just starting with rust, can you please advise how do  you automatically reflow text in the comments? It looks like `rustfmt` does not touch comments by default

BurntSushi:
@denzenin Good question. I don't have a way of doing it automatically unfortunately. Instead, I do it manually via a combination of the [`par`](http://www.nicemice.net/par/) CLI utility and [neovim keybindings](https://github.com/BurntSushi/dotfiles/blob/f52cb7b0213a98690bddc73b3c60654a99942865/.config/nvim/include/keybinds.vim#L11-L20).

denzenin:
@BurntSushi thanks for the tips :)

PR #1202, opened 2024-06-06
cargo: point `repository` metadata to clonable URLs

lucab:
This tweaks the `repository` fields in Cargo metadata in order to use the correct (i.e. git clonable) URL. The existing GitHub webUI URLs for each package have been retained and moved to `homepage` fields.


lucab:
For reference, without this PR the remote fetch would fail with:
```
$ git clone https://github.com/rust-lang/regex/tree/master/regex-automata

Cloning into 'regex-automata'...
fatal: repository 'https://github.com/rust-lang/regex/tree/master/regex-automata/' not found
```

PR #1203, opened 2024-06-08
fix: escape invalid UTF-8 bytes in debug output for Match

notJoon:
# Description

1. The `Debug` implementation for `Match` has been updated to use `DebugHaystack`. This provides a way to handle the formatting of `&[u8]` for debug output.

 - Valid UTF-8 characters are output as is.
 - Invalid UTF-8 bytes are output as hex escape sequences (`\xHH`).
 - ASCII escape characters (e.g., `\t`, `\n`) are properly escaped.

2. Additional test cases have been added

### DIFF
--- a/src/regex/bytes.rs
+++ b/src/regex/bytes.rs
@@ -1,4 +1,4 @@
-use alloc::{borrow::Cow, string::String, sync::Arc, vec::Vec};
+use alloc::{borrow::Cow, format, string::String, sync::Arc, vec::Vec};
 
 use regex_automata::{meta, util::captures, Input, PatternID};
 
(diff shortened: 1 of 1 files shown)


src/regex/bytes.rs
```diff
@@ -1557,18 +1557,25 @@ impl<'h> core::fmt::Debug for Match<'h> {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
         let mut fmt = f.debug_struct("Match");
         fmt.field("start", &self.start).field("end", &self.end);
-        if let Ok(s) = core::str::from_utf8(self.as_bytes()) {
-            fmt.field("bytes", &s);
+
+        let bytes = self.as_bytes();
+        let formatted = bytes_to_string_with_invalid_utf8_escaped(bytes);
```
BurntSushi, line 1562: Can you use [`regex_automata::util::escape::DebugHaystack`](https://docs.rs/regex-automata/latest/regex_automata/util/escape/struct.DebugHaystack.html) instead? It will basically do what you have here, but will only escape invalid UTF-8. What you've implemented here will escape not only invalid UTF-8, but all UTF-8 that isn't ASCII. (I think that would be a cure worse than the disease.)

src/regex/bytes.rs
```diff
@@ -2620,3 +2627,54 @@ fn no_expansion<T: AsRef<[u8]>>(replacement: &T) -> Option<Cow<'_, [u8]>> {
...
+    fn test_debug_output_invalid_utf8() {
+        let haystack = b"Hello, \xFFworld!";
+        let m = Match::new(haystack, 7, 13);
+        let debug_str = format!("{:?}", m);
+
+        assert_eq!(
+            debug_str,
+            r#"Match { start: 7, end: 13, bytes: "\\xFFworld" }"#
+        );
+    }
```
BurntSushi, line 2679: Please add some tests with non-ASCII UTF-8.

BurntSushi (changes requested):

src/regex/bytes.rs
```diff
@@ -1557,18 +1557,25 @@ impl<'h> core::fmt::Debug for Match<'h> {
...
+        fmt.field("bytes", &formatted);
+
+        fmt.finish()
+    }
+}
+
+fn bytes_to_string_with_invalid_utf8_escaped(bytes: &[u8]) -> String {
+    let mut result = String::new();
+    for &byte in bytes {
+        if byte.is_ascii() {
```
BurntSushi, line 1572: > outputs valid UTF-8 characters as is

This is why what you said isn't accurate here. This only outputs ASCII characters as-is. Everything else, including valid UTF-8 that isn't ASCII, is emitted as escape byte sequences.

src/regex/bytes.rs
notJoon: Added along with other tests.
https://github.com/rust-lang/regex/pull/1203/commits/d18841ea60c35285aa8bc6e01aa6c98fd112d1d1

src/regex/bytes.rs
notJoon: Modified to use DebugHaystack. I thought there would be such a feature but couldn't find it. Thanks for your suggestion. https://github.com/rust-lang/regex/pull/1203/commits/88112b31c21fcd5cd6d57fdb6a056f0a96880166

BurntSushi (approved):
Thanks!

BurntSushi:
This PR is on crates.io in `regex 1.10.5`.

Issue #1216, opened 2024-07-27
Compilation error when compiling with nightly with `unstable` feature enabled


jqnatividad:
#### What version of regex are you using?

1.10.5

#### Describe the bug at a high level.

Compilation error when compiling with rust 1.82.0-nightly (7120fdac7 2024-07-25)

#### What are the steps to reproduce the behavior?

Compile a program that uses regex 1.10.5 with latest rust nightly with the `unstable` feature enabled.

#### What is the actual behavior?

```
[...]

PR #1219, opened 2024-08-02
Rust nightly removed the lifetime from Pattern

mischnic:
Closes https://github.com/rust-lang/regex/issues/1216

### DIFF
--- a/src/pattern.rs
+++ b/src/pattern.rs
@@ -13,7 +13,7 @@ pub struct RegexSearcher<'r, 't> {
 impl<'r> Pattern for &'r Regex {
     type Searcher<'t> = RegexSearcher<'r, 't>;
 
-    fn into_searcher<'h>(self, haystack: &'h str) -> RegexSearcher<'r, 'h> {
+    fn into_searcher<'t>(self, haystack: &'t str) -> RegexSearcher<'r, 't> {
         RegexSearcher {
             haystack,
             it: self.find_iter(haystack),
--- a/src/pattern.rs
+++ b/src/pattern.rs
@@ -10,10 +10,10 @@ pub struct RegexSearcher<'r, 't> {
     next_match: Option<(usize, usize)>,
 }
 
-impl<'r, 't> Pattern<'t> for &'r Regex {
-    type Searcher = RegexSearcher<'r, 't>;
+impl<'r> Pattern for &'r Regex {
+    type Searcher<'t> = RegexSearcher<'r, 't>;
 
-    fn into_searcher(self, haystack: &'t str) -> RegexSearcher<'r, 't> {
+    fn into_searcher(self, haystack: &str) -> RegexSearcher<'r, '_> {
         RegexSearcher {
             haystack,
             it: self.find_iter(haystack),
--- a/src/pattern.rs
+++ b/src/pattern.rs
@@ -13,7 +13,7 @@ pub struct RegexSearcher<'r, 't> {
 impl<'r> Pattern for &'r Regex {
     type Searcher<'t> = RegexSearcher<'r, 't>;
 
-    fn into_searcher(self, haystack: &str) -> RegexSearcher<'r, '_> {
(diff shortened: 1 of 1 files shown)


src/pattern.rs
```diff
@@ -10,10 +10,10 @@ pub struct RegexSearcher<'r, 't> {
     next_match: Option<(usize, usize)>,
 }
 
-impl<'r, 't> Pattern<'t> for &'r Regex {
-    type Searcher = RegexSearcher<'r, 't>;
+impl<'r> Pattern for &'r Regex {
+    type Searcher<'t> = RegexSearcher<'r, 't>;
 
-    fn into_searcher(self, haystack: &'t str) -> RegexSearcher<'r, 't> {
+    fn into_searcher(self, haystack: &str) -> RegexSearcher<'r, '_> {
```
BurntSushi, line 16: Should this be?

```
fn into_searcher(self, haystack: &'t str) -> RegexSearcher<'r, 't> {
```

mischnic: This lifetime setup is also used by the various `Pattern` impls in the standard library:
https://github.com/rust-lang/rust/commit/772315de7c1f68bb95de5a7eb85cb2f167834c53

Your version doesn't compile directly, the first suggestion below is exactly what I have implemented now (but with `'t` being elided) and the second suggestion is what was used before my PR and now fails to compile.
```
error[E0261]: use of undeclared lifetime name `'t`
  --> src/pattern.rs:16:68
   |
16 |     fn into_searcher(self, haystack: &'t str) -> RegexSearcher<'r, 't> {
   |                                                                    ^^ undeclared lifetime
   |
help: consider introducing lifetime `'t` here
   |
16 |     fn into_searcher<'t>(self, haystack: &'t str) -> RegexSearcher<'r, 't> {
   |                     ++++
help: consider introducing lifetime `'t` here
   |
13 | impl<'t, 'r> Pattern for &'r Regex {
   |      +++
```

BurntSushi: I guess the problem I have here is how does the `RegexSearcher` connect its lifetime to the haystack? I must be missing something.

mischnic: What I wrote here is equivalent to this, which is how the `&str` lifetime flows into the `RegexSearcher` lifetime (I can commit that instead if you want)
```
fn into_searcher<'h>(self, haystack: &'h str) -> RegexSearcher<'r, 'h>
```

BurntSushi: Ooooo, yes, I get it now. Yes, please do write that. I'm usually fine with lifetime elision, but I think it actually helps a lot here.

BurntSushi: Right right. This all makes sense now. Got it. I was confused myself.

mischnic: Perfect 😄 
I've added back that explicit lifetime

src/pattern.rs
```diff
@@ -10,10 +10,10 @@ pub struct RegexSearcher<'r, 't> {
     next_match: Option<(usize, usize)>,
 }
 
-impl<'r, 't> Pattern<'t> for &'r Regex {
-    type Searcher = RegexSearcher<'r, 't>;
+impl<'r> Pattern for &'r Regex {
+    type Searcher<'t> = RegexSearcher<'r, 't>;
 
-    fn into_searcher(self, haystack: &'t str) -> RegexSearcher<'r, 't> {
+    fn into_searcher<'h>(self, haystack: &'h str) -> RegexSearcher<'r, 'h> {
```
BurntSushi, line 16: Sorry, can this be `'t` instead of `'h` to be consistent? I don't think there is a shadowing concern here.

BurntSushi (approved):
Perfect, thanks!

BurntSushi:
This fix is on crates.io in `regex 1.10.6`.

mischnic:
Great, thank you for the fast release!

PR #1221, opened 2024-08-10
capi: add split

weebney:
Adds splitting to C API.


weebney:
cc @BurntSushi

weebney:
cc @BurntSushi

weebney:
cc @BurntSushi some feedback here would be wonderful

PR #1227, opened 2024-09-19
api: Add as_match method

tmccombs:
To simplify getting the whole match from Captures

### DIFF
--- a/src/regex/bytes.rs
+++ b/src/regex/bytes.rs
@@ -1665,6 +1665,15 @@ impl<'h> Captures<'h> {
             .map(|sp| Match::new(self.haystack, sp.start, sp.end))
     }
 
+    /// Return the overall match for the capture.
+    ///
+    /// This returns the match for index `0`. That is it is equivalent to
+    /// `get(0).unwrap()`
+    #[inline]
+    pub fn as_match(&self) -> Match {
+        self.get(0).unwrap()
+    }
+
     /// Returns the `Match` associated with the capture group named `name`. If
     /// `name` isn't a valid capture group or it refers to a group that didn't
     /// match, then `None` is returned.
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -1675,6 +1675,15 @@ impl<'h> Captures<'h> {
             .map(|sp| Match::new(self.haystack, sp.start, sp.end))
     }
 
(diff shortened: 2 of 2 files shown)


src/regex/bytes.rs
```diff
@@ -1665,6 +1665,15 @@ impl<'h> Captures<'h> {
             .map(|sp| Match::new(self.haystack, sp.start, sp.end))
     }
 
+    /// Return the overall match for the capture.
+    ///
+    /// This returns the match for index `0`. That is it is equivalent to
+    /// `get(0).unwrap()`
```
BurntSushi, line 1671: Can you use `m.get(0).unwrap()`?

src/regex/bytes.rs
```diff
@@ -1665,6 +1665,15 @@ impl<'h> Captures<'h> {
             .map(|sp| Match::new(self.haystack, sp.start, sp.end))
     }
 
+    /// Return the overall match for the capture.
```
BurntSushi, line 1668: Please add an example. I try to include an example for every public API item.

BurntSushi (changes requested):

src/regex/bytes.rs
```diff
@@ -1665,6 +1665,24 @@ impl<'h> Captures<'h> {
...
+    /// `m.get(0).unwrap()`
+    ///
+    /// ```
+    /// use regex::bytes::Regex;
+    ///
+    /// let re = Regex::new(r"[a-z]+([0-9]+)").unwrap();
+    /// let caps = re.captures(b"   abc123-def").unwrap();
+    ///
+    /// assert_eq!(caps.as_match().as_bytes(), b"abc123");
+    /// ```
```
BurntSushi, line 1680: This example is not formatted in the same way as all other examples.

Please look at other public API items and format this one in the same way. It should have an `# Example` heading, for example.

BurntSushi (changes requested):

BurntSushi:
Also, I think the name `as_match` is not ideal. I used [`get_match`](https://docs.rs/regex-automata/latest/regex_automata/util/captures/struct.Captures.html#method.get_match) in the lower level API. `get_match` isn't ideal either, but I think it's probably better than `as_match`.

PR #1228, opened 2024-09-19
api: add SetMatches::matched_all

tmccombs:
Which compliments `matched_any` with a means to check if a set of patterns ALL matched the haystack.

### DIFF
--- a/src/regexset/bytes.rs
+++ b/src/regexset/bytes.rs
@@ -482,18 +482,18 @@ impl SetMatches {
         !self.0.is_empty()
     }
 
-    /// Whether all patterns in this set matched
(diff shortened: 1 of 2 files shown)


src/regexset/bytes.rs
```diff
@@ -482,6 +482,24 @@ impl SetMatches {
         !self.0.is_empty()
     }
 
+    /// Whether all patterns in this set matched
```
BurntSushi, line 485: Looks like you missed a period here (and below).

BurntSushi (changes requested):

BurntSushi (approved):

PR #1235, opened 2024-11-03
Suggest `std::sync::LazyLock` instead of `once_cell::sync::Lazy`

### LINKED ISSUE #1217: README: we recommend using the once_cell crate to ensure that regular expressions are compiled
LazyLock is stabilized in 1.80, maybe it's recommend than once_cell crate

noritada:
This PR suggests `std::sync::LazyLock` instead of `once_cell::sync::Lazy`.

This issue was discussed three months ago right after the stabilization of `std::sync::LazyLock` in Rust 1.80,
and since two more minor versions of Rust have been released since then,
I think it's about time.

Many thanks,

Closes #1217

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,18 +80,24 @@ compilation itself expensive, but this also prevents optimizations that reuse
 allocations internally to the matching engines.
 
 In Rust, it can sometimes be a pain to pass regular expressions around if
-they're used from inside a helper function. Instead, we recommend using the
-[`once_cell`](https://crates.io/crates/once_cell) crate to ensure that
-regular expressions are compiled exactly once. For example:
+they're used from inside a helper function. Instead, we recommend using
+[`std::sync::LazyLock`] stabilized in [Rust 1.80], the [`once_cell`] crate, or
+the [`lazy_static`] crate to ensure that regular expressions are compiled
+exactly once. For example:
+
+[`std::sync::LazyLock`]: https://doc.rust-lang.org/std/sync/struct.LazyLock.html
+[Rust 1.80]: https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html
+[`once_cell`]: https://crates.io/crates/once_cell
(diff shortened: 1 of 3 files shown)


nyurik:
thx!! Do we really want to mention 3rd party crates when the exact same functionality is available in stdlib?

README.md
```diff
@@ -80,18 +80,24 @@ compilation itself expensive, but this also prevents optimizations that reuse
 allocations internally to the matching engines.
 
 In Rust, it can sometimes be a pain to pass regular expressions around if
-they're used from inside a helper function. Instead, we recommend using the
-[`once_cell`](https://crates.io/crates/once_cell) crate to ensure that
-regular expressions are compiled exactly once. For example:
+they're used from inside a helper function. Instead, we recommend using
+[`std::sync::LazyLock`] stabilized in [Rust 1.80], the [`once_cell`] crate, or
+the [`lazy_static`] crate to ensure that regular expressions are compiled
+exactly once. For example:
```
gibfahn, line 86: Is there a reason not to just use the `LazyLock` option? As a user without much idea on the history, it would be nice if there was just one standard and recommended option, the stdlib one.

Also the one that clippy recommends: https://rust-lang.github.io/rust-clippy/master/index.html#non_std_lazy_statics

gibfahn:
I found this PR (eventually) because clippy now has a pedantic lint against lazy_static: https://rust-lang.github.io/rust-clippy/master/index.html#non_std_lazy_statics

Would be nice to get this landed so it's easy for folks who see that lint to update.

README.md
noritada: @gibfahn
Thanks for your comment.
As `regex` and `once_cell` can be used with `no_std` where `std::sync::LazyLock` is not available, I thought it would be a good idea to leave mention of `once_cell`.
Since I have not tried it, if I am wrong please let me know.

shilangyu (changes requested):
ok

PR #1240, opened 2024-11-13
Increase pool stacks to 32

fbryden:


BurntSushi:
A change like this without literally any accompanying explanation or benchmarks isn't really appropriate. The obvious downside of this change is that it [increases memory usage across the board](https://github.com/rust-lang/regex/pull/1240/files#diff-e74c6359b5e11d923b2bacbf940a690332009b4aa6419133e07ab12f96e64a1fR496). In particular, that `CacheLine` is problematic since it takes up 64 bytes on its own. So this change increases base memory usage from `8 * 64 = 512 bytes` to `32 * 64 = 2,048 bytes`.

fbryden:
@BurntSushi I am so so sorry, I did not mean to open this pull request _against the original repo_ - it was only meant for my fork. I completely understand (and agree) this change is not appropriate across the board!

fbryden:
Thank you so much!
I've filed https://github.com/rust-lang/regex/issues/1241

PR #1254, opened 2025-02-17
Update README.md to use LazyLock instead of OnceCell

noamraph:
A small things that you might consider: currently the README recommends using the external crate `once_cell`. Since Rust 1.80 we can use LazyLock from the std to get the same behavior. Perhaps you'd want to update the readme.

Thanks for your great work!
Noam


noritada:
This appears to be a duplicate of #1235.

PR #1274, opened 2025-07-31
Search `&[u8]` haystacks in `regex_lite`

01mf02:
To quote the [`regex_lite` docs](https://docs.rs/regex-lite/latest/regex_lite/#differences-with-the-regex-crate):

> Currently, this crate only supports searching &str. It does not have APIs for searching &[u8] haystacks, although it is planned to add these in the future if there’s demand.

This PR adds such an API in form of `regex_lite::bytes::Regex`, mirroring `regex::bytes::Regex`.
Thanks, @BurntSushi, for leaving some breadcrumbs in the source code that made this easier, in particular your hint in `interpolate::bytes`. This is much appreciated, as well as all your effort that went not only in the careful and beautiful API design, but also your very detailed and user-friendly documentation. Thanks a lot for your awesome work.

I made this PR because I need this functionality as part of [jaq](https://github.com/01mf02/jaq), where I want to transition from `String` to (a variation of) `Vec<u8>`, inspired by your ["UTF-8 by convention" idea](https://burntsushi.net/bstr/).
I did not `cargo fmt` this PR in order to make reviewing as easy as possible.

(Note to self: If I ever have to convert a lot of `b"foo"` to `&b"foo"[..]` again, this is how to do it in vim for all matches on the current line: `:s/b"\([^"]*\)"/\&b"\1"[..]/g`.)

### DIFF
--- a/regex-lite/src/bytes.rs
+++ b/regex-lite/src/bytes.rs
@@ -419,9 +419,7 @@ impl Regex {
         CaptureMatches {
             haystack,
             re: self,
-            it: self
-                .pikevm
-                .captures_iter(self.pool.get(), haystack),
+            it: self.pikevm.captures_iter(self.pool.get(), haystack),
         }
     }
 
--- a/regex-lite/src/bytes.rs
+++ b/regex-lite/src/bytes.rs
@@ -1,0 +1,2980 @@
+//! Search for regex matches in `&[u8]` haystacks.
+//!
+//! TODO: Adapt documentation from `src/bytes.rs`?
+
+use alloc::{
+    borrow::Cow, boxed::Box, string::String, string::ToString, sync::Arc, vec,
+    vec::Vec,
+};
+
+use crate::{
+    error::Error,
+    hir::{self, Hir},
+    int::NonMaxUsize,
+    interpolate,
+    nfa::{self, NFA},
+    pikevm::{self, Cache, PikeVM},
+    pool::CachePool,
+};
+
+/// A compiled regular expression for searching Unicode haystacks.
+///
+/// A `Regex` can be used to search haystacks, split haystacks into substrings
+/// or replace substrings in a haystack with a different substring. All
+/// searching is done with an implicit `(?s:.)*?` at the beginning and end of
+/// an pattern. To force an expression to match the whole string (or a prefix
+/// or a suffix), you must use an anchor like `^` or `$` (or `\A` and `\z`).
+///
+/// While this crate will handle Unicode strings (whether in the regular
+/// expression or in the haystack), all positions returned are **byte
+/// offsets**. Every byte offset is guaranteed to be at a Unicode code point
+/// boundary. That is, all offsets returned by the `Regex` API are guaranteed
+/// to be ranges that can slice a `&str` without panicking.
+///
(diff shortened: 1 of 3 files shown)


01mf02:
So I think that I would like to keep `regex-lite` in `jaq` for the time being, given their nice match of design philosophy.

If it helps you with the reviewing of this PR: I literally just copied `string.rs` to `bytes.rs` and replaced a lot of `&str` by equivalent `&[u8]` parts. Using `diff string.rs bytes.rs` should be rather quick to breeze through.

BurntSushi:
> Is this what you mean by disabling Unicode mode? (I searched the regex-automata docs for "Unicode mode", and the syntax::Config::unicode function was the only one that came up.)

I liked you to [`syntax::Config::unicode`](https://docs.rs/regex-automata/0.4.11/regex_automata/util/syntax/struct.Config.html#method.unicode) above. :-)

> This does not affect the performance at all. Whatever values I pass to `utf8()` and `unicode()`, the performance remains stable up to the millisecond. (Remember, I'm not actually searching with the regex, just compiling it.)

Oh wow that is interesting. But yeah, compiling is where I'd expect it to help.

> By the way, something that I noticed when I read your [rebar](https://github.com/BurntSushi/rebar) README and several other posts by you, such as your [regex-lite introduction on Reddit](https://www.reddit.com/r/rust/comments/14ralde/announcing_regexlite_similar_to_the_regex_crate/): I found it often ambiguous when you wrote about "compile times", as in:

Yeah I agree I should be clearer. In rebar, it's only talking about regex compile times. In the `regex-lite` crate docs, it's talking about rustc compile times. But I should probably mention regex compile times in the `regex-lite` crate docs. It's just that usually people don't care about regex compile times because they put them in a `LazyLock` or can otherwise amortize their cost. Many more care about rustc compile times. And indeed, rustc compile times and binary size were the primary motivations that led me to build `regex-lite`. The fact that this results in fast regex compile times is mostly incidental but also a fairly direct consequence of its design philosophy. Less code generally means less optimization and less time spent doing that optimization.

> If it helps you with the reviewing of this PR: I literally just copied `string.rs` to `bytes.rs` and replaced a lot of `&str` by equivalent `&[u8]` parts. Using `diff string.rs bytes.rs` should be rather quick to breeze through.

Thanks! But there are some [subtle differences](https://docs.rs/regex/latest/regex/bytes/index.html#syntax) that I need to think/review carefully. And there is also how iteration is handled in the presence of invalid UTF-8. And making sure the tests are all wired up correctly.

I'll try to get to it soon, but I don't make any promises.

01mf02:
> Thanks! But there are some [subtle differences](https://docs.rs/regex/latest/regex/bytes/index.html#syntax) that I need to think/review carefully. And there is also how iteration is handled in the presence of invalid UTF-8. And making sure the tests are all wired up correctly.
> 
> I'll try to get to it soon, but I don't make any promises.

I understand. Take your time, and I think that I'll publish a crate that contains my current state, in the hope that I can at some point replace it with your upstream version again.

01mf02:
@BurntSushi: Whoops, I accidentally deleted the repository with my changes. I would still be interested to get this merged, though. But before I investigate how to restore the repository, may I ask whether you are still interested in this functionality?

PR #1277, opened 2025-08-14
automata: `Input::haystack` return with original lifetime

Chaoses-Ib:
`Input::haystack` currently returns the haystack with the lifetime of `&self`, not `'h`. This sometimes is not enough and can only be worked around by `transmute`. Should we return the original lifetime instead?


BurntSushi:
#1278 is needed to fix CI. Once that's merged, can you rebase this on master?

PR #1286, opened 2025-08-24
Inline format-args automata

nyurik:
This makes the code a bit easier to read and smaller.

Some of it was done with this command, and later fixed by hand:

```
 cargo clippy --workspace --allow-dirty --fix --benches --tests --bins -- -A clippy::all -W clippy::uninlined_format_args
 ```


nyurik:
Fixed. Do note that there are quiet a few cases of `rg '^.{80,}' -g '*.rs'`

PR #1289, opened 2025-08-25
Also exclude tests/fuzz from the package

erickt:
I missed more binary test data `tests/fuzz/` in #1281, which should remove another 56KiB from the regex crate.


BurntSushi (approved):
Nice, thanks!

PR #1308, opened 2025-10-20
perf: Avoid re-canonicalizing the entire IntervalSet on push

Marwes:
Canonicalize is taking up a significant amount due to a regex with a huge amount of character ranges (generated by [lalrpop](https://github.com/lalrpop/lalrpop)'s lexer expanding multiple `\w` in a token). While this could perhaps be fixed in lalrpop I did notice the TODO in the code and after addressing this so we automatically union and compress on each push instead of re-canonicalizing on every push and that fixed the performance problem.

I did see the earlier attempt at this https://github.com/rust-lang/regex/pull/1051 and it seems like that was reverted and regression tests were added so I hope that and the existing tests are enough (I don't have a clear idea on what tests might be missing).

### DIFF
--- a/regex-syntax/src/hir/interval.rs
+++ b/regex-syntax/src/hir/interval.rs
@@ -86,9 +86,10 @@ impl<I: Interval> IntervalSet<I> {
             return;
         };
 
-        // The search finds us the first index where the previous interval start is less
-        // than or equal to the new interval start. Since the existing intervals are non-overlapping
-        // we only need to try to union this single preceding interval
+        // The search finds us the first index where the previous interval
+        // start is less than or equal to the new interval start. Since the
+        // existing intervals are non-overlapping we only need to try to union
+        // this single preceding interval
         let mut start = i;
         if let Some(before_i) = i.checked_sub(1) {
             let before = &self.ranges[before_i];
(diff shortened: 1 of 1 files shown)


Marwes:
I have done some optimization around the code where this occurs which brings up `canonicalize` to about 5% in our benchmarks, any chance this could be merged or what is necessary to add to this to get it over the line?

regex-syntax/src/hir/interval.rs
```diff
@@ -80,11 +80,38 @@ impl<I: Interval> IntervalSet<I> {
...
+        for after_i in i..self.ranges.len() {
+            let after = &self.ranges[after_i];
+            if let Some(union) = interval.union(after) {
+                interval = union;
+                end = after_i + 1;
+            } else {
+                break;
+            }
+        }
+
```
BurntSushi:
```suggestion
```

regex-syntax/src/hir/interval.rs
```diff
@@ -80,11 +80,38 @@ impl<I: Interval> IntervalSet<I> {
...
+        let mut start = i;
+        if let Some(before_i) = i.checked_sub(1) {
+            let before = &self.ranges[before_i];
+            if let Some(union) = before.union(&interval) {
+                interval = union;
+                start = before_i;
+            }
+        }
+        // `interval` may overlap any number of intervals following the insertion point
+        // so will union each of them until we reach the first non-overlapping interval
```
BurntSushi: Can you wrap comments to 79 columns inclusive please?

regex-syntax/src/hir/interval.rs
```diff
@@ -80,11 +80,38 @@ impl<I: Interval> IntervalSet<I> {
...
+        let mut end = i;
+        for after_i in i..self.ranges.len() {
+            let after = &self.ranges[after_i];
+            if let Some(union) = interval.union(after) {
+                interval = union;
+                end = after_i + 1;
+            } else {
+                break;
+            }
```
BurntSushi:
```suggestion
            let after = &self.ranges[after_i];
            let Some(union) = interval.union(after) else { break };
            interval = union;
            end = after_i + 1;
```

BurntSushi (changes requested):
Hi! Yes, thank you and apologies for the late reply.

The change here itself looks good to me. Thank you for working through this subtle optimization. :-) I have only minor nits on this PR and a request.

Could you add a benchmark to rebar? I think probably [this directory](https://github.com/BurntSushi/rebar/blob/289cb1089a003a7953d6c52db636460e2f73605f/benchmarks/definitions/unicode/compile.toml) is the place to add it. And if you could show a before-and-after on the rebar results here, that would be great.

BurntSushi (approved):

BurntSushi:
> I did see the earlier attempt at this #1051 and it seems like that was reverted and regression tests were added so I hope that and the existing tests are enough (I don't have a clear idea on what tests might be missing).

Yeah the revert in #1102 did add regression tests.

PR #1266, opened 2025-05-15
Add support for unbounded look-behind expressions

Multimodcrafter:
As an example consider the regex `(?<=Title:\s+)\w+` which would match the
following strings (matches underlined with `~`):

```
Title: HelloWorld
       ~~~~~~~~~~
```

```
Title: Title: foo
       ~~~~~
              ~~~
```

But **does not** match:

- `No heading`
- `title: bad case`
- `Title:nospace`

## What

This PR implements the streaming algorithm from
[Linear Matching of JavaScript Regular Expressions (Section 4.4)](https://aurele-barriere.github.io/papers/linearjs.pdf#page=13)
for unbounded look-behinds. The same algorithm has been
[implemented and merged into V8](https://chromium-review.googlesource.com/c/v8/v8/+/5093860).
The addition of look-around expressions to this crate was mentioned previously
in [#1153](https://github.com/rust-lang/regex/discussions/1153).

This PR adds support for positive and negative look-behinds with arbitrary
nesting. With the following limitations

### Limitations

- Look-behind expressions cannot contain capture groups
- The algorithm is implemented only in the PikeVM and with prefiters off
- Only look-behinds and no look-aheads

Capture groups outside of look-arounds are supported. With the current capture
group semantics, no linear time algorithm which would allow for capture groups
inside of look-arounds is known. However, look-behinds could be implemented in
other engines and with prefilters on. Look-aheads could also be implemented with
additional memory.

## How

We implemented the streaming algorithm presented in Section 4.4 of the paper
mentioned above. The algorithm works by running the sub-automata for any
look-behind expressions in parallel to the main automaton. This is achieved by
compiling the look-behind expressions as usual but storing their start states
separately, not reachable from the main automaton.

Instead of a `match` state, the sub-automata for look-behinds have a
`WriteLookAround` state. This state causes the current position in the haystack
to be recorded in a global look-around table.

The main automaton (and the sub-automata in the case of nested look-behinds) can
then read from this table by means of a `CheckLookAround` instruction and
compare the stored index with the current position in the haystack. These states
work as conditional epsilon transitions, similar to the already supported "look"
assertions (e.g. `^`, `\b`, `$`).

`PikeVM`'s cache has been expanded to preserve good performance of single-match
searches (stop the look-around threads once the main automaton finishes) and of
all-matches searches (remember the look-around states when resuming a search to
prevent having to rescan the haystack from the beginning).

## Testing

We have added unit tests for the new functionality in the individual test
[...]

PR #1315, opened 2025-11-10
Look-behinds [1 / 6]: regex-syntax support

shilangyu:
Here is the first out of six PRs aiming to bring look-behinds to rust-regex. See #1266 for more information, notably [the implementation plan](https://github.com/rust-lang/regex/pull/1266#issuecomment-3304059556).

In this PR we focus on adding look-behind support to the regex-syntax crate. New additions are hidden behind a feature flag to not break compatibility with other crates.

Currently the CI fails on generation of docsrs. This is due to all features being enabled during generation which breaks regex-automata. This will be fixed in the second PR which works on regex-automata.

### DIFF
--- a/regex-syntax/Cargo.toml
+++ b/regex-syntax/Cargo.toml
@@ -35,6 +35,8 @@ unicode-perl = []
 unicode-script = []
 unicode-segment = []
 
+look-behinds = []
+
 [dependencies]
 arbitrary = { version = "1.3.0", features = ["derive"], optional = true }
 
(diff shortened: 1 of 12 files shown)


SamB:
> Currently the CI fails on generation of docsrs. This is due to all features being enabled during generation which breaks regex-automata. This will be fixed in the second PR which works on regex-automata.

@BurntSushi said in <https://github.com/rust-lang/regex/pull/1266#issuecomment-3228260553>:

> * A change that adds _opt-in_ support for parsing look-behind in `regex-syntax`. **(This may have "mundane" changes to `regex-automata` to reject HIR that have look-behind in them.)**

(Emphasis mine.)

It looks like we very much do want that. But how do we conditionalize this on the regex-syntax feature, though?

shilangyu:
> But how do we conditionalize this on the regex-syntax feature, though?

ATM it is behind a compile-time flag, so it does not break regex-automata.

PR #1334, opened 2026-02-18
Use core::error::Error instead of std::error::Error

ppentchev:
Hi,

First of all, thanks a lot for writing and maintaining the regex crate!

What do you think about this change that makes all the error types derive `core::error::Error` instead of `std::error::Error` so that they are usable even in `no-std` mode? There are several error handling crates like `anyhow`, `eyre`, etc. which also work in `no-std` mode, and they refuse to accept the error objects defined in these crates then.

Thanks in advance, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/regex-automata/src/dfa/automaton.rs
+++ b/regex-automata/src/dfa/automaton.rs
@@ -2114,8 +2114,7 @@ impl StartError {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for StartError {}
+impl core::error::Error for StartError {}
 
 impl core::fmt::Display for StartError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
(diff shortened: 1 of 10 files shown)


ppentchev:
...sorry, I just realized why it was this way - at the very least since `core::error` was only stabilized in 1.81. Oof. Many apologies, I don't know how I managed to overlook that!

BurntSushi:
I'm actually okay with doing the MSRV bump necessary for this impl. I was actually thinking of bumping to Rust 1.85 and doing a migration to Rust 2024.

BurntSushi:
But yes, MSRV is why this hasn't been done until now.

ppentchev:
> I'm actually okay with doing the MSRV bump necessary for this impl. I was actually thinking of bumping to Rust 1.85 and doing a migration to Rust 2024.

Hah, that would be great! Do you want me to also bump MSRV as part of this PR, or would you prefer to do it separately, in a different change, etc?

BurntSushi:
I can do that in a separate change.

PR #1337, opened 2026-03-16
fix: prevent panic in cache reset methods when cache is None

### LINKED ISSUE #1155: `meta::Cache::reset` can panic

Latest / `0.4.4`

Calling `.reset(..)` on a `meta::Cache` can panic if called on a different regex than it was created with (which is the whole point of `reset`).

```rust
use regex_automata::meta::Regex;

fn main() {
    let re1 = Regex::new("").unwrap();
    let re2 = Regex::new("\\b").unwrap();
    re1.create_cache().reset(&re2);
}
```

```
thread 'main' panicked at /playground/.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-automata-0.4.3/src/meta/wrappers.rs:507:29:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:127:5
   3: core::option::Option<T>::unwrap
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/option.rs:931:21
   4: regex_automata::meta::wrappers::OnePassCache::reset
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-automata-0.4.3/src/meta/wrappers.rs:507:13
   5: <regex_automata::meta::strategy::Core as regex_automata::meta::strategy::Strategy>::reset_cache
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-automata-0.4.3/src/meta/strategy.rs:688:9
   6: regex_automata::meta::regex::Cache::reset
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/regex-automata-0.4.3/src/meta/regex.rs:2385:9
   7: playground::main
             at ./src/main.rs:6:5
   8: core::ops::function::FnOnce::call_once
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

It should not panic, and allow using that cache on the new regex.

#### Additional info

The panic reproduced above comes from: https://github.com/rust-lang/regex/blob/master/regex-automata/src/meta/wrappers.rs#L799
although it looks like other `reset` implementations have similar issues.

saschabuehrle:
## Bug
Fixes #1155 — `meta::Cache::reset` can panic if called on a different regex than it was created with.

## Root Cause
The `reset` methods in `OnePassCache`, `HybridCache`, and `ReverseHybridCache` checked if `builder.0` was `Some` but then called `self.0.as_mut().unwrap()` without verifying `self.0` was also `Some`. This could happen when a cache was created with `::none()` (which sets the internal cache to `None`) but then reset with a builder that had an engine.

## Fix
The fix checks if `self.0` is `Some` before calling reset:
- If `Some`, reset the existing cache
- If `None`, create a new cache from the builder's engine

This allows caches to be safely reset with any compatible regex, which is the intended behavior of the `reset` method.

## Testing
The reproduction case from the issue now works without panicking:
```rust
let re1 = Regex::new("").unwrap();
let re2 = Regex::new("\\b").unwrap();
re1.create_cache().reset(&re2); // No longer panics
```

Greetings, saschabuehrle

### DIFF
--- a/regex-automata/src/meta/wrappers.rs
+++ b/regex-automata/src/meta/wrappers.rs
@@ -489,7 +489,11 @@ impl OnePassCache {
     pub(crate) fn reset(&mut self, builder: &OnePass) {
         #[cfg(feature = "dfa-onepass")]
         if let Some(ref e) = builder.0 {
-            self.0.as_mut().unwrap().reset(&e.0);
+            if let Some(ref mut cache) = self.0 {
+                cache.reset(&e.0);
+            } else {
+                self.0 = Some(e.0.create_cache());
+            }
         }
     }
 
@@ -781,7 +785,11 @@ impl HybridCache {
     pub(crate) fn reset(&mut self, builder: &Hybrid) {
         #[cfg(feature = "hybrid")]
         if let Some(ref e) = builder.0 {
-            self.0.as_mut().unwrap().reset(&e.0);
+            if let Some(ref mut cache) = self.0 {
+                cache.reset(&e.0);
+            } else {
+                self.0 = Some(e.0.create_cache());
+            }
         }
     }
 
@@ -1186,7 +1194,11 @@ impl ReverseHybridCache {
     pub(crate) fn reset(&mut self, builder: &ReverseHybrid) {
         #[cfg(feature = "hybrid")]
         if let Some(ref e) = builder.0 {
-            self.0.as_mut().unwrap().reset(&e.0);
+            if let Some(ref mut cache) = self.0 {
+                cache.reset(&e.0);
+            } else {
+                self.0 = Some(e.0.create_cache());
+            }
         }
     }
 


pawamoy:
Sorry to interrupt, I wanted to let you know that this PR is most probably entirely AI-generated, including replies to your comments. Just see the batch of PRs they opened recently to form your own opinion.

PR #1346, opened 2026-05-11
meta: fix non-leftmost match in reverse-suffix/inner optimizations

### LINKED ISSUE #1345: meta::Regex returns non-leftmost match for pattern with negated class around optional group
## Summary

For the pattern `[^()]*(?:\([^()]*\))?[^()]*:` searched against the haystack `$(:):`, `regex_automata::meta::Regex::find` returns a match at `2..3` (just the colon between the parentheses). `regex_automata::nfa::thompson::pikevm::PikeVM` and `regex_automata::hybrid::regex::Regex` both return `0..5` (the entire string) on the same input. The PikeVM/lazy-DFA result is the leftmost match — META is picking a strictly later start position.

## Repro

`Cargo.toml`:

```toml
[package]
name = "repro"
version = "0.0.0"
edition = "2021"

[dependencies]
regex-automata = "0.4.14"
```

`src/main.rs`:

```rust
use regex_automata::Input;

fn main() {
    let pat = r"[^()]*(?:\([^()]*\))?[^()]*:";
    let hay = b"$(:):";

    let meta = regex_automata::meta::Regex::new(pat).unwrap().find(hay);

    let pikevm = {
        use regex_automata::nfa::thompson::pikevm::PikeVM;
        let pv = PikeVM::new(pat).unwrap();
        let mut c = pv.create_cache();
        pv.find_iter(&mut c, Input::new(hay)).next()
    };

    let dfa = {
        use regex_automata::hybrid::regex::Regex as H;
        let r = H::new(pat).unwrap();
        let mut c = r.create_cache();
        r.find(&mut c, Input::new(hay))
    };

    println!("META  : {:?}", meta.map(|m| (m.start(), m.end())));
    println!("PikeVM: {:?}", pikevm.map(|m| (m.start(), m.end())));
    println!("DFA   : {:?}", dfa.map(|m| (m.start(), m.end())));
}
```

`cargo run --release` output:

```
META  : Some((2, 3))
PikeVM: Some((0, 5))
DFA   : Some((0, 5))
```

## Discussion

Standard PCRE/RE2 leftmost-first semantics on `$(:):` for the pattern above:

- At position 0, the pattern matches `$(:):` (length 5): `[^()]*` → `$`, `(?:\([^()]*\))?` → `(:)`, `[^()]*` → empty, `:` → `:`.
- At position 2, the pattern matches `:` (length 1): both `[^()]*` are empty, the optional group is empty, the literal `:` matches.

Position 0 is strictly earlier, so it's the leftmost match. PikeVM and the lazy DFA agree on it. META picks the position-2 match.

The pattern shape is artificial-looking but I ran into it via a real-world syntect / sublime-syntax grammar; the [reduction](https://github.com/fancy-regex/fancy-regex/pull/249#issuecomment-4415079363) traces back to a fancy-regex seek pre-filter approximation for a Makefile rule-detection lookahead. Three properties seem to matter for triggering it:

- A negated character class (e.g. `[^()]`) on both sides of an optional group.
- An optional group whose body can match at multiple offsets inside the negation gap (the `\([^()]*\)` here — the `\(` and `\)` are literally outside the negated class, and `[^()]*` inside is permissive).
- A trailing literal (`:`) that appears both inside *and* outside the gap.

The earliest variant I have where all three engines agree on the leftmost is when the leading negated-class run is forced to non-empty by an equality somewhere in the haystack ([`$(a=b):` agrees across engines](https://github.com/fancy-regex/fancy-regex/pull/249#issuecomment-4415079363) — there META correctly picks `0..6`). It seems to be the *combination* of a permissively-empty leading run plus an optional group that can absorb the rest that drives META's overshoot.

I've confirmed the same shape of divergence on slightly longer patterns derived from a C++ `<>` template-detection lookahead (META picks the inner identifier; PikeVM picks the outer one), so the pattern above is the smallest of a small family. Happy to share the longer variants and the syntect-side context if useful, but the 5-byte haystack above should be enough to reproduce.

## Versions

- `regex-automata = "0.4.14"` (latest at time of filing — `cargo update` resolves to `0.4.14`)
- Rust 1.91.1 (release build, no LTO)
- macOS 25.4.0, aarch64

Reduced via [stefanobaghino/fancy-regex@wip/diagnostic-pr249](https://github.com/stefanobaghino/fancy-regex/tree/wip/diagnostic-pr249) (`src/seek.rs::tests::diagnostic_*`) where it surfaces as a leftmost-first disagreement between META and PikeVM on the seek over-approximation for three independent `.sublime-syntax` grammars. See [fancy-regex/fancy-regex#249](https://github.com/fancy-regex/fancy-regex/pull/249) for the downstream context.

stefanobaghino:
Closes #1345.

`ReverseSuffix::try_search_half_start` and `ReverseInner::try_search_full` returned on the first successful reverse search, which assumed that the leftmost occurrence of the suffix (resp. inner) literal corresponds to the end of the leftmost-first match. That assumption breaks when the regex prefix is non-monotonic — e.g. an optional group whose body can absorb the literal between consecutive occurrences — so a strictly later literal occurrence may have a strictly earlier overall match start. The minimal repro filed in #1345 is `[^()]*(?:\([^()]*\))?[^()]*:` on `$(:):`: META was returning `2..3`, while the leftmost-first match is `0..5` (the optional group absorbs `(:)` and pins to the second `:`).

Both loops now track the best (smallest-start) candidate and bound each subsequent reverse search by `best.offset()`. The existing limited reverse DFA in `regex-automata/src/meta/limited.rs` returns `RetryError::Quadratic` the moment scanning past the bound is required, which propagates through `?` and triggers the existing `Core` fallback in `Strategy::search`. The loops also early-exit when the candidate already starts at `input.start()`, so the common (monotonic-prefix) cases stay one-shot.

The bug was reduced from real-world `.sublime-syntax` grammars driven through fancy-regex's seek-prefilter approximation; see fancy-regex/fancy-regex#249 for the downstream context.

`testdata/regression.toml` gets the minimal repro as `non-monotonic-reverse-suffix`; it failed against `meta` and passed against `pikevm`/`hybrid` before this change.

### DIFF
--- a/regex-automata/src/meta/strategy.rs
+++ b/regex-automata/src/meta/strategy.rs
@@ -1214,10 +1214,17 @@ impl ReverseSuffix {
         input: &Input<'_>,
     ) -> Result<Option<HalfMatch>, RetryError> {
         let mut span = input.get_span();
-        let mut min_start = 0;
+        let mut best: Option<HalfMatch> = None;
         loop {
+            // We can't possibly improve on a match that already starts at the
+            // very beginning of the search span.
+            if let Some(b) = best {
+                if b.offset() == input.start() {
+                    return Ok(best);
+                }
+            }
             let litmatch = match self.pre.find(input.haystack(), span) {
-                None => return Ok(None),
+                None => break,
                 Some(span) => span,
             };
             trace!("reverse suffix scan found suffix match at {litmatch:?}");
@@ -1225,20 +1232,33 @@ impl ReverseSuffix {
                 .clone()
                 .anchored(Anchored::Yes)
                 .span(input.start()..litmatch.end);
+            // We bound the reverse scan by the current best start position:
+            // any match at a position >= best.offset() can't improve on the
+            // current candidate, and if the reverse DFA would need to scan
+            // past best.offset() to find a strictly-better start, it returns
+            // Quadratic, which propagates to the caller and falls back to the
+            // core engine. This preserves leftmost-first semantics without
+            // unbounded re-scanning: if the prefix matching is "monotonic" in
+            // the suffix end (the common case), the first iteration's reverse
+            // search already finds the leftmost start and we early-exit;
+            // otherwise we fall back to the core engine.
+            let min_start = best.map_or(input.start(), |b| b.offset());
             match self
                 .try_search_half_rev_limited(cache, &revinput, min_start)?
             {
-                None => {
-                    if span.start >= span.end {
-                        break;
+                None => {}
+                Some(hm) => {
+                    if best.map_or(true, |b| hm.offset() < b.offset()) {
+                        best = Some(hm);
                     }
-                    span.start = litmatch.start.checked_add(1).unwrap();
                 }
-                Some(hm) => return Ok(Some(hm)),
             }
-            min_start = litmatch.end;
+            if span.start >= span.end {
+                break;
+            }
+            span.start = litmatch.start.checked_add(1).unwrap();
         }
-        Ok(None)
+        Ok(best)
     }
 
     #[cfg_attr(feature = "perf-inline", inline(always))]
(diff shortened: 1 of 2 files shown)


stefanobaghino:
Addressed already in https://github.com/rust-lang/regex/commit/64ad0b618e043b791ed5385dd5504a436da1ddae, closing this.

PR #1350, opened 2026-05-19
perf: literal-prefix capture-extraction fast path

Dandandan:
For anchored regexes of the shape

    ^<literal-prefix-set>([^X]+)X.*$    with replacement `\${1}` / `\$1`

capture 1's bounds are structurally trivial: skip the prefix, find the terminator with memchr. This PR makes both the meta engine and `Regex::replacen` exploit that.

Also see https://github.com/BurntSushi/rebar/pull/28 for the benchmark - here it shows a 14x improvement.

## Changes

- **`regex-automata`** — new `LiteralPrefixCapture` strategy in `meta/strategy.rs`. Recognizes the shape via HIR walking (single pattern, anchored start+end, default flags, ASCII terminator, finite literal-alternation prefix set capped at 32). Strategy methods extract the slots directly with memchr, bypassing PikeVM/BoundedBacktracker. Wires in alongside the existing reverse strategies. Pattern fails recognition -> `Err(core)` falls through.
- **`regex`** — `Regex::replacen` gets a borrowed-output fast path for replacements that are exactly `\$N` / `\${N}`, detected via a new `Replacer::single_capture_ref` method (default `None`; opted into for `&str` / `String` / `Cow<str>`). When `limit == 1` and the match covers the whole haystack, returns `Cow::Borrowed` of the captured slice — no `Captures::expand`, no output `String` allocation.

## Bench

500k synthetic Referer-style rows, 5-iter mean, same machine:

| Pattern, corpus | Before | After |
|---|---:|---:|
| `^https?://(?:www\.)?([^/]+)/.*\$`, 80% match | 281 ms | **39 ms** (7.3x) |
| `^key=([^,]+),.*\$`, 100% match | 113 ms | **27 ms** (4.2x) |

## Remaining overhead

`Regex::captures_read` with a reused `CaptureLocations` is only ~1.2x the prototype direct-memchr path on the same workload, so the meta-engine + Strategy dispatch is essentially free. The remaining gap to the prototype is dominated by the per-call `Captures` slot allocation in `Regex::captures` — pooling that is a separate change.

### DIFF
--- a/regex-automata/src/meta/strategy.rs
+++ b/regex-automata/src/meta/strategy.rs
@@ -1997,6 +1997,8 @@ impl LiteralPrefixCapture {
             return None;
         }
         let bytes = input.haystack();
+        let must_validate_utf8 =
+            self.requires_valid_utf8 && !input.haystack_is_known_valid_utf8();
         'prefix: for prefix in self.prefixes.iter() {
             if !bytes.starts_with(prefix) {
                 continue;
@@ -2029,8 +2031,7 @@ impl LiteralPrefixCapture {
             {
                 continue;
             }
-            if self.requires_valid_utf8 && core::str::from_utf8(bytes).is_err()
-            {
+            if must_validate_utf8 && core::str::from_utf8(bytes).is_err() {
                 return None;
             }
             return Some((cap_start, cap_end));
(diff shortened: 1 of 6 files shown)


neilconway:
@Dandandan Are you aware of any open issues in this PR? Would be great to get it to non-draft status; if I can help with something, lmk.

PR #1351, opened 2026-05-23
Implement `Regex::source_and_options_equals`

### LINKED ISSUE #1349: Provide an API for comparing the equivalence of two `Regex` values (while permitting false negatives)
I'd like to test a function that is supposed to output a regular expression, however I wasn't able to find a reasonably accurate way to do this, and I suspect there currently isn't any.

I'm aware that the reason for removal has been explained in #178. A workaround using `r1.as_str() == r2.as_str()` has been mentioned in both #313 and #364, however this does not appear to handle the cases where the regexes are compiled from the same string while using different options, i.e. has false positives in the context of "do two regexes return the same output for any possible input?".

A method that would not require the user of this library to maintain their own comparison code (i.e. `r1.source_and_options_equals(r2)`) would be preferable to a mechanism for exposing the compile options (i.e. `r1.as_str() == r2.as_str() && r1.case_insensitive() == r2.case_insensitive() && ...`).

dolezvo1:
`Regex::source_and_options_equals` provides API for testing regex equivalence while permitting false negatives (closes #1349)

### DIFF
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -101,6 +101,7 @@ use crate::{error::Error, RegexBuilder};
 pub struct Regex {
     pub(crate) meta: meta::Regex,
     pub(crate) pattern: Arc<str>,
+    pub(crate) syntaxc: Arc<regex_automata::util::syntax::Config>,
 }
 
 impl core::fmt::Display for Regex {
@@ -1274,6 +1275,31 @@ impl Regex {
     #[inline]
     pub fn as_str(&self) -> &str {
         &self.pattern
+    }
+
+    /// Compares the source strings and compile options of two regexes,
+    /// returning true iff both are equal.
+    /// Such result implies equivalence of the regexes, but the contrary
+    /// provides no information about their equivalence.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use regex::{Regex, RegexBuilder};
+    ///
+    /// let r1 = Regex::new(r"a+").unwrap();
+    /// let r2 = Regex::new(r"aa*").unwrap();
+    /// assert!(!r1.source_and_options_equals(&r2));
+    ///
+    /// let r3 = RegexBuilder::new(r"a+")
+    ///     .case_insensitive(true)
+    ///     .build()
+    ///     .unwrap();
+    /// assert!(!r1.source_and_options_equals(&r3));
+    /// ```
+    #[inline]
+    pub fn source_and_options_equals(&self, other: &Self) -> bool {
+        self.pattern == other.pattern && self.syntaxc == other.syntaxc
     }
 
     /// Returns an iterator over the capture names in this regex.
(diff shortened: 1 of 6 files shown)


src/regex/string.rs
```diff
@@ -1276,6 +1277,31 @@ impl Regex {
         &self.pattern
     }
 
+    /// Compares the source strings and compile options of two regexes,
```
dolezvo1: *syntax compile options

src/regex/string.rs
```diff
@@ -1276,6 +1277,31 @@ impl Regex {
...
+    /// assert!(!r1.source_and_options_equals(&r2));
+    ///
+    /// let r3 = RegexBuilder::new(r"a+")
+    ///     .case_insensitive(true)
+    ///     .build()
+    ///     .unwrap();
+    /// assert!(!r1.source_and_options_equals(&r3));
+    /// ```
+    #[inline]
+    pub fn source_and_options_equals(&self, other: &Self) -> bool {
```
BurntSushi, line 1306: This naming will need work. I don't like how it is right now. But I don't have the spoons to come up with a better one at the moment.

src/regex/string.rs
```diff
@@ -101,6 +101,7 @@ use crate::{error::Error, RegexBuilder};
 pub struct Regex {
     pub(crate) meta: meta::Regex,
     pub(crate) pattern: Arc<str>,
+    pub(crate) syntaxc: Arc<regex_automata::util::syntax::Config>,
```
BurntSushi: This isn't a good idea here. `Regex` already being 32 bytes is kind of not great. Adding this here makes that worse for a rather niche use case, which isn't a good trade-off IMO.

I think ideally we'd tuck this down inside of `regex_automata::meta::Regex`. But that cannot be done because a meta regex can be constructed from `regex_syntax::Hir` directly, which means a syntax config may not be present.

That leaves the least bad option here as a new struct here:

```rust
struct RegexInfo {
    syntax_config: regex_automata::util::syntax::Config,
    pattern: Arc<str>,
}
```

And `regex::Regex` becomes:

```rust
pub struct Regex {
    pub(crate) meta: meta::Regex,
    pub(crate) info: Arc<RegexInfo>,
}
```

This should also shrink the size of `Regex` to 24 bytes on 64-bit targets. Which is a nice little bonus.

BurntSushi (changes requested):
I think this is on the right track. We'll want to add this to the other regex APIs (bytes, set and bytes set). And also `regex-lite`.

src/regex/string.rs
dolezvo1: I think you're right that the name could be improved, but I'm not quite sure how. The `_equals` part I believe should stay. `source_and_options` could perhaps be replaced by `compile_inputs` or `regex_info`, but I'm not sure whether that is better. Replacing `source` by `str` or `pattern` could also make sense.

dolezvo1: @BurntSushi Hello, did you happen to have any ideas about a better name?

PR #1355, opened 2026-06-02
fix: Return the leftmost match from ReverseSuffix

### LINKED ISSUE #1354: Reverse suffix doesn't return the leftmost match when alternates overlap

1.12.3

The reverse suffix strategy doesn't return the leftmost match first when the first alternate overlaps with the second alternate.

```rust
use regex; // 1.12.3

fn main() {
    let r = regex::Regex::new(".abb|b").unwrap();

    let m = r.find("zabb").unwrap();

    assert_eq!(m.as_str(), "zabb")
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2273db01d4564806836a207fb7af5c52

The first match is "b".

The first match should be the leftmost match, "zabb".

JakubValtar:
This PR delays returning the first half-start match from `ReverseSuffix` strategy until we confirm that there isn't another match after it that would be more leftmost, which can happen for patterns where alternates are overlapping.

I'm a first time contributor and went with a straightforward way to fix the issue, any feedback is welcome. If there's a better approach to fix it, I'm also happy to rework the PR.

- Fixes #1354

### DIFF
--- a/regex-automata/src/meta/strategy.rs
+++ b/regex-automata/src/meta/strategy.rs
@@ -1215,9 +1215,10 @@ impl ReverseSuffix {
     ) -> Result<Option<HalfMatch>, RetryError> {
         let mut span = input.get_span();
         let mut min_start = 0;
+        let mut first: Option<HalfMatch> = None;
         loop {
             let litmatch = match self.pre.find(input.haystack(), span) {
-                None => return Ok(None),
+                None => break,
                 Some(span) => span,
             };
             trace!("reverse suffix scan found suffix match at {litmatch:?}");
@@ -1228,7 +1229,13 @@ impl ReverseSuffix {
             if let Some(hm) =
                 self.try_search_half_rev_limited(cache, &revinput, min_start)?
             {
-                return Ok(Some(hm));
+                // We only track the first half-match. We can never find a
+                // half-match with lower offset than the first half-match, it
+                // would mean quadratic behavior. We can only confirm this
+                // half-match or return a quadratic error.
+                if first.is_none() {
+                    first = Some(hm);
+                }
             }
 
             if span.start >= span.end {
@@ -1237,7 +1244,7 @@ impl ReverseSuffix {
             span.start = litmatch.start.checked_add(1).unwrap();
             min_start = litmatch.end;
         }
-        Ok(None)
+        Ok(first)
     }
 
     #[cfg_attr(feature = "perf-inline", inline(always))]
@@ -1623,9 +1630,10 @@ impl ReverseInner {
         let mut span = input.get_span();
         let mut min_match_start = 0;
         let mut min_pre_start = 0;
+        let mut first: Option<Match> = None;
         loop {
             let litmatch = match self.preinner.find(input.haystack(), span) {
-                None => return Ok(None),
+                None => break,
                 Some(span) => span,
             };
             if litmatch.start < min_pre_start {
@@ -1662,10 +1670,12 @@ impl ReverseInner {
                         span.start = litmatch.start.checked_add(1).unwrap();
                     }
                     Ok(hm_end) => {
-                        return Ok(Some(Match::new(
-                            hm_start.pattern(),
-                            hm_start.offset()..hm_end.offset(),
-                        )))
+                        if first.is_none() {
+                            first = Some(Match::new(
+                                hm_start.pattern(),
+                                hm_start.offset()..hm_end.offset(),
+                            ));
+                        }
                     }
                 }
             }
@@ -1676,7 +1686,7 @@ impl ReverseInner {
             span.start = litmatch.start.checked_add(1).unwrap();
             min_match_start = litmatch.end;
         }
-        Ok(None)
+        Ok(first)
     }
 
     #[cfg_attr(feature = "perf-inline", inline(always))]
--- a/regex-automata/src/meta/strategy.rs
+++ b/regex-automata/src/meta/strategy.rs
@@ -1225,17 +1225,16 @@ impl ReverseSuffix {
                 .clone()
                 .anchored(Anchored::Yes)
                 .span(input.start()..litmatch.end);
-            match self
-                .try_search_half_rev_limited(cache, &revinput, min_start)?
+            if let Some(hm) =
+                self.try_search_half_rev_limited(cache, &revinput, min_start)?
             {
-                None => {
-                    if span.start >= span.end {
-                        break;
-                    }
-                    span.start = litmatch.start.checked_add(1).unwrap();
-                }
-                Some(hm) => return Ok(Some(hm)),
+                return Ok(Some(hm));
             }
+
+            if span.start >= span.end {
+                break;
+            }
+            span.start = litmatch.start.checked_add(1).unwrap();
             min_start = litmatch.end;
         }
         Ok(None)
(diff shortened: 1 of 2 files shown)


JakubValtar:
Hm, this might need more work. It returns the first match correctly, but matches get skipped when searching for all matches.

```
regex-cli find match meta -p '.abb|b' -y 'zabbb'

  total matches:  2
0:0:4:zabb
0:4:5:b
```

Edit: This actually looks like the correct behavior for overlapping matches.

BurntSushi:
That looks right to me for all non-overlapping successive matches.

Overlapping is a totally different ball game and the meta regex engine (which is where this bug is) doesn't provide a way to do overlapping search. You can only get overlapping when you drop down to the lower layers:

```
$ cargo r -qp regex-cli -- find half hybrid --no-table -p '.abb|b' -y 'zabbb' -kall --overlapping
0:3
0:4
0:5
```

regex-automata/src/meta/strategy.rs
```diff
@@ -1213,32 +1214,58 @@ impl ReverseSuffix {
         cache: &mut Cache,
         input: &Input<'_>,
     ) -> Result<Option<HalfMatch>, RetryError> {
+        let max_len =
+            self.core.info.props_union().maximum_len().unwrap_or(usize::MAX);
```
BurntSushi: This is my main hesitation with this fix. In all but very simple cases for fixed length regexes, `max_len` is going to be `usize::MAX` here. Which means any branch below for `x >= max_len` is basically never going to be true.

I imagine this was added for performance reasons.

Anyway, I'm going to dig into benchmarking this a bit more tomorrow. Thank you for getting the ball rolling here.

JakubValtar: I think that's a fundamental issue with this strategy, you can't return a half-start until you know it's leftmost.

Previously, reverse suffix would often also bail early due to the quadratic behavior when the suffix could be matched elsewhere in the pattern, e.g.
`regex-cli -- find match meta -p '[a-z]b' -y '000bbb'`, which I assume would be the case for many patterns that don't have max size.

Curious about the benchmarks.

BurntSushi: I ran the full suite of rebar benchmarks for `rust/regex` before and after this PR, and there aren't any regressions there from what I can see:

```
[andrew@duff rebar]$ rebar diff tmp/regex-master.csv tmp/regex-fix-reverse-suffix-leftmost.csv -t 1.2
benchmark                                  engine      tmp/regex-master.csv  tmp/regex-fix-reverse-suffix-leftmost.csv
---------                                  ------      --------------------  -----------------------------------------
imported/sherlock/repeated-class-negation  rust/regex  17.3 GB/s (1.39x)     24.0 GB/s (1.00x)
```

(Some benchmarks are noisy, so I'd ignore the improvement here.)

BurntSushi: OK, so I dove deeper into this and I flubbed the benchmark. This change has significant performance regressions unfortunately. I'm going to need to think about this one.

The reverse inner optimization has the same kind of bug too.

Disabling these optimizations would fix the bug, but these are critical load bearing optimizations.

BurntSushi: If you think about what this change is doing, it virtually guarantees that it will exhaustively search the entire remaining haystack for _every_ occurrence of literal (assuming the existing quadratic trip-wire doesn't catch it first, which it won't in cases where the literal doesn't overlap with the first part of the regex). This is even worse than disabling the optimization in the first place and itself introduces more quadratic behavior when iterating over matches. So I think this particular fix is firmly bunk unfortunately.

BurntSushi: We might need to actually compute whether there is overlap between the suffix (or inner) literal and the parts of the pattern that come before that. My quadratic trip-wire was partially meant as a hack to avoid doing that.

BurntSushi: > OK, so I dove deeper into this and I flubbed the benchmark.

Actual benchmark results:

```
[andrew@duff rebar]$ rebar diff tmp/regex-master.csv tmp/regex-fix-reverse-suffix-leftmost-04.csv -t 1.2
benchmark                                            engine      tmp/regex-master.csv  tmp/regex-fix-reverse-suffix-leftmost-04.csv
---------                                            ------      --------------------  --------------------------------------------
curated/06-cloud-flare-redos/simplified-long         rust/regex  83.9 GB/s (1.00x)     55.1 GB/s (1.52x)
hyperscan/literal-suffix-nosom                       rust/regex  19.2 GB/s (1.00x)     148.1 MB/s (132.48x)
hyperscan/literal-suffix-som                         rust/regex  19.4 GB/s (1.00x)     129.8 MB/s (153.06x)
hyperscan/literal-inner-nosom                        rust/regex  22.2 GB/s (1.00x)     311.7 MB/s (73.00x)
hyperscan/literal-inner-som                          rust/regex  22.2 GB/s (1.00x)     292.5 MB/s (77.79x)
imported/leipzig/shing                               rust/regex  25.7 GB/s (1.00x)     33.8 MB/s (776.71x)
imported/leipzig/word-ending-nn                      rust/regex  31.5 GB/s (1.00x)     348.3 MB/s (92.73x)
imported/leipzig/certain-long-strings-ending-x       rust/regex  5.6 GB/s (1.00x)      330.9 MB/s (17.29x)
imported/leipzig/tom-sawyer-huckle-fin-prefix-short  rust/regex  19.3 GB/s (1.00x)     377.8 MB/s (52.24x)
imported/leipzig/tom-sawyer-huckle-fin-prefix-long   rust/regex  18.6 GB/s (1.00x)     410.9 MB/s (46.30x)
imported/leipzig/ing                                 rust/regex  3.1 GB/s (1.00x)      99.5 MB/s (31.78x)
imported/leipzig/ing-whitespace                      rust/regex  3.0 GB/s (1.00x)      133.1 MB/s (22.76x)
imported/leipzig/awyer-inn                           rust/regex  14.6 GB/s (1.00x)     153.5 MB/s (97.55x)
imported/mariomka/email                              rust/regex  50.5 GB/s (1.00x)     1032.1 MB/s (50.11x)
imported/mariomka/uri                                rust/regex  9.2 GB/s (1.00x)      11.1 MB/s (848.70x)
imported/sherlock/before-holmes                      rust/regex  18.0 GB/s (1.00x)     147.8 MB/s (124.43x)
imported/sherlock/before-after-holmes                rust/regex  20.9 GB/s (1.00x)     297.1 MB/s (71.94x)
imported/sherlock/word-ending-n                      rust/regex  838.4 MB/s (1.00x)    119.2 MB/s (7.03x)
imported/sherlock/repeated-class-negation            rust/regex  17.3 GB/s (1.00x)     368.4 MB/s (48.05x)
imported/sherlock/ing-suffix                         rust/regex  3.5 GB/s (1.00x)      136.4 MB/s (25.91x)
imported/sherlock/ing-suffix-limited-space           rust/regex  3.5 GB/s (1.00x)      166.9 MB/s (21.76x)
opt/reverse-inner/holmes                             rust/regex  19.1 GB/s (1.00x)     297.1 MB/s (65.75x)
opt/reverse-inner/email                              rust/regex  82.2 GB/s (1.00x)     29.4 GB/s (2.80x)
opt/reverse-inner/factored-prefix                    rust/regex  9.2 GB/s (1.00x)      40.0 MB/s (235.27x)
opt/reverse-suffix/holmes                            rust/regex  19.3 GB/s (1.00x)     148.5 MB/s (133.19x)
reported/i13-subset-regex/original-ascii             rust/regex  5.6 GB/s (1.00x)      330.5 MB/s (17.37x)
reported/i13-subset-regex/original-unicode           rust/regex  345.6 MB/s (1.00x)    17.9 MB/s (19.35x)
reported/i13-subset-regex/big-ascii                  rust/regex  92.1 MB/s (1.00x)     16.4 MB/s (5.62x)
reported/i13-subset-regex/big-unicode                rust/regex  36.1 MB/s (1.00x)     15.6 MB/s (2.31x)
```

JakubValtar: Thanks for the updated benchmark!

This optimization looks for ends of the matches, but it needs to know the start to return the leftmost match, which sounds like quite a complex problem and might indeed need a change in approach or larger design changes rather than a "fix" like this PR.

I don't think I'm able to contribute further to resolve this issue (apart from high-level ideas) as I don't have enough experience with this crate. I appreciate you looking into it and I'm still curious about the eventual approach!

BurntSushi: Oh it's absolutely brutal. See: https://github.com/rust-lang/regex/pull/1364

I will take it from here. Thank you for the report and attempt. This turned out to be quite the bug!

PR #1361, opened 2026-06-05
Remove unnecessary clones

FranciscoTGouveia:
If I am not overlooking anything, these clones can be safely removed, either because they can be copied or moved instead.

CC @nunoplopes


FranciscoTGouveia:
If possible, could I get a review on this, @BurntSushi?

PR #1362, opened 2026-06-05
docs: clarify unsupported Anchored::Pattern searches

### LINKED ISSUE #1262: add hint about `starts_for_each_pattern` to `regex_automata::MatchErrorKind::UnsupportedAnchored` documentation
I’ve been there, but the comment is a bit sparse:

> An error indicating that a particular type of anchored search was requested, but that the regex engine does not support it.

Just saying that the most common cases/mandatory conditions for DFA/NFA are to have the `starts_for_each_pattern` might help (I assume that `starts_for_each_pattern` is mandatory for `Anchored(PatternId(_))` variant; I don’t know what’s necessary for the `Yes` one)

_Originally posted by @gagbo in https://github.com/rust-lang/regex/discussions/1261#discussioncomment-12971840_

marko1olo:
Closes #1262.

This expands the MatchErrorKind::UnsupportedAnchored rustdoc with the common Anchored::Pattern failure mode for DFA engines built without per-pattern anchored start states, and points readers at starts_for_each_pattern.

Validation:
- git show --check --oneline HEAD
- Static search for starts_for_each_pattern / Anchored::Pattern terminology in 
egex-automata

Not run locally:
- cargo fmt --check
- RUSTDOCFLAGS='-D warnings' cargo doc -p regex-automata --no-deps

Reason: this Windows environment does not currently have cargo on PATH. The change is rustdoc-only and limited to 
egex-automata/src/util/search.rs.

### DIFF
--- a/regex-automata/src/util/search.rs
+++ b/regex-automata/src/util/search.rs
@@ -1879,6 +1879,12 @@ pub enum MatchErrorKind {
     /// An error indicating that a particular type of anchored search was
     /// requested, but that the regex engine does not support it.
     ///
+    /// When this error occurs with [`Anchored::Pattern`], then one common
+    /// cause is that the underlying regex engine was built without support
+    /// for anchored starting states for each pattern. For DFA engines, this
+    /// typically means enabling the `starts_for_each_pattern` configuration
+    /// option when building the searcher.
+    ///
     /// Note that this error should not be returned by a regex engine simply
     /// because the pattern ID is invalid (i.e., equal to or exceeds the number
     /// of patterns in the regex). In that case, the regex engine should report
--- a/regex-automata/src/util/search.rs
+++ b/regex-automata/src/util/search.rs
@@ -1882,8 +1882,11 @@ pub enum MatchErrorKind {
     /// When this error occurs with [`Anchored::Pattern`], then one common
     /// cause is that the underlying regex engine was built without support
     /// for anchored starting states for each pattern. For DFA engines, this
-    /// typically means enabling the `starts_for_each_pattern` configuration
-    /// option when building the searcher.
+    /// typically means enabling the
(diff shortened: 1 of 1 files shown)


regex-automata/src/util/search.rs
```diff
@@ -1879,6 +1879,12 @@ pub enum MatchErrorKind {
     /// An error indicating that a particular type of anchored search was
     /// requested, but that the regex engine does not support it.
     ///
+    /// When this error occurs with [`Anchored::Pattern`], then one common
+    /// cause is that the underlying regex engine was built without support
+    /// for anchored starting states for each pattern. For DFA engines, this
+    /// typically means enabling the `starts_for_each_pattern` configuration
```
BurntSushi: Can you turn `starts_for_each_pattern` into links for `hybrid::dfa::Config::starts_for_each_pattern` and `dfa::dense::Config::starts_for_each_pattern`? The visible text should include that path too  for each, so that it isn't ambiguous when reading it.

BurntSushi (changes requested):
Thank you!

regex-automata/src/util/search.rs
```diff
@@ -1879,6 +1879,15 @@ pub enum MatchErrorKind {
...
     /// requested, but that the regex engine does not support it.
     ///
+    /// When this error occurs with [`Anchored::Pattern`], then one common
+    /// cause is that the underlying regex engine was built without support
+    /// for anchored starting states for each pattern. For DFA engines, this
+    /// typically means enabling the
+    /// [`hybrid::dfa::Config::starts_for_each_pattern`](crate::hybrid::dfa::Config::starts_for_each_pattern)
+    /// or
+    /// [`dfa::dense::Config::starts_for_each_pattern`](crate::dfa::dense::Config::starts_for_each_pattern)
+    /// configuration option when building the searcher.
```
BurntSushi:
```suggestion
    /// configuration option, depending on which regex engine is being built.
```

BurntSushi (approved):

PR #1364, opened 2026-06-07
automata: fix bug in reverse suffix/inner optimization

### LINKED ISSUE #1354: Reverse suffix doesn't return the leftmost match when alternates overlap

BurntSushi:
A minimal reproducer of this bug is on a haystack of `zabb` with the
regex `.bb|b`. The `regex` crate will report a match at `2..3`, but the
correct match is `1..4`.

While this seems like a simple regex, there are a pretty specific set
of circumstances required to trigger the bug:

1. There are no prefix literals that activate a standard prefix
   literal scan.
2. There needs to be an extractable *suffix* or *inner* literal.
3. An actual match needs to be present in the haystack.
4. The regex and haystack need to be such that the first candidate
   (from a literal scan) corresponds to a match whose start offset
   is greater than the start offset of a match confirmed by a
   subsequent candidate.

Crucially, note that because of (3), this bug will never lead to
`Regex::is_match` providing a false positive *or* a false negative.
This bug is strictly about leftmost-first match semantics being
incorrect in some cases and will report an incorrect match span.

(4) could do with a bit more explanation, since it's rather subtle.
Let's trace the minimal example through the regex crate's "reverse
suffix" optimization.

During compilation, there is no prefix literal that can be extracted.
The `.` defeats that class of optimization. Moreover, there is a suffix
literal in the regex. That is, all matches for `.bb|b` must end with
`b`. The regex crate sees this and will scan for matches of `b`. It will
then attempt to match the regex in reverse at each candidate match of
`b`. Let's see what happens:

* Find first occurrence of `b` at offset `2` in `zabb`.
* Start reverse confirmation step at offset `2`.
* The second alternation branch, `b` in `.bb|b`, matches at `2..3`.
* The second alternation branch is reported as the overall match.
  This happens because the first alternation branch, `.bb`, does _not_
  have a match ending at offset `3`.

The fundamental problem here is that there is an overlap between the
reverse automaton for confirming the match and the literal scan. Small
changes, even to the haystack, can result in the bug disappearing.
For example, with a haystack of `zbb`, the correct match of `0..3` is
reported. This occurs because there is a quadratic "trip wire" that
triggers in this case that causes the search to bail out and fall back
to a DFA without using any literal optimizations.

This bug also applies to the "reverse inner" optimization. This
can happen when the literal is extracted from inside the regex
as opposed to it being a suffix literal. For example, the regex
`(?:..acbb|b)a(?:c|d)` on the haystack `xzbacbbac` reported a match at
`2..5`, but the correct match is `1..9`.

Note that #1355 technically fixes this problem and is much simpler, but
in so doing, makes the reverse suffix and inner optimizations completely
ineffective. In other words, it nearly disables them entirely.

This PR keeps them enabled, but does some work to determine whether the
regex pattern combined with the literals exatracted could ever lead to
(4) being true.

This does unfortunately cause some regressions. For example,
`(.*?,){13}z` no longer gets the reverse suffix optimization because it
is difficult to prove that reverse scan of `(.*?,){13}` will never lead
to a later match than the true leftmost-first match after finding an
occurrence of `z`.

Fixes #1354, Closes #1355


BurntSushi:
I wrote this PR with an LLM, so this still needs careful review. From a glance, the solution here seems very over complicated. But it does pass tests and rebar benchmarks.

BurntSushi:
I think this is probably good to go. I need to do another round of review, but I significantly simplified the original LLM fix for this. There will likely unfortunately be some performance regressions here, but I tried to claw back what I could.

BurntSushi:
I'm doing one last run of the rebar benchmark suite first.

BurntSushi:
rebar benchmarks diffed with current master:

```
$ rebar diff x/master.csv x/change07.csv -t 1.2
benchmark                                            engine      x/master.csv        x/change07.csv
---------                                            ------      ------------        --------------
curated/06-cloud-flare-redos/original                rust/regex  566.9 MB/s (1.00x)  314.9 MB/s (1.80x)
curated/06-cloud-flare-redos/simplified-long         rust/regex  84.7 GB/s (1.00x)   42.9 GB/s (1.97x)
imported/leipzig/word-ending-nn                      rust/regex  29.2 GB/s (1.00x)   825.5 MB/s (36.19x)
imported/leipzig/tom-sawyer-huckle-fin-prefix-short  rust/regex  19.2 GB/s (1.00x)   823.7 MB/s (23.88x)
imported/leipzig/tom-sawyer-huckle-fin-prefix-long   rust/regex  18.4 GB/s (1.00x)   823.7 MB/s (22.90x)
imported/leipzig/ing                                 rust/regex  3.1 GB/s (1.00x)    765.5 MB/s (4.10x)
imported/leipzig/bounded-strings-ending-z            rust/regex  49.0 GB/s (1.00x)   824.2 MB/s (60.94x)
imported/rsc/reallyhard0-1k                          rust/regex  10.9 GB/s (1.00x)   794.7 MB/s (14.00x)
imported/rsc/reallyhard0-32k                         rust/regex  45.4 GB/s (1.00x)   825.6 MB/s (56.37x)
imported/rsc/reallyhard0-1mb                         rust/regex  44.0 GB/s (1.00x)   826.5 MB/s (54.46x)
imported/sherlock/word-ending-n                      rust/regex  835.8 MB/s (1.00x)  545.6 MB/s (1.53x)
imported/sherlock/ing-suffix                         rust/regex  3.5 GB/s (1.00x)    694.6 MB/s (5.19x)
opt/reverse-inner/factored-prefix                    rust/regex  9.1 GB/s (1.00x)    803.5 MB/s (11.64x)
```

There ight be a way to claw some of these back, but it's hard and I've already been working on this bug fix for over a month now (off-and-on). So I wanted to ship a correctness fix before doing more work here.

PR #1366, opened 2026-06-11
automata: reject dense DFA start states that are match states

dxbjavid:
a serialized dense DFA can have a start-state table entry that points at a match state. dense::DFA::from_bytes checks that each start id indexes a real state, but unlike the sparse DFA it never checks that the id is not a match state, so the bytes are accepted. a later search then trips the debug assertion in dfa::search::init_fwd (start states can't be match states because matches are delayed a byte), and with assertions off it quietly breaks that delay invariant instead. this adds the same guard to StartTable::validate that the sparse DFA already has.

### DIFF
--- a/regex-automata/src/dfa/dense.rs
+++ b/regex-automata/src/dfa/dense.rs
@@ -4177,6 +4177,11 @@ impl<T: AsRef<[u32]>> StartTable<T> {
                     "found invalid starting state ID",
                 ));
             }
+            if dfa.special.is_match_state(id) {
+                return Err(DeserializeError::generic(
+                    "start states cannot be match states",
+                ));
+            }
         }
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


BurntSushi (commented):
Thank you!  Can you add a regression test for sparse DFAs too?

dxbjavid:
done, added the same test on the sparse side. the sparse check itself was already in place, so the test just pins the behaviour down: it points every start table entry at the minimum match state and checks that from_bytes rejects it with the same error. ran the full regex-automata suite and everything still passes.

dxbjavid:
Hi @BurntSushi ,

Just a friendly follow-up on this PR. When you have a chance, could you please provide an update on its review status?

Happy to make any additional changes if needed.

Thank you!

dxbjavid:
Thank you for the review and merge. Much Appreciated

Issue #1311, opened 2025-10-30
a prefilter should only be constructed if we know we'll use it


BurntSushi:
Currently, the prefilter is constructed before we select the regex search strategy. Some of those strategies, like the reverse anchored strategy, won't use a prefilter at all.

See https://github.com/rust-lang/regex/discussions/1309

Issue #1310, opened 2025-10-30
fix runaway generation of literals for case insensitive regexes


BurntSushi:
See https://github.com/rust-lang/regex/discussions/1309

PR #1368, opened 2026-06-29
Fix runaway literal extraction for case-insensitive regex sets (#1310, #1311)

StefanoD:
## Summary

Building a `RegexSet` (or even a single regex) out of many **case-insensitive**
patterns could blow up at construction time. This was reported in #1309 and
triaged into two bugs, both fixed here.

**Root cause.** Each case-insensitive pattern yields an exponential number of
*inexact* prefix/suffix literals (`(?i)abc` → `ABC, ABc, …, abc`). The
per-pattern `Extractor` caps these, but `prefixes()`/`suffixes()` union them
**across patterns** with the raw `Seq::union`, which has **no total cap** and
re-`dedup`s the whole growing list each time. So an N-pattern set is
effectively **O(N²)** to build and produces a giant literal set. For
end-anchored patterns (`…$`), that set is then compiled into a large
Aho-Corasick prefilter that the selected **reverse-anchored** strategy never
even uses.

## Changes

- **#1310 — bound the cross-pattern literal union**
  (`regex-automata/src/util/prefilter/mod.rs`). When the running *inexact*
  sequence exceeds a total limit, collapse it to a short shared prefix/suffix
  (the same keep-4-bytes trick as `Extractor::union`), and drop the prefilter
  entirely if it is still too big. **Exact** sequences (faithful enumerations,
  e.g. plain-literal sets) are left untouched, so that behavior is unchanged.
- **#1311 — don't build a prefilter we won't use**
  (`regex-automata/src/meta/strategy.rs`). Skip prefix literal extraction when
  the regex is always anchored at the end; those use the reverse-anchored
  strategy, which never consults a prefix prefilter. This mirrors the existing
  always-anchored-start skip.

Note: #1310 points at `limit_total` in `regex-syntax`; that per-pattern limit
already works — the leak is the cross-pattern union, which is why the fix lives
in `prefixes()`/`suffixes()`.

## Results

`(?i)domainN\.com` sets, build time (M2 Air):

| N | endswith (before → after) | contains (before → after) |
|---|---|---|
| 600 | ~150ms → ~10ms | ~243ms → ~8ms |
| 1200 | — | ~796ms → ~20ms |
| 4000 | super-linear → ~44ms | super-linear → ~37ms |

`debug` logs confirm end-anchored sets now log `skipping literal extraction
since regex is anchored at the end` (no Aho-Corasick built), and unanchored
case-insensitive sets extract a small bounded set (e.g. 16 literals → a fast
Teddy prefilter, vs 1280 → a 2690-state Aho-Corasick before).

## Test plan

- New unit tests in `prefilter/mod.rs`: the case-insensitive prefix/suffix
  sequences stay bounded (verified to produce 6400 literals *without* the cap),
  and a large *exact* set is preserved.
- New correctness cases in `testdata/regression.toml` and `testdata/set.toml`
  for end-anchored and unanchored case-insensitive sets (exercised by all four
  test suites).
- Full workspace suite passes (`regex-automata`: 459 doctests + 140 lib + 58
  integration; top-level integration: 64).

Note: This has been generated with Claude Opus 4.8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+1.12.5 (TBD)
+============
+This release fixes a performance problem when building a `RegexSet` (or even a
+single regex) out of many case-insensitive patterns. Literal extraction could
+blow up combinatorially and build a large prefilter that the chosen search
+strategy would then never use.
+
+Bug fixes:
+
+* [BUG #1310](https://github.com/rust-lang/regex/issues/1310):
+Bound the number of literals extracted across patterns. This fixes runaway
+literal generation (and quadratic build time) for case-insensitive regex sets.
+* [BUG #1311](https://github.com/rust-lang/regex/issues/1311):
+Skip building a prefix prefilter for regexes that are always anchored at the
+end, since those use the reverse anchored strategy and never consult it.
+
+
 1.12.4 (2025-06-09)
 ===================
 This release includes a performance optimization for compilation of regexes
--- a/regex-automata/src/meta/strategy.rs
+++ b/regex-automata/src/meta/strategy.rs
@@ -102,6 +102,24 @@ pub(super) fn new(
         // carry a 'Prefilter' with it, but I moved away from that.)
         debug!("skipping literal extraction since regex is anchored");
         None
+    } else if info.is_always_anchored_end() {
+        // If the regex is always anchored at the end, then the reverse
+        // anchored strategy (selected below) will be used. That strategy does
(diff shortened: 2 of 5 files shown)


BurntSushi:
I don't mind code changes being done with an LLM, but please don't use LLMs to communicate with humans. See: https://github.com/rust-lang/regex/pull/1369

StefanoD:
That's a pitty that you rather leave two issues unsolved than objectively looking into the PR. 🙂
Other projects accepted PRs without any issues.

PR #1371, opened 2026-07-03
Introduce the `regex!` macro for easier reusable regex

### LINKED ISSUE #709: provide a helper routine for building a static regex using std::lazy
Introduce a macro that lazily initialise a static regex expression, call it `lazy_regex`.
This is helpful whenever you want to avoid the compile-at-each-iteration pattern when using regexes across threads.
The code is simply,

```rust
macro_rules! lazy_regex_1 {
    ($re:literal $(,)?) => {{
        static RE: SyncOnceCell<regex::Regex> = SyncOnceCell::new();
        RE.get_or_init(|| Regex::new($re).unwrap())
    }};
}
```
or
```rust
macro_rules! lazy_regex_2 {
    ($re:literal $(,)?) => {{
        static RE: SyncLazy<regex::Regex> = SyncLazy::new(|| Regex::new($re).unwrap());
        &RE
    }};
}
```

Which is only a slight alteration from the one provided in the `once_cell` documentation. There are many possible variations of this --- there are many ways to lazily initialise a static variable. I think this is something that is common enough that it should be included.

Thank you.

tgross35:
Add a wrapper around `regex_automata`'s  `Lazy` as a simple way to
construct a `Regex` that is compiled once but used multiple times.
Sample usage:

    if regex!(r"\d+").is_match("123") { /* ... */ }
    let re: &Regex = regex!(r"\d+");

This idea has been discussed in [#709]. To address a few of the concerns
from that issue:

1. More than 1 `regex!` can be used in a block (for the anonymous
   version, the name is in a scope).
2. Using `regex_automata`'s `Lazy` means there are no MSRV concerns, and
   this works without std.
3. There is no compile-time checking. The docs make it clear that
   `regex!` should not be used with invalid regex and suggest enabling
   the clippy lint, leaving the door somewhat open for e.g. a `const fn`
   syntax validator in the future (though this seems unlikely).

A `regex::bytes::regex!` version is also included.

Closes: https://github.com/rust-lang/regex/issues/709

[#709]: https://github.com/rust-lang/regex/issues/709

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,11 @@ fn main() {
 Specifically, in this example, the regex will be compiled when it is used for
 the first time. On subsequent uses, it will reuse the previous compilation.
 
+The [`regex!`] macro can also be used, which handles lazy compilation.
+
 [`std::sync::LazyLock`]: https://doc.rust-lang.org/std/sync/struct.LazyLock.html
 [`once_cell`]: https://crates.io/crates/once_cell
+[`regex!`]: https://docs.rs/regex/*/regex/macro.regex.html
 
 ### Usage: match regular expressions on `&[u8]`
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,31 @@ assert_eq!(results, vec![
 # Ok::<(), Box<dyn std::error::Error>>(())
 ```
 
+Or, make use of the [`regex!`](crate::regex) macro to compile the regex once
+and re-use that same regex automatically. This is useful inside functions.
+For example:
+
+```rust
+use regex::regex;
+
+fn is_match(line: &str) -> bool {
+    // The regex will be compiled approximately once and reused automatically.
+    // This avoids the footgun of using `Regex::new` here, which would
+    // guarantee that it would be compiled every time this routine is called.
+    // This would likely make this routine much slower than it needs to be.
+    regex!(r"bar|baz").is_match(line)
+}
+
+let hay = "\
+path/to/foo:54:Blue Harvest
+path/to/bar:90:Something, Something, Something, Dark Side
+path/to/baz:3:It's a Trap!
+";
+
+let matches = hay.lines().filter(|line| is_match(line)).count();
+assert_eq!(matches, 2);
+```
+
 # Overview
 
 The primary type in this crate is a [`Regex`]. Its most important methods are
@@ -497,6 +522,8 @@ fn main() {
 Specifically, in this example, the regex will be compiled when it is used for
 the first time. On subsequent uses, it will reuse the previously built `Regex`.
 Notice how one can define the `Regex` locally to a specific function.
+
+The [`regex!`] macro can also be used, which handles lazy compilation.
 
 [`std::sync::LazyLock`]: https://doc.rust-lang.org/std/sync/struct.LazyLock.html
 [`once_cell`]: https://crates.io/crates/once_cell
@@ -1350,4 +1377,10 @@ mod regexset;
 /// expression.
 pub fn escape(pattern: &str) -> alloc::string::String {
     regex_syntax::escape(pattern)
+}
+
+/// Public-but-unstable API for macro support.
+#[doc(hidden)]
+pub mod __private {
+    pub use regex_automata::util::lazy::Lazy;
 }
--- a/src/regex/bytes.rs
+++ b/src/regex/bytes.rs
@@ -4,6 +4,59 @@ use regex_automata::{meta, util::captures, Input, PatternID};
 
 use crate::{bytes::RegexBuilder, error::Error};
 
+/// A convenient way to construct regex patterns from string literals.
+///
+/// This macro can be used to construct reusable instances of [`Regex`] with
+/// reduced boilerplate. The constructed `Regex` is stored in a static so the
+/// pattern is compiled approximately once, even when called multiple times.
+///
+/// There is *no compile-time checking of patterns* with `regex!`. Instead,
+/// invalid patterns will panic the first time the regex is used. Invalid
+/// patterns should still not be used with `regex!`; if compile-time checking
+/// becomes feasible in the future, it may be added within a non-semver-breaking
+/// release. In the meantime, consider enabling [`clippy::invalid_regex`].
+///
+/// # Examples
+///
+/// ```
+/// use regex::bytes::{Regex, regex};
+///
+/// assert!(regex!("[a-z]").is_match(b"a"));
+/// assert!(regex!("(inconceivable!|classic blunder)").is_match(b"inconceivable!"));
+///
+/// let re: &Regex = regex!(r"(\d{3})-(\d{4})");
+/// assert_eq!(&re.captures(b"867-5309").unwrap()[1], b"867");
+/// ```
+///
+/// An invalid pattern will panic when it is first used:
+///
+/// ```should_panic
+/// use regex::bytes::regex;
+///
+/// let re = regex!("invalid -> ("); // no panic here
+/// re.is_match(b"invalid -> (");    // panic!
+/// ```
+///
+/// [`clippy::invalid_regex`]: https://rust-lang.github.io/rust-clippy/master/#invalid_regex
+// `macro_export` always makes the macro available at crate root. We hide
+// from documentation there and instead re-export it in `crate::bytes` to get
+// the desired behavior.
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __bytes_regex {
+    ($re:literal) => {{
+        static REGEX: $crate::__private::Lazy<$crate::bytes::Regex> =
+            $crate::__private::Lazy::new(|| {
+                $crate::bytes::Regex::new($re).expect("invalid regex pattern")
+            });
+
+        // Coerce returned type from `&Lazy<Regex>` to `&Regex` to avoid making the
+        // inner type public.
+        let re: &$crate::bytes::Regex = &REGEX;
+        re
+    }};
+}
+
 /// A compiled regular expression for searching Unicode haystacks.
 ///
 /// A `Regex` can be used to search haystacks, split haystacks into substrings
--- a/src/regex/string.rs
+++ b/src/regex/string.rs
@@ -4,6 +4,55 @@ use regex_automata::{meta, util::captures, Input, PatternID};
 
 use crate::{error::Error, RegexBuilder};
 
+/// A convenient way to construct regex patterns from string literals.
+///
+/// This macro can be used to construct reusable instances of [`Regex`] with
+/// reduced boilerplate. The constructed `Regex` is stored in a static so the
+/// pattern is compiled approximately once, even when called multiple times.
+///
+/// There is *no compile-time checking of patterns* with `regex!`. Instead,
+/// invalid patterns will panic the first time the regex is used. Invalid
+/// patterns should still not be used with `regex!`; if compile-time checking
+/// becomes feasible in the future, it may be added within a non-semver-breaking
+/// release. In the meantime, consider enabling [`clippy::invalid_regex`].
+///
+/// # Examples
+///
+/// ```
+/// use regex::{Regex, regex};
+///
+/// assert!(regex!("[a-z]").is_match("a"));
+/// assert!(regex!("(inconceivable!|classic blunder)").is_match("inconceivable!"));
+///
+/// let re: &Regex = regex!(r"(\d{3})-(\d{4})");
+/// assert_eq!(&re.captures("867-5309").unwrap()[1], "867");
+/// ```
+///
+/// An invalid pattern will panic when it is first used:
+///
+/// ```should_panic
+/// use regex::regex;
+///
+/// let re = regex!("invalid -> ("); // no panic here
+/// re.is_match("invalid -> (");     // panic!
+/// ```
+///
+/// [`clippy::invalid_regex`]: https://rust-lang.github.io/rust-clippy/master/#invalid_regex
+#[macro_export]
+macro_rules! regex {
+    ($re:literal) => {{
+        static REGEX: $crate::__private::Lazy<$crate::Regex> =
+            $crate::__private::Lazy::new(|| {
+                $crate::Regex::new($re).expect("invalid regex pattern")
+            });
+
+        // Coerce returned type from `&Lazy<Regex>` to `&Regex` to avoid making the
+        // inner type public.
+        let re: &$crate::Regex = &REGEX;
+        re
+    }};
+}
+
 /// A compiled regular expression for searching Unicode haystacks.
 ///
 /// A `Regex` can be used to search haystacks, split haystacks into substrings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,37 @@
-1.12.4 (2025-06-09)
+1.13.0 (2026-07-09)
+===================
+This release includes a new API, a `regex!` macro, for lazy compilation of
+a regex from a string literal. If you use regexes a lot, it's likely you've
+already written one exactly like it. The new macro can be used like this:
+
+```rust
+use regex::regex;
+
+fn is_match(line: &str) -> bool {
+    // The regex will be compiled approximately once and reused automatically.
+    // This avoids the footgun of using `Regex::new` here, which would
+    // guarantee that it would be compiled every time this routine is called.
+    // This would likely make this routine much slower than it needs to be.
+    regex!(r"bar|baz").is_match(line)
+}
+
+let hay = "\
+path/to/foo:54:Blue Harvest
+path/to/bar:90:Something, Something, Something, Dark Side
+path/to/baz:3:It's a Trap!
+";
+
+let matches = hay.lines().filter(|line| is_match(line)).count();
+assert_eq!(matches, 2);
+```
+
+Improvements:
+
+* [#709](https://github.com/rust-lang/regex/issues/709):
+Add a new `regex!` macro for efficient and automatic reuse of a compiled regex.
+
+
+1.12.4 (2026-06-09)
 ===================
 This release includes a performance optimization for compilation of regexes
 with very large character classes.
@@ -9,7 +42,7 @@ Improvements:
 Avoid re-canonicalizing the entire interval set when pushing new class ranges.
 
 
-1.12.3 (2025-02-03)
+1.12.3 (2026-02-03)
 ===================
 This release excludes some unnecessary things from the archive published to
 crates.io. Specifically, fuzzing data and various shell scripts are now
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -89,3 +89,7 @@ In general, one should expect performance on `&[u8]` to be roughly similar to
 performance on `&str`.
 */
 pub use crate::{builders::bytes::*, regex::bytes::*, regexset::bytes::*};
+
+// Re-export the public but hidden macro to make it usable as `bytes::regex`.
+#[doc(inline)]
+pub use crate::__bytes_regex as regex;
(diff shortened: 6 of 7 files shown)


src/regex/string.rs
```diff
@@ -4,6 +4,80 @@ use regex_automata::{meta, util::captures, Input, PatternID};
 
 use crate::{error::Error, RegexBuilder};
 
+/// A convenient way to construct regex patterns.
```
BurntSushi:
```suggestion
/// A convenient way to construct regex patterns from string literals.
```

src/regex/string.rs
```diff
@@ -4,6 +4,80 @@ use regex_automata::{meta, util::captures, Input, PatternID};
...
+///
+/// assert!(re!("[a-z]").is_match("a"));
+/// assert!(re!("(inconceivable!|classic blunder)").is_match("inconceivable!"));
+/// assert_eq!(&re!(r"\d+").captures("1234").unwrap()[0], "1234");
+/// ```
+///
+/// The static can also be named, allowing for reuse across functions:
+///
+/// ```
+/// # use regex::re;
```
BurntSushi:
```suggestion
/// use regex::re;
```

I don't like it when examples leave out stuff like this. I know std does it and... I'm not a huge fan. I would rather show users how the thing is being imported. I also find that the examples more closely model reality. :-)

src/regex/string.rs
```diff
@@ -4,6 +4,80 @@ use regex_automata::{meta, util::captures, Input, PatternID};
...
+/// assert_eq!(cap2("867-5309"), "5309");
+///
+/// // Visibility can be provided if needed.
+/// re!(pub IOCANE, "powder");
+/// ```
+///
+/// An invalid pattern will panic when it is first used:
+///
+/// ```should_panic
+/// # use regex::re;
```
BurntSushi:
```suggestion
/// use regex::re;
```

src/regex/string.rs
```diff
@@ -4,6 +4,80 @@ use regex_automata::{meta, util::captures, Input, PatternID};
...
+/// let re = re!("invalid -> ("); // no panic here
+/// re.is_match("invalid -> (");  // panic!
+/// ```
+///
+/// [`LazyLock`]: https://doc.rust-lang.org/std/sync/struct.LazyLock.html
```
BurntSushi: I think just [`LazyLock`] will link to the std docs automatically.

src/regex/string.rs
```diff
@@ -4,6 +4,80 @@ use regex_automata::{meta, util::captures, Input, PatternID};
...
+#[cfg(feature = "std")]
+macro_rules! re {
+    // ($vis:vis) => {};
+    ($re:literal) => {{
+        $crate::re!(REGEX, $re);
+        // Coerce returned type from `&LazyLock<Regex>` to `&Regex`
+        let re: &$crate::Regex = &REGEX;
+        re
+    }};
+    ($vis:vis $name:ident, $re:literal) => {
```
BurntSushi: Is there precedent for these sorts of macros that introduce a name in the ecosystem? Is this the right syntax? What about `re!(SOME_REGEX = "blah blah")`?

src/regex/string.rs
```diff
@@ -4,6 +4,80 @@ use regex_automata::{meta, util::captures, Input, PatternID};
 
 use crate::{error::Error, RegexBuilder};
 
+/// A convenient way to construct regex patterns.
+///
+/// This macro can be used to construct reusable instances of [`Regex`] with
+/// reduced boilerplate. The constructed `Regex` is wrapped in a [`LazyLock`] so
```
BurntSushi: I don't like that we discuss the implementation here. I think we should talk about semantics and not how it's achieved. (Unless you feel like it's really important, in which case, it should go in a separate "Implementation" block that makes it clear this is just an FYI and could change.)

For example, I kind of wonder whether it would be better to use [`regex-automata::util::lazy::Lazy`](https://docs.rs/regex-automata/latest/regex_automata/util/lazy/struct.Lazy.html). It gives up the "at most once" guarantee, but works in core-only environments.

It might be nice to keep the `#[cfg(feature = "std")]` gate here for now, but switch to regex-automata's `Lazy`. If no issues appear after a while, then we could remove the feature gate. But that still gives us the flexibility to switch to `LazyLock` if there are problems.

BurntSushi: Note that in order to use `Lazy` from `regex-automata`, we'd need to define a wrapper type to avoid making `regex-automata` a public dependency. Then that needs to be exported in the crate root. That's kind of a bummer... And if we ever switched back to `LazyLock`, that type would need to stay and it would be vestigial. Blah...

You might also wonder why targeting core-or-alloc-only use cases even matters... I'm thinking about libraries that depend on `regex` without any features enabled. They can't use this macro unless they setup their own `std` feature. Basically, any time you have a `#[cfg(feature = "...")]` around a public API, you're passing the buck of dealing with that to other downstream libraries. So I try hard to avoid putting such things in public APIs for exactly that reason.

Thoughts?

BurntSushi (changes requested):
Thank you!

I think we also want a copy of this macro in the `bytes` sub-module.

I continue to overall be in favor of this, but the naming here is brutal. I really really continue to like the name `regex!` for this. The repetition is unfortunate, but even with the terser name of `re` here, you chose to use its unqualified form in examples. Which to me is an argument in favor of using `regex` as the name for the macro. IMO, the benefit of a terse name is being able to use its qualified form, i.e., `regex::re!(...)`.

I think `re` is not a horrible name though. I think it's either that or `regex` personally.

src/regex/string.rs
tgross35: I had gone back and forth on this a bit; I didn't know there was a reasonably easy solution in `regex_automata`, that sounds well worth using to me.

Any thoughts about using `#[doc(hidden)]` pub mod __private { pub use regex_automata::util::lazy::Lazy; }`? That pattern always feels a bit hacky but it does avoid the need for a public wrapper, and expresses some intent to keep the inner workings opaque. Technically somebody with a `regex_automata` dependency could still do:

```rust
re!(FOO, "abc");
let _: &regex_automata::util::lazy::Lazy<Regex> = &FOO;
```

But that doesn't seem worth worrying about.

A wrapper could also live in a private module like that, and/or make the needed API match `LazyLock` so it could become a type alias one day if that is desired.

src/regex/string.rs
tgross35: I'm not familiar with any, and with an extremely quick skim of the top ~100 crates on crates.io I didn't see anything. I have a slight preference for the comma version because it reads a bit more like what you might find in `(...)` function call syntax elsewhere

It's also possible to drop this named version if you prefer. I do think it's useful though, I have a macro like this in a few crates and use it nearly as much as the anonymous version (e.g. grouping related regex patterns at the top of a file when they are used in different functions).

tgross35:
> Thank you!
> 
> I think we also want a copy of this macro in the `bytes` sub-module.

I made a mention of this in the top post but is there a way to do this? `#[macro_export]` makes it available at the root, `pub use some_macro;` complains about it only being public within the crate. It could be under a different name, but that feels less nice.

> I continue to overall be in favor of this, but the naming here is brutal. I really really continue to like the name `regex!` for this. The repetition is unfortunate, but even with the terser name of `re` here, you chose to use its unqualified form in examples. Which to me is an argument in favor of using `regex` as the name for the macro. IMO, the benefit of a terse name is being able to use its qualified form, i.e., `regex::re!(...)`.
> 
> I think `re` is not a horrible name though. I think it's either that or `regex` personally.

Generally I think that with any name, this is something much more likely to be imported than qualified - I only mentioned that because it was brought up in the issue. I just prefer `re!` because it's a very familiar abbreviation and types/reads/formats a tiny bit closer to "just a string" like one of the `/.../` languages.

But that's subjective and three extra characters won't make or break anything, so I'm happy to defer to your preference :)

src/regex/string.rs
BurntSushi: I do kind of like it. But yeah, I'd say drop it for now. I'm a little hesitant to be blazing a trail here. Also, this can be added back later without breaking changes.

src/regex/string.rs
BurntSushi: The problem with that approach is that it becomes difficult to name the type returned by the macro. I _hate_ it when libraries do this, because there are a variety of reasons why naming a type is useful. I would much rather just add a wrapper type to the crate root than be the source of such annoyances. :-) It's a bit odd and out of place, but I think I can stomach it.

Having the wrapper type's API match a subset of `LazyLock` seems very wise. Good idea.

BurntSushi:
> I made a mention of this in the top post but is there a way to do this? `#[macro_export]` makes it available at the root, `pub use some_macro;` complains about it only being public within the crate. It could be under a different name, but that feels less nice.

Ugh, really!?!? I'm surprised I didn't know this. Probably because I almost never publish macros.

Yeah unfortunately just leave it out. Using a different name would suck. And byte regexes are likely used way less frequently.

> Generally I think that with any name, this is something much more likely to be imported than qualified - I only mentioned that because it was brought up in the issue. I just prefer `re!` because it's a very familiar abbreviation and types/reads/formats a tiny bit closer to "just a string" like one of the `/.../` languages.
> 
> But that's subjective and three extra characters won't make or break anything, so I'm happy to defer to your preference :)

I'm going to link this PR on some socials and see what we get as a vibe check.

src/regex/string.rs
tgross35: The returned type of the anonymous syntax should just be a `&'static Regex`, the bit at https://github.com/rust-lang/regex/pull/1371/changes#diff-3be4b210bb3c85bbca46bfb05dfed4dc66c9564a5761c1a9b7bd1e9524c57b24R69-R70 is to keep the `LazyLock` workings opaque. I think the only way the `LazyLock`/`Lazy` could leak is via the named versions. As-written, users _should_ do:

```rust
re!(FOO, "abc");
let _: &Regex = FOO;
```

but _can_ do:

```rust
let _: &LazyLock<Regex> = FOO;
```

If we're dropping the named version then I think this concern goes away?

(With https://github.com/rust-lang/rust/issues/63065 we could hide this more by expanding to `static FOO: impl Deref<Target = Regex>`, but that's not happening anytime soon. Maybe there are some better tricks.)

BurntSushi: I was just coming back to write a comment correcting what I wrote, hah. OK, in this case, I'm okay with using a `doc(hidden)` hack on a re-export of `regex-automata`'s type directly. And if we ever add the named version back, then we can figure out what to do then.

src/regex/string.rs
tgross35: Applied in the latest push

src/regex/string.rs
tgross35: I tend not to like it in most cases, but feel it helps the flow when multiple examples can be read together as a block with interspersed docs. That said, this example is no longer relevant and removed :)

src/regex/string.rs
tgross35: Applied in the latest push

src/regex/string.rs
tgross35: Huh, that's good to know. Link no longer needed, in any case

src/regex/string.rs
tgross35: Removed in the latest push

src/regex/string.rs
tgross35: Switched to the `regex-automata` version in the latest update and updated the docs to just say it's a static without giving details.

soundslocke:
Arriving here from socials to provide a vibe check! I dig this a lot and look forward to the convenience. Seems quite nice to me, I'd immediately use it. :+1:

I like and prefer `regex!` for the name but also can't say I have a horse in the race.

src/regex/string.rs
```diff
@@ -4,6 +4,59 @@ use regex_automata::{meta, util::captures, Input, PatternID};
...
+/// pattern is only compiled once, even when called multiple times.
+///
+/// There is *no compile-time checking of patterns* with `re!`. Instead, invalid
+/// patterns will panic the first time the regex is used. Invalid patterns
+/// should still not be used with `re!`; if compile-time checking becomes
+/// feasible in the future, it may be added within a non-semver-breaking
+/// release. In the meantime, consider enabling [`clippy::invalid_regex`].
+///
+/// Because `LazyLock` is used internally, `re!` can only be used with the
+/// `std` feature and with Rust 1.80 or greater.
```
tisonkun: The PR description says:

> Using regex_automata's Lazy means there are no MSRV concerns, and this works without std.

So I guess this is no longer a prerequisite?

tisonkun:
> re!

There are three common abbreviations for regular expressions:

1. `re`: as in Python
2. `regex`: as in this crate
3. `regexp`: as a flavor on cpan.org

This crate uses `regex` as its name, and there is no occurrence of `re` or `regexp` elsewhere. So I'd suppose `regex::regex!` is a better choice, with a `use regex::regex` and thus `regex!(...)` in most cases.

BurntSushi:
I also favor `regex!` personally. Its unqualified use feels a lot nicer.

With that said, `re` is a very common variable name for regexes. e.g., `let re = Regex::new(..).unwrap()`.

BurntSushi:
MSRV is a non-issue here in any direction. I should bump to Rust 2024 anyway.

lespea:
What if you tack on a `_b` for the byte macros (to address the name conflicts).  So it'd just be `re_b!(...)` for example?

BurntSushi:
I'm pretty meh on `re_b!(..)` or `regex_b!(..)`. The underscore in particular with a single letter drives me nuts. I'd probably prefer `reb` or `regexb` or `bre` or `bregex` or even `byte_regex`. With that said, I think we should start without it. If folks end up pining for the byte regex version of this, we can revisit. But I suspect byte regexes aren't used nearly as much as str regexes.

tisonkun:
You can have a macro being used as `regex::bytes::regex!`, with a cost of a "hidden" top-level macro:

```rust
pub mod bytes {
    #[macro_export]
    macro_rules! __regex {
        ($re:literal) => {{ .. }};
    }

    pub use __regex as regex;
}
```

IIRC this is a historical issue that should be resolved with https://github.com/rust-lang/rust/issues/39412 but 39412 is also a long-standing issue :P

BurntSushi:
I'm okay with a hidden top-level macro, as long as we can put a `#[doc(hidden)]` on it.

isuffix:
Came here from the bluesky post :)

You can add `#[doc(hidden)]` to the original macro, and it will work with `#[doc(inline)]` on the re-export. In Typst we do the same thing, and I recently added a [detailed comment](https://github.com/typst/typst/blob/abccee142cc19f3f3dc45ff9e3fac177c315a87b/crates/typst-library/src/diag.rs#L184-L206) explaining it, which further links to the ["Import and Export" chapter](https://lukaswirth.dev/tlborm/decl-macros/minutiae/import-export.html) of [The Little Book of Rust Macros](https://lukaswirth.dev/tlborm).

You can see the resulting docs [here](https://docs.rs/typst-library/0.15.0/typst_library/diag/#macros).

tgross35:
With the `#[doc(hidden)]` tricks, what is the way to make documentation look correct? With this in `bytes`:

```rust
/// __bytes_regex summary
#[doc(hidden)]
#[macro_export]
macro_rules! __bytes_regex { /* ... */}

/// Export summary
#[doc(inline)]
pub use __bytes_regex as regex;
```

That module shows (no idea why the logic seems to be to join the summaries):

image

Removing the `#[doc(inline)]`  hides both macros.

src/regex/string.rs
tgross35: Thanks, you're right. Removed

src/regex/bytes.rs
```diff
@@ -4,6 +4,29 @@ use regex_automata::{meta, util::captures, Input, PatternID};
 
 use crate::{bytes::RegexBuilder, error::Error};
 
+// TODO needs advice
+
+/// __bytes_regex summary
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __bytes_regex {
+    // ($vis:vis) => {};
```
tisonkun:
```suggestion
```

Seems like a redundant comment-out line?

src/regex/string.rs
```diff
@@ -4,6 +4,56 @@ use regex_automata::{meta, util::captures, Input, PatternID};
...
+/// use regex::regex;
+///
+/// let re = regex!("invalid -> ("); // no panic here
+/// re.is_match("invalid -> (");  // panic!
+/// ```
+///
+/// [`clippy::invalid_regex`]: https://rust-lang.github.io/rust-clippy/master/#invalid_regex
+#[macro_export]
+macro_rules! regex {
+    // ($vis:vis) => {};
```
tisonkun:
```suggestion
```

ditto

BurntSushi:
Yeah, `__bytes_regex` should not show up in the top-level crate docs. Do we need `doc(inline)`? I guess if _not_ having it means both macros get removed, then we're stuck with either "show both" or "show neither"? That's a bummer. It looks like typst has it working okay though? https://docs.rs/typst-library/0.15.0/typst_library/diag/index.html

src/regex/bytes.rs
tgross35: Thanks, removed

tgross35:
Oh, sorry for not clarifying; the screenshot above is in the bytes module. Neither `__bytes_regex` nor the bytes form of `regex` show up in the top-level module, only the default string `regex`.

That said, this looks identical to the `typst` config so I'm hoping @isuffix might have a hint. Unless my local toolchain (2026-06-29) happens to do something different than docs.rs.

isuffix:
It seems the issue is that the macro is in `src/regex/bytes.rs` which is then re-exported by `src/bytes.rs` via `pub use crate::{builders::bytes::*, regex::bytes::*, regexset::bytes::*};` which somehow (???) causes `rustdoc` to add docs for both items.

Luckily it seems that moving just the re-export to `src/bytes.rs` solves it!

```rust
// src/bytes.rs:
/*! ... 90 line doc comment ... */
pub use crate::{builders::bytes::*, regex::bytes::*, regexset::bytes::*};

#[doc(inline)]
pub use crate::__bytes_regex as regex;
// Note that `crate::` is now required
```

tgross35:
That works 🎉! I think everything else should be addressed

src/lib.rs
```diff
@@ -498,6 +498,8 @@ Specifically, in this example, the regex will be compiled when it is used for
 the first time. On subsequent uses, it will reuse the previously built `Regex`.
 Notice how one can define the `Regex` locally to a specific function.
 
+The [`regex!`] macro can also be used, which handles `LazyLock` creation.
```
BurntSushi:
```suggestion
The [`regex!`] macro can also be used, which handles lazy compilation.
```

README.md
```diff
@@ -105,8 +105,11 @@ fn main() {
 Specifically, in this example, the regex will be compiled when it is used for
 the first time. On subsequent uses, it will reuse the previous compilation.
 
+The [`regex!`] macro can also be used, which handles `LazyLock` creation.
```
BurntSushi:
```suggestion
The [`regex!`] macro can also be used, which handles lazy compilation.
```

BurntSushi (approved):
Let's do this!

BurntSushi:
I pushed up some small tweaks to the wording:

* I added a more prominent example to the top of the crate docs. We should shout this new thing to the world!
* I added a CHANGELOG entry.
* I clarified that this will compile the regex "approximately once" instead of "at most once." `LazyLock` provides the latter guarantee, but `Lazy` from `regex-automata` doesn't give a strong guarantee. It's plausible we may want to use `LazyLock` when `std` is enabled to provide this, but I think we can do that later if it comes up.

BurntSushi:
This PR is on crates.io in `regex 1.13.0`!

PR #1377, opened 2026-07-24
Expose Regex memory usage info (fixes #943)

### LINKED ISSUE #943: API to get size of compiled regex

ashdnazg:
Based on https://github.com/rust-lang/regex/pull/1180

### DIFF
--- a/src/regex/bytes.rs
+++ b/src/regex/bytes.rs
@@ -1482,6 +1482,15 @@ impl Regex {
     pub fn locations(&self) -> CaptureLocations {
         self.capture_locations()
     }
+
+    /// Return the total approximate heap memory, in bytes, used by this `Regex`.
+    ///
+    /// Note that this value is unrelated to [`RegexBuilder::size_limit`], and can
+    /// in many cases exceed it.
+    #[inline]
+    pub fn memory_usage(&self) -> usize {
+        self.meta.memory_usage() + self.pattern.len()
+    }
 }
 
 /// Represents a single match of a regex in a haystack.
--- a/src/regex/bytes.rs
+++ b/src/regex/bytes.rs
@@ -1485,8 +1485,8 @@ impl Regex {
 
     /// Return the total approximate heap memory, in bytes, used by this `Regex`.
     ///
-    /// Note that this value is unrelated to [`RegexBuilder::size_limit`], and can
-    /// in many cases exceed it.
+    /// Note that this value does not have a defined relationship to
+    ///  to [`RegexBuilder::size_limit`].
     #[inline]
     pub fn memory_usage(&self) -> usize {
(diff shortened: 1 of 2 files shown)


src/regex/bytes.rs
```diff
@@ -1482,6 +1482,15 @@ impl Regex {
     pub fn locations(&self) -> CaptureLocations {
         self.capture_locations()
     }
+
+    /// Return the total approximate heap memory, in bytes, used by this `Regex`.
+    ///
+    /// Note that this value is unrelated to [`RegexBuilder::size_limit`], and can
+    /// in many cases exceed it.
```
BurntSushi:
```suggestion
    /// Note that this value does not have a defined relationship to
    ///  to [`RegexBuilder::size_limit`].
```

src/regex/string.rs
```diff
@@ -1481,6 +1481,15 @@ impl Regex {
     pub fn locations(&self) -> CaptureLocations {
         self.capture_locations()
     }
+
+    /// Return the total approximate heap memory, in bytes, used by this `Regex`.
+    ///
+    /// Note that this value is unrelated to [`RegexBuilder::size_limit`], and can
+    /// in many cases exceed it.
```
BurntSushi:
```suggestion
    /// Note that this value does not have a defined relationship to
    ///  to [`RegexBuilder::size_limit`].
```

BurntSushi (changes requested):

BurntSushi:
So actually, thinking through this more, I think this routine is probably totally bunk. In particular, it does not account for the heap memory used by the lazy DFA, which is probably the biggest (ab)user of heap in most regexes.

It's non-trivial to get that heap usage because it's guarded behind a [`Pool`](https://docs.rs/regex-automata/latest/regex_automata/util/pool/struct.Pool.html), which specifically does not expose any way of telling how big it is. And doing so is not straight-forward since it's guarded by mutexes.

Unfortunately I think that if you really really need to know about the sizes of regex values, then you'd have to drop down to `regex-automata` and dead with [`Cache`](https://docs.rs/regex-automata/latest/regex_automata/meta/struct.Cache.html) values directly.

Unfortunately, I just cannot think of a way to write documentation for this routine, as-is, that isn't some kind of outright lie in a meaningful sense.

ashdnazg:
That makes sense, but it kinda means `meta::Regex::memory_usage()` is bunk as well.

PR #1378, opened 2026-08-02
Move prefilter into core builder without cloning

fyrsta7:
## Summary

This moves the optional prefilter into `Core::new` instead of cloning it when building a meta strategy.

## Why this can improve performance

The builder already owns the optional prefilter at this point, and `Info::new` has finished inspecting it before `Core::new` is called. Moving the value avoids one clone during regex construction while keeping ownership local to the built strategy.

## Validation

- `git diff --check`
- `cargo check -p regex-automata`

## Benchmark

Current machine, release microbenchmarks, 9 samples, median:

| Benchmark | Base | This change | Delta |
|---|---:|---:|---:|
| Literal-heavy construction, 12 `Regex::new` calls | 2,321,393.938 ns | 2,314,070.512 ns | 0.32% faster |
| Mixed construction, 64 `Regex::new` calls | 3,956,117.780 ns | 4,046,411.233 ns | 2.28% slower |

I am opening this as a draft because the targeted prefilter-heavy construction case improves slightly, while the mixed construction benchmark was not favorable on this machine.


fyrsta7:
Closing this draft because follow-up local benchmarking showed the original change was not strong enough for a performance PR. I opened a smaller replacement draft with positive fresh benchmark data in #1379.

